PR #549, opened 2018-07-28
WIP. Display CryptoKitties

hboon:
@James-Sangalli Added some code in a feature branch to display CryptoKitty image in the app. Here's what's popped up:

The asset definition XML for CryptoKitties needs:

* A field representing the kitten ID and how to extract it
* A way to specify that this field is the ID used to form the URL to retrieve the image (and that there is image support for this contract)
* URL format to get the image. e.g for CryptoKitties: "https://img.cn.cryptokitties.co/#{contract_address}/#{id}.svg" (also to finalize which delimiter to use. #{})
* Be good if when specifying the URL format for the static image, we can also specify the image format (SVG, PDF, JPEG, PNG, etc)

CryptoKitties-specific
---

Their API for getting a kitty https://api.cryptokitties.co/kitties/1 includes the URLs for retrieving the static image under the keys `image_url` and `image_url_cdn` with https://api.cryptokitties.co/kitties/800058:

```
"image_url":"https://storage.googleapis.com/ck-kitty-image/0x06012c8cf97bead5deae237070f9587f8e7a266d/800058.svg"
"image_url_cdn":"https://img.cn.cryptokitties.co/0x06012c8cf97bead5deae237070f9587f8e7a266d/800058.svg"
```

We can probably use the value of `image_url_cdn`?

But out of pure luck, I found out that kitty 1 (https://api.cryptokitties.co/kitties/1) uses png:

```
"image_url":"https://storage.googleapis.com/ck-kitty-image/0x06012c8cf97bead5deae237070f9587f8e7a266d/1.png"
"image_url_cdn":"https://img.cn.cryptokitties.co/0x06012c8cf97bead5deae237070f9587f8e7a266d/1.png"
```

I tried a few kitty IDs randomly and they all seem to be using `.svg`, so maybe it's just that odd one out. So it looks like in addition to rasterized images, we need to be able to display `SVG`s.

General
---
It seems like we need a way to describe which fields and how they are laid out in the UI


hboon:
Closing, replaced by #581

Issue #559, opened 2018-08-03
Encode/decode rpc calls inside the app 


James-Sangalli:

hboon:
@James-Sangalli anything else we need to do for this?

James-Sangalli:
Should be all good, claiming orders does not require this.

PR #566, opened 2018-08-11
Experiment with using web3swift pod to fetch ERC 875 balance with a single call instead of 3 calls: encode+fetch+decode

hboon:
This is a throwaway PR for #559. Don't merge.

@James-Sangalli 

In this PR, we make a *blocking* call with the web3swift followed by the original 3-RPC-HTTP-calls with web3js (encode + fetch balance + decode). This is the console output (the balance is fetched 4 times, in case it might be confusing):

```
time taken with web3swift pod to fetch ERC875 balance: 1795.12596130371
time taken with web3swift pod to fetch ERC875 balance: 715.092062950134
time taken with web3swift pod to fetch ERC875 balance: 765.774011611938
time taken with manual encode/decode with fetch ERC875 balance : 2740.95106124878
time taken with manual encode/decode with fetch ERC875 balance : 3814.82100486755
time taken with manual encode/decode with fetch ERC875 balance : 2081.75408840179
time taken with web3swift pod to fetch ERC875 balance: 841.801047325134
time taken with manual encode/decode with fetch ERC875 balance : 507.031917572021
```

The results are a little strange. When it's slow (maybe due to RPC server load?), it's much slower with the 3-calls. Naturally.
But when it's not so slow, the 3-calls might even be faster than the web3swift call, though not as significantly, which is really odd.

It still seems worth migrating over, the code is shorter and the performance improvements is probably more noticeable on higher latency networks.

What do you think?

PS: It's a blocking call for now because web3swift waits for the promise to be fulfilled (!!), so we can either issue a PR to fix that or just use GCD to do it on a background queue.


James-Sangalli:
@hboon yes we should, more sensible and doesn't rely on node for encoding and decoding

James-Sangalli:
@hboon this still uses rpc? We can actually encode and decode natively.

hboon:
@James-Sangalli oh yes, working on this.

It uses RPC only for the actual call: balanceOf(), etc. Encoding, decoding is done natively.

Issue #578, opened 2018-08-15
GitHub's "We found a potential security vulnerability in one of your dependencies" warning


hboon:
There's a security warning when we visit https://github.com/alpha-wallet/alpha-wallet-ios

colourful-land:
FYI:
A dependency defined in JS/package-lock.json has known security vulnerabilities and should be updated.

PR #580, opened 2018-08-15
update hoek dependency

James-Sangalli:
#578


hboon:
@James-Sangalli do we need to do this? http://www.codedependant.net/2016/01/26/override-nested-dependencies-npm-3/

James-Sangalli:
@hboon I updated all references of it in other dependencies as well so it should be fine.

Issue #574, opened 2018-08-13
Implement OpenSea.io for popular collectables


James-Sangalli:

James-Sangalli:
@hboon can use this branch: https://github.com/alpha-wallet/alpha-wallet-ios/tree/opensea-implementation-erc721

hboon:
- [ ] Make pretty

Issue #573, opened 2018-08-13
Fallback erc721 


James-Sangalli:
If there is no equal of getKitty(indexId) then fallback to tokensOfOwner() and fill the UI with the token indices e.g. 851020, 777777, ... etc

James-Sangalli:
Appears that tokensOfOwner is not mandatory, therefore it seems logical to use balanceOf and list by count

PR #581, opened 2018-08-15
opensea.io implementation erc721

hboon:
For #574

This PR pulls CryptoKitty information for kitties in the wallet using opensea.io's API, stores and display them

1. Need some design help to make it more presentable, probably after #573 is done too.

2. There are still references to the word "Ticket" such as in "Tickets To Transfer". Will fix those later (maybe with #573)

3. We stuff this as the balance (per cat). Not the best DB schema design, but I was thinking that since this is localized to the app internals, we can delay any changes to the schema. We can look into this again if need be when we need to handle a more general case:

```
{"traits":[{"count":96546,"type":"mouth","value":"soserious"},{"count":63591,"type":"body","value":"ragdoll"},{"count":55099,"type":"pattern","value":"amur"},{"count":24679,"type":"eyes","value":"wonky"},{"count":46592,"type":"generation","value":"2"},{"count":56123,"type":"cooldown_index","value":"1"},{"count":26287,"type":"colorprimary","value":"bananacream"},{"count":19473,"type":"coloreyes","value":"coralsunrise"},{"count":19978,"type":"colorsecondary","value":"egyptiankohl"},{"count":10419,"type":"colortertiary","value":"frosting"}],"externalLink":"https:\/\/www.cryptokitties.co\/kitty\/800058","description":"Yaya. My name is Kitty #800058. You might not believe it, but I've seen a centaur before, and it was actually very frivolous. I've been trying to model my life after Crookshanks, but most of the time all I can think about is salt water taffy. Are you a parking ticket? ‘Cause you’ve got fine written all over you.","tokenId":"800058","thumbnailUrl":"https:\/\/storage.googleapis.com\/opensea-prod.appspot.com\/0x06012c8cf97bead5deae237070f9587f8e7a266d-thumbnail\/800058.png"}
```

4. Have incremented the Realm DB version despite no changes to the schema just so we have a point where we can look back in the future, if we need to handle data migrations

5. @James-Sangalli: I have forced pushed this branch

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -277,6 +277,7 @@
 		5E7C701BFF4469B35A074EB9 /* RequestViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C767497AD8DEE83F384D7 /* RequestViewModel.swift */; };
 		5E7C70AE62DBB193399C7F5E /* ServerViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C7CDB0BAD5D27D2F24F57 /* ServerViewCell.swift */; };
 		5E7C70BE9AE35408038E1971 /* HelpContentsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C7B089FD4C96810DD10FD /* HelpContentsViewController.swift */; };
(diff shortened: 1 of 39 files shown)


hboon:
@JamesSmartCell These might be useful for you too:

* Constants.cryptoKittiesCooldowns at https://github.com/alpha-wallet/alpha-wallet-ios/pull/581/files#diff-e699f6eea97b0ba8a959808d46b63cf6R59
* TokensDatastore.getCryptoKittyBalance() at https://github.com/alpha-wallet/alpha-wallet-ios/pull/581/files#diff-e4536461d5971b9f2128ec3033d695f8R218

hboon:
This PR displays the wrong issuer, which should be fixed with #529

Issue #575, opened 2018-08-13
Implement new format for magic links


James-Sangalli:
Adds an extra byte to the front indicating whether it is the index or tokenId format

PR #586, opened 2018-08-16
added new byte to link

James-Sangalli:
for #575 

Probably needs a refactor...

### DIFF
--- a/AlphaWallet/Market/OrderHandler.swift
+++ b/AlphaWallet/Market/OrderHandler.swift
@@ -80,7 +80,7 @@ public class OrderHandler {
                                tickets: [UInt16],
                                contractAddress: String) -> [UInt8] {
         //ticket count * 2 because it is 16 bits not 8
-        let arrayLength: Int = 84 + tickets.count * 2
+        let arrayLength: Int = 85 + tickets.count * 2
         var buffer = [UInt8]()
         buffer.reserveCapacity(arrayLength)
 
--- a/AlphaWallet/Market/UniversalLinkHandler.swift
+++ b/AlphaWallet/Market/UniversalLinkHandler.swift
@@ -96,11 +96,17 @@ public class UniversalLinkHandler {
     
     //shortens price and expiry
     func formatMessageForLink(signedOrder: SignedOrder) -> String {
-        let message = signedOrder.message
+        //append encodable byte to differiantiate between spawnable and index
+        //based links
+        var message = [UInt8]()
+        message.append(UInt8(0x01))
+        for i in 0..<signedOrder.message.count {
+            message.append(signedOrder.message[i])
+        }
         let indices = decodeTicketIndices(indices: signedOrder.order.indices)
         var messageWithSzabo = [UInt8]()
-        let price = Array(message[0...31])
-        let expiry = Array(message[32...63])
+        let price = Array(message[0...32])
+        let expiry = Array(message[33...64])
         let priceHex = MarketQueueHandler.bytesToHexa(price)
         let expiryHex = MarketQueueHandler.bytesToHexa(expiry)
         //removes leading zeros
@@ -110,13 +116,13 @@ public class UniversalLinkHandler {
         let priceSzabo = priceInt / 1000000000000
         var priceBytes = formatTo4Bytes(priceSzabo.serialize().bytes)
         var expiryBytes = formatTo4Bytes(expiryInt.serialize().bytes)
-        for i in 0...3 {
+        for i in 0...4 {
             messageWithSzabo.append(priceBytes[i])
         }
         for i in 0...3 {
             messageWithSzabo.append(expiryBytes[i])
         }
-        for i in 64...83 {
+        for i in 65...84 {
             messageWithSzabo.append(message[i])
         }
         for i in 0...indices.count - 1 {


AlphaWallet/Market/OrderHandler.swift
```diff
@@ -80,7 +80,7 @@ public class OrderHandler {
                                tickets: [UInt16],
                                contractAddress: String) -> [UInt8] {
         //ticket count * 2 because it is 16 bits not 8
-        let arrayLength: Int = 84 + tickets.count * 2
+        let arrayLength: Int = 85 + tickets.count * 2
```
hboon: This doesn't seem to do anything. `reserveCapacity()` is only to give a hint  for the array's internal capacity for better performance.

AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -96,11 +96,17 @@ public class UniversalLinkHandler {
     
     //shortens price and expiry
     func formatMessageForLink(signedOrder: SignedOrder) -> String {
-        let message = signedOrder.message
+        //append encodable byte to differiantiate between spawnable and index
```
hboon: This PR only makes generation of universal links include an additional byte, and not parse universal links with that additional byte?

The test UniversalLinkHandlerTests.testUniversalLinkParser() is failing.

James-Sangalli:
@hboon this now handles importing a spawnable link but needs to be tested with the website and issuance of token ids.

James-Sangalli:
@hboon this is ready for review now as it is backward compatible

James-Sangalli:
@hboon note that the paymaster and website needs to be updated

AlphaWallet/Tokens/Coordinators/ClaimOrderCoordinator.swift
```diff
@@ -22,23 +22,52 @@ class ClaimOrderCoordinator {
...
+    func claimOrder(signedOrder: SignedOrder,
                     expiry: BigUInt,
                     v: UInt8,
                     r: String,
                     s: String,
                     completion: @escaping (Result<String, AnyError>) -> Void
         ) {
-        let request = ClaimERC875Order(expiry: expiry, indices: indices, v: v, r: r, s: s)
+
+        if let tokenIds = signedOrder.order.tokenIds {
```
hboon: I think the conditions here has to be tweaked a bit because `signedOrder.order.tokenIds` might be `nil`. I suppose when it is `nil`, we want it to execute the `else` too? If that's what we want, then you can change this:

    if let tokenIds = signedOrder.order.tokenIds {
        if !tokenIds.isEmpty {

to:

    if let tokenIds = signedOrder.order.tokenIds, if !tokenIds.isEmpty {
        //the if's body as it was for spawable
    } else {
        //the else's body as it was for non-spawable
    }

Also, maybe better to extract another method, say, `claimNonSpawnableOrder()`? So it's like this:

    func claimOrder(...) {
        if let tokenIds = signedOrder.order.tokenIds, if !tokenIds.isEmpty {
            claimSpawnableOrder()
        } else {
            claimNonSpawnableOrder()
        }
    }

    func claimNonSpawnableOrder(...) {
        //What was in else's body
    }

AlphaWallet/Transactions/Coordinators/TicketsCoordinator.swift
```diff
@@ -0,0 +1,611 @@
+//  TicketsCoordinator.swift
```
hboon: `AlphaWallet/Transactions/Coordinators/TicketsCoordinator.swift` is a leftover from previous rename/refactoring. We can remove it.

AlphaWallet/Tokens/Coordinators/ClaimOrderCoordinator.swift
hboon: Maybe as a future refactoring, we can also create something like:

enum OrderType {
    case spawnable
    case nonSpawnable
}

and stuff some variables or even logic inside. The usual strong-typing stuff.

AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -56,7 +63,16 @@ public class UniversalLinkHandler {
...
+        guard var linkBytes = Data(base64Encoded: linkInfo)?.array else { return nil }
+        let encodingByte = linkBytes[0]
+        switch encodingByte {
+        case Constants.oldFormat: break
+        //new link format, remove extra byte and continue
+        case Constants.notSpawnable: linkBytes.remove(at: 0)
+        case Constants.spawnable: return handleSpawnableLink(linkBytes: linkBytes)
+        case Constants.customizable: return handleSpawnableLink(linkBytes: linkBytes)
+        default: break
+        }
```
hboon: A `handleNonSpawnableLink()` would be good?

hboon: We can encode the link format byte value into the `LinkFormat` enum, keeping the byte values together in the enum makes it a bit easier to maintain them (and remove the corresponding `oldFormat` in `Constants.swift`):

    enum LinkFormat: UInt8 {
        case unassigned = 0x00
        case normal = 0x01
        case spawnable = 0x02
        case customizable = 0x03
    }

Then we can replace the switch with:

    if let format = LinkFormat(rawValue: encodingByte) {
        switch format {
        case .unassigned:
            break
        case .normal:
            //new link format, remove extra byte and continue
            linkBytes.remove(at: 0)
        case .spawnable:
            return handleSpawnableLink(linkBytes: linkBytes)
        case .customizable:
            return handleSpawnableLink(linkBytes: linkBytes)
        }
    } else {
        //TODO invalid format. Do something
    }

AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -113,17 +126,9 @@ public class UniversalLinkHandler {
...
-        var startPos: Int = 0
-        for i in 0..<tokenCount {
-            let tokenId: [UInt8] = Array(tokens[startPos...32 * i])
-            let tokenIdBigUInt = BigUInt(Data(bytes: tokenId))
-            tokenIds.append(tokenIdBigUInt)
-            startPos += 32
-        }
-        return tokenIds
+        let bytes = linkBytes[84..<linkBytes.count]
+        let tokenIds = bytes.split(separator: 32)
```
hboon: Nice! `let bytes...` is much simpler.

These 2 are different though:

    let tokens = bytes.split(intoChunksOf: 32)

    let tokenIds = bytes.split(separator: 32)

The former breaks the array into subarrays, each with length 32, the latter breaks the array into subarrays too, but by using the value `32` as the separator.

James-Sangalli:
@hboon should be ok now

hboon:
@James-Sangalli Think these aren't covered yet:

[  ] https://github.com/alpha-wallet/alpha-wallet-ios/pull/586/files#r219225965
[  ] https://github.com/alpha-wallet/alpha-wallet-ios/pull/586/files#r219228114

James-Sangalli:
@hboon ok, ready now.

PR #588, opened 2018-08-16
Run tests on Travis directly with `xcodebuild` instead of fastlane

hboon:
With this PR, tests should still fail, but at least Travis will give a hint of which test(s) failed.

> Executed 161 tests, with 1 failure (0 unexpected) in 3.722 (4.230) seconds

> Failing tests:
>	InCoordinatorTests.testShowRequstFlow()
> ** TEST FAILED **

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 
 script:
   - set -o pipefail
-  - fastlane travis_tests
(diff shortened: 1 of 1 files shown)


hboon:
The failing test(s) themselves have been fixed in #589

PR #589, opened 2018-08-16
Fix broken test(s) on Travis

hboon:
Ok. The key to reproducing this on a development machine is to delete the app in the simulator first.

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -93,7 +93,6 @@ class InCoordinator: Coordinator {
     }
 
     func fetchEthPrice() {
-        let keystore = try! EtherKeystore()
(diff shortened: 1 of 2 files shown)


hboon:
@James-Sangalli the debug button seems to be missing here https://travis-ci.com/alpha-wallet/alpha-wallet-ios. Maybe after the repo migration? Do you know how we can get it back?

hboon:
@James-Sangalli oh. According to https://docs.travis-ci.com/user/running-build-in-debug-mode/#enabling-debug-mode

> Private repositories have debug mode enabled by default, and no changes need to be made. To limit access to debug, grant users only read access to the repo, and use a fork + PR workflow. For public repositories, we have to enable it on a repository basis.
To enable debug for your public repositories, please email us at support@travis-ci.com and let us know which repositories you want activated.

James-Sangalli:
@hboon just transferred the subscription to alpha-wallet org

Issue #591, opened 2018-08-17
Reintroduce the dApp browser


James-Sangalli:
Probably best to base of the latest trust implementation for now.

hboon:
Ok, on this.

hboon:
Going to have to update some dependencies.

hboon:
- [ ] Implement
- [ ] Localise for Chinese

PR #595, opened 2018-08-20
Reintroduce (updated) dapps browser

hboon:
For #591 

Vendored some code in vendor/Trust-Core and some modification to the dapps browser to get it to work with our codebase. This is to avoid upgrading the TrustKeystore dependency and adding Trust-Core (which requires an upgraded TrustKeystore). Upgrading and adding those dependencies would change the account/wallet structure and will take much more time to code and test.

@@James-Sangalli can you please help to verify that it's working correctly?


James-Sangalli:
@hboon was able to do functions like sign and deploy, should be all good with all web3js functions in the browser

Issue #546, opened 2018-07-26
Rename code (not string) references of tickets to tokens (or nonfungible tokens)


hboon:

hboon:
There's just a few more types, including tests with "Tickets" in them. Going to rename them.

hboon:
- [ ] A few remaining types' names, include tests
- [ ] Method names
- [ ] var names

PR #614, opened 2018-08-22
Clean up more ticket refs to type names

hboon:
For #546


hboon:
@James-Sangalli do we need to be more explicit in these type names that they are NonFungibleToken(s)?

PR #623, opened 2018-08-24
downgrade web3js dependency to fix dapp browser

JamesANZ:


hboon:
That seems an awful lot of dependencies updated. There's a chance things might break (now due to code changes or later when we need to tweak dependencies). Can you try `pod install` with the updated `Podfile`, but the original `Podfile.lock`, please? It should update `Podfile.lock` and update it with the commit ID from the fork. If that doesn't work, we can go with updating all the dependencies as a fallback..

James-Sangalli:
Still seems shady, I am going to release a test build of this branch but won't merge it unless it works properly in testing with others.

PR #626, opened 2018-08-24
Replace ticket string references with tokens

hboon:
For #607

This PR:

* abstracts the token type name (e.g "ticket"). "tickets" and "cryptokitties" are hardcoded to be used for their appropriate addresses and will fallback to "tokens" for the rest. This sets it up so that in the future, we can read the type name from the XML before the fallback to "tokens".
* The ticket token is no longer always displayed below ETH in the wallet tab


hboon:
Fixed, rebased master.

PR #639, opened 2018-08-27
with nervos as web3js provider for dapp browser

James-Sangalli:
Seems to work with all functions in web3js, will keep as WIP until full testing and/or we find a way to fix the original


James-Sangalli:
@hboon seems to be working fine in dapp browser, safe to merge for now.

PR #654, opened 2018-08-29
Open all remote URLs in a new dapp browser instance

hboon:
Present a new one instead of using the one in the tab bar


hboon:
@James-Sangalli good for review

hboon:
Fixed broken tests. (Forced push, with rebase master)

Issue #648, opened 2018-08-28
Blacklist and gray list


colourful-land:
Some addresses should be protected

- the address denoted by a public key at the infinite point on elliptic curve can't have a private key.
- practically, 0x0000000 has no corresponding private key, because if someone can find the key for that, he is able to find the key for any given address, meaning hash + elliptic curve cryptography is broken.
- the same can be said to 0x1111....1111 and 0x999999.

Some organisations publicly announce that their private key is lost (e.g. Republic Of Conscience ICO) and any money donated to it is lost forever.

--

We might have a few grey levels but let's first have a hard blacklist where transactions simply can't go through.

James-Sangalli:
@colourful-land can check against all bytes being the same but we rely on 0x0000... for ether transfers to mark that they are not contracts.

PR #658, opened 2018-08-29
Check against null addresses

James-Sangalli:
for #648


AlphaWallet/Tokens/Types/TokenObject.swift
```diff
@@ -47,7 +47,10 @@ class TokenObject: Object {
     }
 
     var address: Address {
-        return Address(string: contract)!
+        if let ethAddress = Address(string: contract) {
+            return ethAddress
+        }
+        return EtherKeystore.current!.address
```
hboon: This part seems a bit strange. It's returning a wallet address as a fallback when the contract address is invalid?

AlphaWallet/Tokens/Types/TokenObject.swift
```diff
@@ -47,7 +47,10 @@ class TokenObject: Object {
     }
 
     var address: Address {
-        return Address(string: contract)!
+        if let ethAddress = Address(string: contract) {
+            return ethAddress
+        }
+        return Address(uncheckedAgainstNullAddress: Constants.nullAddress)!
```
hboon: Since the purpose of the PR is to make it safer by not allowing transactions with blacklisted addresses, returning `0x0` on an invalid address can break this promise. In addition, the original version `Address(string: contract)!` before this PR would crash with an invalid address whereas with this PR, it'll be taken as `0x0`.

How about if we do this:

1. Use `Address(uncheckedAgainstNullAddress: contract)!` or keep it as `Address(string: contract)!` here.
2. Have a PR in the future to make as much of the addresses that is now represented as `String`s and turn them into `Address`s so we can pass them around safely (knowing they are at least not blacklisted) and not need to check for nils or force unwrap (similar to what is happening here) after their initial creation? Might even be a good idea to have 2 distinct "address" classes for contract and wallets.. It's Swift after all, we can take advantage of the strong typing :)

What do you think?

AlphaWallet/Tokens/Types/TokenObject.swift
```diff
@@ -47,7 +47,7 @@ class TokenObject: Object {
     }
 
     var address: Address {
-        return Address(string: contract)!
+        return Address(uncheckedAgainstNullAddress: Constants.nullAddress)!
```
hboon: You'll probably catch it, but just in case I forgot:

`contract`

not:

`Constants.nullAddress`

James-Sangalli: Woops, I thought I put that!

PR #672, opened 2018-09-03
appends xml name onto contract

James-Sangalli:


James-Sangalli:
@hboon works but is later overriden by the xml only name..

Issue #664, opened 2018-08-31
Show only titles on Non XML ERC875 asset cards


James-Sangalli:
When an ERC875 token does not have an XML file, only show the title and blank out the other fields and icons which would normally have n/a

zhangzhongnan928:
![image](https://user-images.githubusercontent.com/33795543/44900694-dac5c200-ad37-11e8-8920-7d8bd821c3e9.png)

hboon:
@James-Sangalli @zhangzhongnan928 is it worth display the token's bytes32 value?

James-Sangalli:
@hboon not for now, also there would be an array of bytes32

hboon:
@James-Sangalli yeah, but we don't bundle them so there's only 1 per card.

PR #678, opened 2018-09-03
Show only title and quantity in ERC 875 token card if there's no asset definition

hboon:
For #664


James-Sangalli:
@hboon Excellent! Just missing it on import screen

hboon:
simulator screen shot - iphone x - 2018-09-04 at 10 49 59
simulator screen shot - iphone x - 2018-09-04 at 10 50 02

It shows the 1st screenshot for a while and then the 2nd after it gets the contract name — from the local datastore if available, otherwise after making an RPC call). The latter might take a while. Is that what you noticed? We can show the "token name" (from the XML) as a placeholder while/if it needs to make the RPC call.

hboon:
@James-Sangalli Ready for review again

Issue #674, opened 2018-09-03
Web3Swift (the pod) causing tests to stall


hboon:
Still trying to verify and fix this. But it is causing tests to hang on my (Boon's) development machine. After a few test cases which contains code that uses `Web3Swift` (the vendored version, not the pod) have run and completed, one of the test cases that also contains code which uses `Web3Swift` will hang. This is repeatable and always happens with `ClaimOrderCoordinatorTests` [1]. Every instance of `Web3Swift` runs Javascript in an instance of `WKWebView` to issue (HTTP) RPC calls. After some experimentation, I'm *guessing*:

1. This happens because `WKWebView` has a limit on the number of HTTP calls it can make at a single time.
2. `WKWebView` instances are cleaned up quickly enough, so latter instances are "blocked".
3. It doesn't affect production because users can't trigger that many calls that quickly (yet).
4. It happens only on faster machines. Hence James’ machine and Travis aren't affected (yet).

[1] Commenting out certain code/tests can move the point where it hangs to another test case which still uses `Web3Swift`, at that point it is repeatable too.

hboon:
Well, it turns out to _be_ the web3swift pod

PR #683, opened 2018-09-05
Fix: some tests stalling because of web3swift pod

hboon:
For #674 

This PR makes use of the Promise for making RPC calls instead of waiting for it, blocking the executing thread. This involves using a patched version of the `web3swift` pod to expose the Promise for making RPC calls.


hboon:
@James-Sangalli yeah, we have 2 commits in our fork — 1 for changing a dependency and 1 for this. I'll make a PR for this change.

hboon:
@James-Sangalli `master` on web3swift has a similar fix so we don't need to create PR.

PR #709, opened 2018-09-11
Fix: non-fungible tokens with empty symbol showing up in wallet

hboon:
For #710 

An earlier commit https://github.com/alpha-wallet/alpha-wallet-ios/commit/188e8f1d6f079c5115cf9822b169c1ab82f8c664 did this for tokens identified as ERC20 + empty symbols.

This PR does the same thing for non-fungible tokens.

In particular. https://ropsten.etherscan.io/address/0xd9161204f675200776a8f1BbCfAe5943cC6fe842 returns `true` for `isStormBirdContract()`.


James-Sangalli:
Not quite sure I understand, when the contract has no symbol it makes one?

hboon:
No, it has no name and no symbol. So it shows up as a single “0” or “2” for the balance. We had the same problem for delegate contracts. I’m presuming that this is happening for this ERC875 tokens because it was an incomplete test from the earlier days? Anyway good to have it for all unfungibles until it becomes something useful?

hboon:
Oh. And this PR hides such tokens. Should have said it the first time!

James-Sangalli:
Oh, technically tho, we should still be able to display empty name and symbol contracts as they are still valid

hboon:
@James-Sangalli We can do that. How should we display them? And which tokens do we display instead of hide (like the delegate contracts earlier)?

James-Sangalli:
@hboon for now we can use something like "N/A", the second question is tricky as it is hard to know, maybe we have to just let it go for now. 

These delegates often have symbol functions defined, maybe querying decimals will help sort? Decimals might also be tricky however as non fungibles often default to zero.

I guess for now, we just let it go.

Issue #692, opened 2018-09-07
Support building with Xcode 10


hboon:
* Upgrade Realm pod
    * pod 'RealmSwift', '~> 3.8'
* Get Trust-Keystore to build
    * In our fork, fix the build error by updating Secp256k1.swift, pubicKey(from:), move the "var len" out of the closure

PR #715, opened 2018-09-13
Update for Xcode10 iOS 12

hboon:
For #692 

This PR will make the project build with Xcode 10. Involves updating pod dependencies.

@James-Sangalli Only merge after you have verified this builds with Xcode 9.x or you have upgraded to Xcode 10. It breaks on my Xcode 9.x, but maybe because I also have Xcode 10 on the same machine.


hboon:
@James-Sangalli wonderful. Thanks!

PR #717, opened 2018-09-17
Fix: access to localization strings for Dapp browser was not going through R

hboon:
Missing the benefits of static type checking. Also easier to miss not localised strings. Like in this case.


hboon:
@James-Sangalli wonderful!

Issue #731, opened 2018-09-20
Upgrade infura to latest API spec 


James-Sangalli:

James-Sangalli:
https://community.infura.io/t/infura-responds-that-no-id-is-attached/424

James-Sangalli:
https://ethereum.stackexchange.com/questions/68637/infura-v3-spec-cannot-find-project-id

PR #732, opened 2018-09-20
with new api route for infura

James-Sangalli:
for #731


James-Sangalli:
@hboon hold up

James-Sangalli:
API is not stable yet, closing

PR #740, opened 2018-09-24
Remove unused code

hboon:


hboon:
@James-Sangalli in addition to the label, you can also use this if you like https://github.com/alpha-wallet/alpha-wallet-ios/pulls/review-requested/james-sangalli (constructed from the drop downs)

PR #755, opened 2018-09-26
Update Localizable.strings

bootingman:
update the untranslated strings


bootingman:
update the untranslated strings

PR #771, opened 2018-10-01
silenced notifications from SNS

James-Sangalli:


James-Sangalli:
Shelf this for now, just needed it for latest build

PR #806, opened 2018-10-08
Fix: ERC721 token detection and fetch balance

hboon:
This PR will:

* Detect correctly that a non-CryptoKitties ERC721 token is actually ERC721
* Fetch balance of ERC721 tokens correctly (when autodetecting or adding a custom token)

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -386,6 +386,7 @@ class InCoordinator: Coordinator {
         tokensCardCoordinator.start()
         switch (type, session.account.type) {
         case (.send, .real), (.request, _):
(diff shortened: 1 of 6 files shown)


James-Sangalli:
@hboon I understand that you can reliably get the token balance but how does it handle getting the token ids for transferring etc?

hboon:
@James-Sangalli we can't. This PR only gets the balance as a single number, 8 tokens (not even the token index). We still have to use OpenSea to get the balance (as in tokenID). But this means that if a contract implements ERC721 correctly, even if it's not in OpenSea, we will display them as an un-bundled list of "x1 Tokens" in the app.

James-Sangalli:
@hboon but then we need to blank out the transfer button?

hboon:
@James-Sangalli good for review. Have made it so it only enables the transfer button for ERC721 tokens if it's CryptoKitties.

Issue #768, opened 2018-09-29
Check DApp transaction gas limit


JamesSmartCell:
Most DApp transaction builders leave out gas limit. When you form the transaction normally gas limit = 0 gets built into the transaction. This always fails.

I notice metamask takes a '0' gas limit in the DApp transaction to mean that the wallet should use it's own default value.

You can test it on Tenzorum's TENZ-ID site here:

https://tenzorum.org/tenz_id/

try to register an ENS. You can test with testnet crypto - the gas limit gets added the same way. Even with testnet the website will report success.

I modified Android DApp browser to detect the zero gas limit and replace with the current gas limit setting.

James-Sangalli:
@JamesSmartCell are you suggesting that the wallet sometimes sends tx with no gas?

JamesSmartCell:
yes that exactly it. Victor checked and AlphaWallet iOS failed to push the transaction because it sends tx with zero gas.

The JS in the website should be fixed to include the gas limit, however, it's not important to be right (push the transaction with zero gas because the JS told me to) only that our wallet works when people are trying to do things.

It seems that Metamask will interpret zero gas limit with the wallet default. If user tries on our wallet it just fails, if they try on Metamask it works, the developer of the website will just recommend people to use Metamask only rather than fixing their site.

I'll try to find the forum post but Bitforward say exactly this when people are struggling to use their new ENS system - only metamask works and they don't know why. I left a post with how to fix their gas setup but they don't need to take any action because it works on Metamask.

James-Sangalli:
@JamesSmartCell @zhangzhongnan928 it seems that they have fixed this because it is working on my side (See screenshot below) 
![unnamed 1](https://user-images.githubusercontent.com/16630514/46287709-01388080-c5b6-11e8-9ba7-0248afde0269.jpg)

zhangzhongnan928:
@JamesSmartCell @James-Sangalli > AlphaWallet iOS failed to push the transaction because it sends tx with zero gas.

Not because of zero gas, because not enough gas, by default iOS provide 90000 gas limit, but transection cost more than 140000 gas, 
so the first transection failed:
https://etherscan.io/tx/0x77ab9d892f6ed32eb486d8957960a591dde2c2b92f208f09f9d6a11d0f07096b

the 2nd transection, I doubled the gas limit and it went through
[...]

PR #811, opened 2018-10-15
remove gas related Constants and use max for gas limit

James-Sangalli:
for #768

Gas is already being estimated but might fail sometimes when the gas price exceeds the limit of default. Removed the default and used the max for gas limit, also removed the Constants that were duplicate.

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -589,7 +589,6 @@ extension InCoordinator: TokensCoordinatorDelegate {
     // transfer from a buddy), the user can send the data to a paymaster.
     // This function deal with the special case that the token price = 0
     // but not sent to the paymaster because the user has ether.
-
     func importPaidSignedOrder(signedOrder: SignedOrder, tokenObject: TokenObject, completion: @escaping (Bool) -> Void) {
         let web3 = self.web3()
         web3.start()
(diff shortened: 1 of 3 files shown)


AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -52,7 +52,7 @@ class TransactionConfigurator {
 
         self.configuration = TransactionConfiguration(
             gasPrice: min(max(transaction.gasPrice ?? GasPriceConfiguration.default, GasPriceConfiguration.min), GasPriceConfiguration.max),
-            gasLimit: transaction.gasLimit ?? GasLimitConfiguration.default,
+            gasLimit: transaction.gasLimit ?? GasLimitConfiguration.max,
```
hboon: Looks like `GasLimitConfiguration.default` isn't used anymore. Shall we remove it?

AlphaWallet/InCoordinator.swift
```diff
@@ -613,9 +612,9 @@ extension InCoordinator: TokensCoordinatorDelegate {
                         value: BigInt(signedOrder.order.price),
                         to: address,
                         data: Data(bytes: payload.hexa2Bytes),
-                        gasLimit: Constants.gasLimit,
+                        gasLimit: GasLimitConfiguration.max,
                         tokenId: .none,
-                        gasPrice: Constants.gasPriceDefaultERC875,
+                        gasPrice: GasLimitConfiguration.default,
```
hboon: GasLimitConfiguration.default (`90,000`) is much lower than Constants.gasPriceDefaultERC875 was. Is it enough?

James-Sangalli: The other one was arbitrarily high, gasPrice is generic, therefore the gas price should not distinguish between erc875 transactions and any other

Issue #786, opened 2018-10-03
Security alert SNS


colourful-land:
1. iOS app should automatically acquire an SNS endpoint (register if non-existed, enable if it is disabled).

2. iOS app should subscribe to the security SNS topic (fill in the SNS topic ID here) if not already.

3. iOS app should display an alert if the current version matches the alert's version. (We will leave the matching rule to the later stage).

James-Sangalli:
@colourful-land started [here](https://github.com/AlphaWallet/alpha-wallet-ios/tree/sns-experiments)

colourful-land:
Deleted Nervos Project 𝑙𝑎𝑏𝑒𝑙 because now it is managed by Nervos Project 𝑝𝑟𝑜𝑗𝑒𝑐𝑡.

PR #814, opened 2018-10-16
Sns experiments

James-Sangalli:
For #786


AlphaWallet/AppDelegate.swift
```diff
@@ -123,12 +124,37 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
                 if let endpointArnForSNS = createEndpointResponse.endpointArn {
                     print("endpointArn: \(endpointArnForSNS)")
                     UserDefaults.standard.set(endpointArnForSNS, forKey: "endpointArnForSNS")
+                    //every user should subscribe to the security topic
+                    self.subscribeToTopicSNS(token: token, topicEndpoint: self.SNSSecurityTopicEndpoint)
+                    if let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
+                        //TODO subscribe to version topic when created
```
hboon: We can perform filtering for version numbers when the app receive a silent notification (though it's not perfect), but it should help to reduce some housekeeping work for creating and maintaining a topic for each version.

Worth a refactor to reduce code added to the AppDelegate though. Maybe later.

Issue #780, opened 2018-10-02
Support all collectables in OpenSea.io


James-Sangalli:
Can make just a rudimentary version first that simply parses the text and shows it to the user

PR #815, opened 2018-10-17
Support all tokens from OpenSea API output

hboon:
For #780.

This PR

* Makes the wallet tab display all tokens supported by OpenSea
* Adds some special handling for displaying more popular tokens like HyperDragons, MyCryptoHeroes, MLBCB, CryptoSaga
* Renames most types and symbols which refer to "cryptokitties" to "opensea non-fungible tokens"
* Throttles calls and cache results from OpenSea for wallets that hold many token (IDs) — A JSON "record" is about 4kb so 1500 HyperDragons is 6MB. Throttling starts when the total number of token IDs in a wallet is 25 and when that happens, OpenSea results are cached for 1 minute. Note that 1 OpenSea call is made for each wallet, not for each contract.

1. It's bit too slow if we are showing many token IDs for a token, e.g. tapping HyperDragons in the wallet tab and waiting for the list of 1500 HyperDragons to show up takes ~4 seconds. Will see if we can improve this in a separate PR. Probably an edge case though.
2. Looks like we can create a separate tab (alongside ALL, CURRENCY and ASSETS) in wallet called COLLECTIBLES displaying tokens returned by OpenSea (which means ERC721 for now). We *might* be able to have a nicer display then just a list because we can probably display an icon for each one. What do you think @james-sangalli?
3. Is transfer supposed to be working for all the tokens on OpenSea? Shall we disable transfer for all of them and only enable for those that we manually verify works @james-sangalli?

Sidenote: I'll work on a PR to change how transactions are fetched. Otherwise the app stalls when testing with wallets that have many collectibles (because they naturally have many transactions).


James-Sangalli:
@hboon looks great! 

Transfer is standardised in erc721, if opensea can return the token ids then it shouldn't be an issue support transfer for all the collectables in opensea.

PR #863, opened 2018-10-28
Fix: import of ERC875 via magic link for legacy contracts broken

hboon:


hboon:
@James-Sangalli validation for my rule that if there's an `if`, look for the missing `else` — except I violated it :(

Issue #880, opened 2018-11-07
(iOS) XML to have TBML namespace


colourful-land:
Test case

1. Copy the following 3 XML files from tbml branch† to mobile phone's storage area, replacing the existing XML files, 

    ./spawnable-contract/schema1/MeetupContract.xml
    ./blockchain-tickets/schema1/blockchain-events-china.xml
    ./blockchain-tickets/schema1/TicketingContract.xml

† https://github.com/AlphaWallet/contracts/tree/tbml

2. Make sure the mobile work as before.

Notice that the same namespace is represented in 3 different shorthands in order to make sure the mobile code does not blindly assume any shorthand.

hboon:
1.26 is the current version we have. So > 1.26 should support it.

Can test with something like: 

```
curl -H "Accept: text/xml; charset=UTF-8" -H "IF-Modified-Since: Mon, 02 Jul 2018 13:23:00 GMT" -H "X-Client-Name: AlphaWallet" -H "X-Client-Version: 1.0.3" -H "X-Platform-Name: iOS" -H "X-Platform-Version: 11.1.2" https://repo.awallet.io/0xd8e5f58de3933e1e35f9c65eb72cb188674624f3
```

hboon:
We should probably standardise our version number format for iOS and Android. Maybe not as far as apply https://semver.org, but at least agree that it's in the format `x.y(.z)?` Helps with parsing on the server side.

colourful-land:
> We should probably standardise our version number format for iOS and Android. Maybe not as far as apply https://semver.org, but at least agree that it's in the format `x.y(.z)?` Helps with parsing on the server side.

Can you file a new issue about that in https://github.com/AlphaWallet/alpha-wallet-android/issues/ where repo server code lies?

colourful-land:
Since we are abandoning awallet.io we will use the method @James-Sangalli suggested:

New builds use repo.attestation.id
Old builds use repo.awallet.io

instead of version numbering.

hboon:
@colourful-land is the new repo at `repo.attestation.id` up?

James-Sangalli:
Keeping issue open until it is working with new server

colourful-land:
> @colourful-land is the new repo at `repo.attestation.id` up?

repo.aw.app is online now. Its XML, however, has not been fully updated to allow ABI embedded in XML, see https://github.com/AlphaWallet/contracts/issues/70

[...]

Issue #869, opened 2018-10-31
 (iOS) support the standard spawnable meetup contract 


colourful-land:
In the standard Peter's contract (the spawnable contract)'s XML there are some fields whose value is obtained through smart contract calls, not embedded in the `tokenID`. These must be displayed correctly and refreshed at a suitable frequency lest the contract changes its internal state.

The standard xml is MeetupContract.xml

 https://github.com/AlphaWallet/contracts/blob/master/spawnable-contract/schema1/MeetupContract.xml

The parts that need to be displayed on the UI are the following (observe closely the `<function>` calls)

    <attribute-type id="expired" syntax="1.3.6.1.4.1.1466.115.121.1.7">
      <origin contract="holding-contract">
        <function>isExpired</function>
      </origin>
    </attribute-type>
    <attribute-type id="street" syntax="1.3.6.1.4.1.1466.115.121.1.7">
      <origin contract="holding-contract">
        <function>getStreet</function>
        <parameter>TokenID</parameter>
      </origin>
    </attribute-type>
    <attribute-type id="building" syntax="1.3.6.1.4.1.1466.115.121.1.7">
      <origin contract="holding-contract">
        <function>getBuildingName</function>
        <parameter>TokenID</parameter>
      </origin>
    </attribute-type>
    <attribute-type id="state" syntax="1.3.6.1.4.1.1466.115.121.1.7">
      <origin contract="holding-contract">
        <function>getState</function>
        <parameter>TokenID</parameter>
      </origin>
    </attribute-type>

Refer to https://github.com/AlphaWallet/alpha-wallet-android/issues/375 for progress in Android

PR #882, opened 2018-11-08
Support namespaces in asset definitions

hboon:
For #880.

This is a poor man's handling of XML namespaces: the current iOS parser library we use has no knowledge about namespaces, so depending on what we use for XML validation, we might need and probably should switch to a different parser later.

Can test this PR by tapping on a token and then AirDrop (or use Finder if simulator) to override the XML. Add, remove, change the XML namespace prefix in the entire XML and the screen should refresh automatically.

Update: this PR has not switched the repo server to `repo.attestation.id` yet because it doesn't seem to be up. But it supports both the old and new repo's XMLs.


hboon:

PR #884, opened 2018-11-08
Call contract functions for asset attributes for spawnable meetup contract

hboon:
@James-Sangalli: This PR includes the commit in PR #882 in order to support namespaces, so #882 
should be merged first and this PR rebased (or updated) before merging.

@colourful-land you might be interested in: (A), (C) and (D).

This PR adds support for the standard spawnable meetup contract:

1. Calls smart contract functions to retrieve asset attribute values
2. Display most fields in the spawnable meetup contract

To add these quickly, there are a few limitations that should be fixed or improved in the near future:

1. The function-call-based asset attributes are only refreshed when the app is launched or becomes active
2. The token bundles displayed on the current screen don't automatically refresh when the function-call-based asset attributes changes. It does reflect the updated values if you go back and display the screen again. (doesn't affect the spawnable meetup contract because bundling is disabled)
3. Function calls aren't throttled
4. It doesn't reload values immediately when the asset definition changes
5. Bundling is disabled for the meetup contract

I noticed a few things:

A. The signature includes this: `<ds:KeyName>Shankai</ds:KeyName>` so the token card displays `Shankai` as the issuer
B. The static attributes stored in the token IDs in the contract are all 0-values except `section`
C. I've changed `category` and `section` to syntax="1.3.6.1.4.1.1466.115.121.1.27" to get them to be parsed correctly based on the value extracted from the token IDs. @James-Sangalli This might not be correct because we might have intended for the token IDs to have encode the category and section as ascii instead. Especially because the bitmask looks quite long for both category and section
D. The asset definition in the old repo (I couldn't access the new repo) is outdated. I've tested by modifying the function-based attributes to (note that the function name of the deployed code is `getBuilding`, not `getBuildingName`):

```
    <attribute-type id="expired" syntax="1.3.6.1.4.1.1466.115.121.1.7"> 
      <origin contract="holding-contract"> 
        <function>isExpired(<value ref="tokenID">uint256</value>)</function> 
      </origin> 
    </attribute-type> 
    <attribute-type id="street" syntax="1.3.6.1.4.1.1466.115.121.1.15"> 
      <origin contract="holding-contract"> 
        <function>getStreet(<value ref="tokenID">uint256</value>)</function> 
      </origin> 
    </attribute-type> 
    <attribute-type id="building" syntax="1.3.6.1.4.1.1466.115.121.1.15"> 
      <origin contract="holding-contract"> 
        <function>getBuilding(<value ref="tokenID">uint256</value>)</function> 
      </origin> 
    </attribute-type> 
    <attribute-type id="state" syntax="1.3.6.1.4.1.1466.115.121.1.15"> 
      <origin contract="holding-contract"> 
        <function>getState(<value ref="tokenID">uint256</value>)</function> 
      </origin> 
    </attribute-type> 
```

Output like this:

simulator screen shot - iphone x - 2018-11-09 at 06 48 18

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -287,6 +287,7 @@
 		5E7C733638D7596F93DEE2A9 /* OnboardingCollectionViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C75CE3F1D6B7993E7A840 /* OnboardingCollectionViewController.swift */; };
 		5E7C7364BF4F6EC3F15804C2 /* GetIsERC721Encode.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C7CA7D65743AEE3411F3A /* GetIsERC721Encode.swift */; };
 		5E7C7376B566E5A59CC8F463 /* ImportMagicTokenViewControllerViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C72D0E7CA03ADE5CFAE7A /* ImportMagicTokenViewControllerViewModel.swift */; };
+		5E7C73CA81FB2CE9BCAFC992 /* CallForAssetAttribute.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C787CA216AFED8023A35F /* CallForAssetAttribute.swift */; };
 		5E7C73FC3990D110C474C3D6 /* WalletFilterViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C75CC640BAFFE0E789F44 /* WalletFilterViewModel.swift */; };
 		5E7C73FD5BD75D90C8D0EF3C /* WalletFilterView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C7C58586099F082973073 /* WalletFilterView.swift */; };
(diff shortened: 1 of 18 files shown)


hboon:
> Also note that the buildingName thing has been fixed in the latest contract code

@James-Sangalli So we'll have to be careful with the contract + xml combination.

hboon:
@James-Sangalli will you be doing the rebase/merge or do you want me to do it? There were no changes to #882 , so both rebase or merge should be clean.

colourful-land:
Thanks. Noted. This can't merge until I get the new repo server working and tested against it (I'm getting the repo server to be namespace-𝑎𝑤𝑎𝑟𝑒).

The syntax for ABI was experimental and will change as I have a grasp of the full possible dataset in ABI.

I'm behind the schedule for both items. Self-blaming...

James-Sangalli:
@hboon it has been merged

hboon:
> Thanks. Noted. This can't merge until I get the new repo server working and tested against it (I'm getting the repo server to be namespace-𝑎𝑤𝑎𝑟𝑒).

> The syntax for ABI was experimental and will change as I have a grasp of the full possible dataset in ABI.

Ok. But just writing this in case the PR needs to be merged earlier for some reason:

We can still adjust the code to accept a different format for the ABI in XML after the merge. It should be a minor change.

The PR works with both default namespace and explicit namespace.

Then we can do a follow up with another PR to switch to the new repo server.

James-Sangalli:
Gonna merge this, can make a seperate PR to account for any changes.

hboon:

Issue #810, opened 2018-10-12
Shall we support ENS in AlphaWallet?


zhangzhongnan928:
It seems to be an better UX service. 

This issue is considered solved if the first of the following is satisfied and the 2nd. Victor will create new issue for 3rd, 4th if/when he sees the needs.

1) when a user sends token or ether to another, he or she can type the ENS instead of an address.
2) when a user goes to his wallet settings, he can see his own wallet labelled in ENS fashion.
3) when a user receives a token or ether it is displayed as the ENS if applicable.
4) when a user goes to his wallet asset list, he can see his name as an asset.
5) when a user wants to get an ENS for his wallet he can do so by acoomplishing a mobile journey.

dtbx:
Agree, would help using services like [tenzID](tenzorum.org/tenz_id) to transact between account.

zhangzhongnan928:
@hboon @JamesSmartCell

zhangzhongnan928:
@hboon @JamesSmartCell 
Seems to be a nice function. 
Any feedbacks?

hboon:
Yes, sounds good. We should also do reverse resolutions and display the wallet names in Settings > Wallets.

Any concerns, @colourful-land ?

colourful-land:
@hboon Yes. There are a lot of user experience issues that need to be considered. e.g. should we display ENS over 0x..., should we copy 0x... when the user holds on ENS, should we list ENS as an asset (if it can be sold and bought by detaching and reattaching to a different name), are other wallets supporting it too. Might need to get @farrahfazirah involved. Updated the main description to provide clarity.

James-Sangalli:
@hboon working on a simple implementation that will allow the user to simply enter an ENS name instead of an address

zhangzhongnan928:
Stage 1 for ENS,  don’t change any UI/UX, just allow user to send through ENS or wallet address. 
All the other flow remain the same. No design required. 





		
Victor Zhang
CEO, AlphaWallet
							
A 1000 DApps in your pocket — Why a wallet is so much more than an app to hold keys by James Sangalli https://t.co/BNjaA6ob2e	Read More

zhangzhongnan928:
![37c9f1b1-c166-4468-befe-cd184fef3d29](https://user-images.githubusercontent.com/33795543/48189923-405bad80-e37c-11e8-9362-e37650d63c7d.jpeg)

[...]

PR #887, opened 2018-11-09
ENS compatibility

James-Sangalli:
For #810 

Allows you to write/paste any valid ENS name into the box and it will be translated into the corresponding address

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
@@ -40,7 +40,7 @@ class GetENSOwnerCoordinator {
         }
 
         //if it does not contain .eth, then it is not a valid ens name
-        if !input.contains(".eth") {
+        if !input.contains(".") {
             completion(.failure(AnyError(Web3Error(description: "Invalid ENS Name"))))
             return
         }
--- a/AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
@@ -69,7 +69,11 @@ class GetENSOwnerCoordinator {
             //if null address is returned (as 0) we count it as invalid
             //this is because it is not assigned to an ENS and puts the user in danger of sending funds to null
             if let owner = result["0"] as? EthereumAddress {
-                completion(.success(owner))
+                if owner.address == Constants.nullAddress {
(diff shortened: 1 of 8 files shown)


AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+
+        //if already an address, send back the address
+        if let ethAddress = EthereumAddress(input) {
+            completion(.success(ethAddress))
+            return
+        }
+
+        //if it does not contain .eth, then it is not a valid ens name
+        if !input.contains(".eth") {
```
hboon: We can use `hasSuffix(".eth")` instead. But it appears that there's also .xyz and .luxe. Maybe we can't perform this check reliably?

AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift
```diff
@@ -0,0 +1,90 @@
...
+        guard let promise = contractInstance.method(function.name, parameters: [node] as [AnyObject], options: nil) else {
+            completion(.failure(AnyError(Web3Error(description: "Error calling \(function.name)() on \(contractAddress.address)"))))
+            return
+        }
+
+        promise.callPromise(options: nil).done { result in
+            //if null address is returned (as 0) we count it as invalid
+            //this is because it is not assigned to an ENS and puts the user in danger of sending funds to null
+            if let owner = result["0"] as? EthereumAddress {
+                completion(.success(owner))
```
hboon: If it fails to resolve, the textbox displays "0x0000000000000000000000000000000000000000". Maybe we can hardcode a check for 0x0 and don't display it?

James-Sangalli:
@hboon ok.

Actually, it is better to simply check for "." as any text input that has a dot in it can be implied to be an ENS name, it is also compatible with subdomains. User is protected because an invalid name will still return a null value.

James-Sangalli:
@hboon 2 is not an issue, it is able to detect the ens name if you use the black and white paste button

hboon:
@James-Sangalli I was referring to this white on black Paste button. Does it trigger ENS resolution?

simulator screen shot - iphone 6s - 2018-11-10 at 22 04 34

hboon:
Another future enhancement when we get the Ethereum name and address to show up together: handle when user types in an ENS name, maybe start resolving say, 0.5 seconds after user stops typing.

PR #893, opened 2018-11-11
Update XML parsing to handle latest asset definition schema

hboon:
This PR updates XML parsing the handle the asset definitions in https://github.com/AlphaWallet/contracts/pull/86/commits/c6a1ad04ff286b532220866be9eb4035fb99043b

I tested this by downloading the 3 XMLs into the `assetDefinitionsOverrides/` directory and then renaming them to the appropriate contracts.

Might be good to update the contract (I tested against `0x16e58F9a98bc46a572CFb9ce28DDEeaf76AF645a`) based on the spawnable meetup contract XML to have a couple more tokens with various fields encoded in the token ID to see if that works.

Might have to wait till https://github.com/AlphaWallet/contracts/pull/86/commits/c6a1ad04ff286b532220866be9eb4035fb99043b is finalised to merge. cc @colourful-land


hboon:

PR #895, opened 2018-11-12
With ens label

James-Sangalli:
Should perhaps also show the label when the user leaves the text box and switches to amount? A better idea?


James-Sangalli:
@hboon A is already done in this PR, if pasting a normal address you won't see anything in the label

B is also basically done but it doesn't have a set point to start resolving, perhaps we wait a few seconds after . is push, to prevent lagging?

hboon:
Right. A is how it used to work with addresses and it still works the same way with this PR.

hboon:
@James-Sangalli good for review.

PR #899, opened 2018-11-13
Don't auto scroll token asset cards when expanding non-OpenSea supported tokens

hboon:
Before this PR, tapping on a token asset card to expand it will auto-scroll that card to the top of the list.

With this PR, we only auto scroll to reveal for OpenSea-supported tokens which are usually taller and have a picture. Because:

A. other tokens like ERC875 tickets are usually too short and all text, making it difficult for user to capture where it has scrolled to
B. OpenSea-supported tokens are tall, so after expanding, chances are user need to scroll quite a lot if we don't auto-scroll


hboon:
Pending #900 and rebase master.

hboon:
Ready for review

PR #900, opened 2018-11-13
Fix broken tests on Travis (and switching to using a fork of TrezorCrypto pod for more control)

hboon:
The author deleted the version (0.0.8) of the pod we are using (really uncommmon and shouldn't be done unless something horrible happened. But every old version was removed..). This situation is bad because we were forced to update the version of the pod we use. It also broke Travis, which couldn't find the pod.

This PR:

1. Forks the repo and use it instead
2. Increases the version from 0.0.8 to 0.0.9 anyway. Had wanted to stick with 0.0.8 with the fork, but can't find the commit (I suppose the pod author must have forced push at some point so pod v0.0.8 and pod v0.0.8 in the repo wasn't exactly the same). We can still use 0.0.8 in the fork, but since both 0.0.8 (in the repo) and 0.0.9 wasn't tested against previously, might as well up version from 0.0.8 -> 0.0.9

With the added maintenance of the fork (we might choose to just pull from upstream and make no changes, then minimal work), at least this wouldn't happen again.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -24,6 +24,7 @@ target 'AlphaWallet' do
   pod 'SwiftyXMLParser', :git => 'https://github.com/yahoojapan/SwiftyXMLParser.git'
   pod 'Kingfisher', '~> 4.0'
   pod 'TrustWeb3Provider', :git=>'https://github.com/alpha-wallet/dapp-web3-provider', :branch => 'nervos-dist'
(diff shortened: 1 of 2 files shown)


hboon:
Maybe. Then it’s in the Trezor library. The pod is just to publish the library into a pod.

PR #909, opened 2018-11-16
Don't (offer to) autocorrect or autocapitalize in address textfields.

hboon:
Especially since we allow entering of ENS names now.


hboon:
@James-Sangalli always good to have a little competition

PR #927, opened 2018-11-29
add functions to allow for spawnable token transactions

James-Sangalli:
Before this PR we could not accept spawnable links from Peter's tool because it checked for a balance which does not exist for spawnable tokens. The function calls to the contract were also out of date and incompatible. 

Note: we seem to have an issue with exceeding gas limits for magic links. I noticed this when importing both a spawnable and non spawnable token link...

### DIFF
--- a/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
@@ -147,10 +147,16 @@ class UniversalLinkCoordinator: Coordinator {
                     return
                 }
                 //filter null tokens
-                let filteredTokens = strongSelf.checkERC875TokensAreAvailable(
+                var filteredTokens = strongSelf.checkERC875TokensAreAvailable(
                         indices: signedOrder.order.indices,
                         balance: balance
                 )
+
+                //if the token is spawnable then let it continue
+                if signedOrder.order.spawnable, let tokens = signedOrder.order.tokenIds {
+                    filteredTokens.append(tokens[0].description)
+                }
+
                 if filteredTokens.isEmpty {
                     strongSelf.showImportError(errorMessage: R.string.localizable.aClaimTokenInvalidLinkTryAgain())
                     return
--- a/AlphaWallet/RPC/Commands/web3wift-web3js/ClaimERC875Order.swift
+++ b/AlphaWallet/RPC/Commands/web3wift-web3js/ClaimERC875Order.swift
@@ -43,14 +43,15 @@ struct ClaimERC875Order: Web3Request {
 
 struct ClaimERC875Spawnable: Web3Request {
     typealias Response = String
-    let tokenIds: [BigUInt]
+    let tokenIds: [String]
     let v: UInt8
     let r: String
     let s: String
     let expiry: BigUInt
+    let recipient: String
 
     var type: Web3RequestType {
-        let abi = "{\"constant\":false,\"inputs\":[{\"name\":\"expiry\",\"type\":\"uint256\"},{\"name\":\"tokenIds\",\"type\":\"uint256[]\"},{\"name\":\"v\",\"type\":\"uint8\"},{\"name\":\"r\",\"type\":\"uint256\"},{\"name\":\"s\",\"type\":\"uint256\"}],\"name\":\"passToSpawnable\",\"outputs\":[],\"payable\":true,\"stateMutability\":\"payable\",\"type\":\"function\"}, [\"\(expiry)\", \(tokenIds), \(v), \"\(r)\", \"\(s)\"]"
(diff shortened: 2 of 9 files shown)


AlphaWallet/Tokens/Coordinators/ClaimOrderCoordinator.swift
```diff
@@ -66,8 +67,15 @@ class ClaimOrderCoordinator {
                              v: UInt8,
                              r: String,
                              s: String,
+                             recipient: String,
                              completion: @escaping (Result<String, AnyError>) -> Void) {
-        let request = ClaimERC875Spawnable(tokenIds: tokenIds, v: v, r: r, s: s, expiry: expiry)
+        var tokenStrings = [String]()
```
hboon: This is a bit shorter:

```
var tokenStrings: [String] = tokenIds.map { MarketQueueHandler.bytesToHexa($0.serialize().array) }
```

But this version is terser and clearer if that's what you are trying to do:

```
var tokenStrings: [String] = tokenIds.map { String(tokenIds[0], radix: 16) }
```

Maybe it's better to just pass tokenIds directly to ClaimERC875Spawnable (like before this PR) and move the conversion to [String] there instead? Based on the comment "conversion from BigUInt fails in abi encoding", marshalling the arguments is its responsibility

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -147,10 +147,16 @@ class UniversalLinkCoordinator: Coordinator {
                     return
                 }
                 //filter null tokens
-                let filteredTokens = strongSelf.checkERC875TokensAreAvailable(
+                var filteredTokens = strongSelf.checkERC875TokensAreAvailable(
```
hboon: We can change the `if` to an `if-else` and keep the `filteredTokens` as a `let` like this:

```
let filteredTokens: [String]
//if the token is spawnable, we don't check against the balance
if signedOrder.order.spawnable, let tokens = signedOrder.order.tokenIds {
    filteredTokens = [tokens[0].description]
} else {
    filteredTokens = strongSelf.checkERC875TokensAreAvailable(
            indices: signedOrder.order.indices,
            balance: balance
    )
}
```

I often break it myself, but I like this rule of thumb: an `if-else` is better than an `if` without `else` because it's always easy to miss out something if there's just a dangling `if` unless it contains an early return

But should this line:

```
filteredTokens = [tokens[0].description]
```

be this instead?

```
filteredTokens = tokens.map { $0.description }
```

And since the balance don't have to checked for spawnable tokens, we can re-organize the code like this:

```
if signedOrder.order.spawnable, let tokens = signedOrder.order.tokenIds {
    foo() //extracted method mentioned below
} else {
    //like it used to be, indented 1 more level now
    getERC875TokenBalanceCoordinator = GetERC875BalanceCoordinator(config: config)
    getERC875TokenBalanceCoordinator?.getERC875TokenBalance(for: recoverAddress, contract: contractAsAddress) { [weak self] result in
        //like it used to be, but extract the code starting from `if filteredTokens.isEmpty {` into a function foo() and call it
    }
}
```

So it'll actually be slightly faster if we redeem a spawnable link, skipping 1 RPC call

James-Sangalli:
@hboon should be good now but my brain is tired haha

AlphaWallet/RPC/Commands/web3wift-web3js/ClaimERC875Order.swift
```diff
@@ -43,15 +43,17 @@ struct ClaimERC875Order: Web3Request {
...
     let v: UInt8
     let r: String
     let s: String
     let expiry: BigUInt
     let recipient: String
 
     var type: Web3RequestType {
-        let abi = "{ \"constant\": false, \"inputs\": [ { \"name\": \"expiry\", \"type\": \"uint256\" }, { \"name\": \"tickets\", \"type\": \"uint256[]\" }, { \"name\": \"v\", \"type\": \"uint8\" }, { \"name\": \"r\", \"type\": \"bytes32\" }, { \"name\": \"s\", \"type\": \"bytes32\" }, { \"name\": \"recipient\", \"type\": \"address\" } ], \"name\": \"spawnPassTo\", \"outputs\": [], \"payable\": false, \"stateMutability\": \"nonpayable\", \"type\": \"function\" }, [\"\(expiry)\", \(tokenIds), \(v), \"\(r)\", \"\(s)\", \"\(recipient)\"]"
+        //BigUInt is cast incorrectly, must be converted to hex string values
+        let tokenStrings: [String] = tokenIds.map { _ in String(tokenIds[0], radix: 16) }
```
hboon:
```
let tokenStrings: [String] = tokenIds.map { _ in String($0, radix: 16) }
```

James-Sangalli:
@hboon should be good now

hboon:
@James-Sangalli can we remove the 2nd arg (indices) for makeTokenHolder()?

James-Sangalli:
@JamesSmartCell this might be relevant to you as I noticed that Peter's links do not work on Android.

PR #930, opened 2018-11-29
change gas limits back to how they should be

James-Sangalli:
For #928


James-Sangalli:
@hboon it should be consistent in that regard.

PR #931, opened 2018-11-29
Show "Done" instead of "Detecting token type" as button title in add custom token screen when it's first shown

hboon:


James-Sangalli:
@hboon build is failing

PR #936, opened 2018-11-30
Fix gas amounts

James-Sangalli:
For #928 

This PR also makes the distinction more clear with more verbose variable names and a note at the top of each struct.


hboon:
I pushed a commit. Can you help fix the tests?

I see what you were referring to with the 90_000 now :)

GasLimitConfiguration.defaultGasLimit (90_000) was only used in InCoordinator, line 423. Which was wrong. It should be using GasPriceConfiguration (it's one of the things the commit fixes). The reason why it seemed to work was because the only place where this value is used is max(x, y) [1] and because 90_000 wei is too low for gas price, it was always dropped and ignored.

[1] In TransactionConfigurator.init(). line 54: `gasPrice: min(max(transaction.gasPrice ?? GasPriceConfiguration.defaultPrice, GasPriceConfiguration.minPrice), GasPriceConfiguration.maxPrice),`

James-Sangalli:
@hboon fixed, thanks

Issue #945, opened 2018-12-02
Mokens 721 contract transfer function failed 


zhangzhongnan928:
https://etherscan.io/tx/0x6f19923ad2255d537685af71879ff4a77e9cb068c30949878685c528016f8e4f

https://etherscan.io/tx/0x6c8cf19b9fd98e71c097f722492939211a53dc96d7187984cc47daf9ddabe439

Tried 2 twice
https://mokens.io/moken/121

hboon:
@James-Sangalli it's https://github.com/AlphaWallet/alpha-wallet-ios/issues/830:

CryptoKitties:

    function transfer(address _to, uint256 _tokenId) external; //ultimately calls _transfer()
    function transferFrom(address _from, address _to, uint256 _tokenId) external; //ultimately calls _transfer()
    function _transfer(address _receiver, uint256 _tokenId) internal;

ERC721:

    function safeTransferFrom(address _from, address _to, uint256 _tokenId, bytes data) external payable;
    function safeTransferFrom(address _from, address _to, uint256 _tokenId) external payable;
    function transferFrom(address _from, address _to, uint256 _tokenId) external payable;

So for ERC721, looks like we should call `safeTransferFrom(address _from, address _to, uint256 _tokenId)` and not `transfer(address _to, uint256 _tokenId)` (also note the missing "From" in the function name), but since CryptoKitties do not implement `safeTransferFrom()`, we'll have to fallback to `transferFrom(address _from, address _to, uint256 _tokenId)` if it `safeTransferFrom()` fails..

PS: EMON/CryptoSaga/ChibiFighters don't implement `safeTransferFrom()` too. I guess they implemented ERC721 fairly early.

James-Sangalli:
@hboon ok, can implement it as default and call the fallback if it fails

hboon:
@James-Sangalli 

These are the ones we know that don't have safeTransferFrom(), but have transfer(address,uint256):

* kitty - 0x06012c8cf97bead5deae237070f9587f8e7a266d 
* CryptoSaga - 0xabc7e6c01237e8eef355bba2bf925a730b714d5f 
* ChibiFighters  - 0x71c118b00759b0851785642541ceb0f4ceea0bd5

PR #948, opened 2018-12-03
default to safeTransferFrom unless contract is legacy 

James-Sangalli:
@hboon seems that 721 does not specify a transfer function... 

for #945

### DIFF
--- a/AlphaWallet/RPC/Commands/web3wift-web3js/ContractERC721Transfer.swift
+++ b/AlphaWallet/RPC/Commands/web3wift-web3js/ContractERC721Transfer.swift
@@ -16,7 +16,7 @@ struct ContractERC721Transfer: Web3Request {
 
     var type: Web3RequestType {
         let abiLegacyTransfer = "{\"constant\":false,\"inputs\":[{\"name\":\"_to\",\"type\":\"address\"},{\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"transfer\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}, [\"\(to)\", \(tokenId)]"
-        let abiSafeTransferFrom = "{ \"constant\": false, \"inputs\": [ { \"name\": \"_from\", \"type\": \"address\" }, { \"name\": \"_to\", \"type\": \"address\" }, { \"name\": \"_tokenId\", \"type\": \"uint256\" } ], \"name\": \"safeTransferFrom\", \"outputs\": [], \"payable\": true, \"stateMutability\": \"payable\", \"type\": \"function\"},  [\"\(from)\", \(to), \(tokenId)]"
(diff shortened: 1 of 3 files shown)


AlphaWallet/RPC/Commands/web3wift-web3js/ContractERC721Transfer.swift
```diff
@@ -8,13 +8,18 @@ import BigInt
...
+    let from: String
+    let to: String
     let tokenId: String
+    let contractAddress: String
 
     var type: Web3RequestType {
-        let abi = "{\"constant\":false,\"inputs\":[{\"name\":\"_to\",\"type\":\"address\"},{\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"transfer\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}, [\"\(address)\", \(tokenId)]"
+        var abi = "{\"constant\":false,\"inputs\":[{\"name\":\"_to\",\"type\":\"address\"},{\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"transfer\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}, [\"\(to)\" \(tokenId)]"
+        let abiSafeTransferTo = "{ \"constant\": false, \"inputs\": [ { \"name\": \"_from\", \"type\": \"address\" }, { \"name\": \"_to\", \"type\": \"address\" }, { \"name\": \"_tokenId\", \"type\": \"uint256\" } ], \"name\": \"safeTransferFrom\", \"outputs\": [], \"payable\": true, \"stateMutability\": \"payable\", \"type\": \"function\"},  [\"\(from)\" \(to), \(tokenId)]"
+        if(Constants.legacy721Addresses.contains(contractAddress.drop0x.lowercased())) {
```
hboon: Looks like the condition is wrong. Should be:

```
if !legacy.contains  {
    abi = abiSafeTransferTo
}
```

But maybe better to separate it so you can avoid the `var` and no dangling `if` like this:

```
let abiLegacyTransfer = "{\"constant\":false,\"inputs\":[{\"name\":\"_to\",\"type\":\"address\"},{\"name\":\"_tokenId\",\"type\":\"uint256\"}],\"name\":\"transfer\",\"outputs\":[],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"}, [\"\(to)\" \(tokenId)]"
let abiSafeTransferFrom = "{ \"constant\": false, \"inputs\": [ { \"name\": \"_from\", \"type\": \"address\" }, { \"name\": \"_to\", \"type\": \"address\" }, { \"name\": \"_tokenId\", \"type\": \"uint256\" } ], \"name\": \"safeTransferFrom\", \"outputs\": [], \"payable\": true, \"stateMutability\": \"payable\", \"type\": \"function\"},  [\"\(from)\" \(to), \(tokenId)]"
let abi: String
if Constants.legacy721Addresses.contains(contractAddress.drop0x.lowercased()) {
    abi = abiLegacyTransfer
} else {
    abi = abiSafeTransferFrom
}
```

Also better to use `sameContract(as:)` so that it works regardless of case:

```
if Constants.legacy721Addresses.contains { $0.sameContract(as: contractAdress) }
```

hboon:
For posterity. ERC721 used to have a [`transfer(address, uint256)`](https://github.com/ethereum/EIPs/commit/8743a2286ce7ebf15faf9dd85aad23a940cfdc6a#diff-a87271f8186c59c075e229cd7421f205R135) function

James-Sangalli:
@hboon ready

PR #950, opened 2018-12-03
Remove unused code for custom networks

hboon:


James-Sangalli:
@hboon shouldn't remove this, means that developers can run the wallet with their custom networks e.g. quorum

PR #951, opened 2018-12-05
WIP: IFRAME

hboon:
@James-Sangalli This is just a throwaway PR. Don't merge.

Code is messy, some things are hardcoded, may force push or create a new branch for actual implementation later.

This branch implements IFRAME:

* Reads embedded IFRAME (HTML) in asset definition
* Assumes 1 of each of `<name>`, `<introduction>`, `<instruction>` tags and show them in that order.
* Strips `<iframe>`, `<script`> tags completely using XSLT
* Replaces `<a>` and `<button>` with their titles only using XSLT
* Wraps all 3 snippets of HTML with `width=device-width`
* Sets a fixed height for each of the 3 snippets of HTML

Makes a few assumptions that should be fixed:

* The structure of the action/function call definition
* The IFRAME provides the localized version in the language the user interface is in and does not fall back to the first localization.
* 0 or 1 action (button) in the IFRAME
* Probably many other assumptions..

**To support Open Door correctly, we'd need to be able to let IFRAME pass the app a message and for the app to prompt the user to sign it (or make a HTTP GET/POST call for the more general case). Just calling an smart contract function alone might be quite limiting?**

Following is an example asset definition that works with the door token. Must remove the `<t:inputs>` in order for "Open Door" button to work (it only makes an RPC call to fetch the `name` of the contract):

https://github.com/AlphaWallet/alpha-wallet-ios/blob/iframe-wip/sample.xml

On simulator:

1. Drop that file into the assetDefinitionsOverrides/ directory (you can figure out the path from the Realm directory path printed in the console when the app launches or when you switch accounts/networks)
2. Switch to an account with the door token on Kovan
3. Tap on the token
4. Modify the XML and it should live reload

On device:

1. Airdrop the XML to the device
2. Switch to an account with the door token on Kovan
3. Tap on the token
4. Modify the XML and Airdrop again

Screenshot below of what happens when "Open Door" is tapped:

screen shot 2018-12-05 at 10 18 29 pm


hboon:
@James-Sangalli With fb2a17206b1c35010a611aa4eef42c0d21bf379f, the branch shows the iframe (introduction tag) in the token list view for both NFTs and ERC20 tokens.

hboon:
Added a5fe1ac which disables tap/2-finger pinch to zoom iframe

hboon:
Forced pushed after rebase master.

James-Sangalli:
@hboon will close this and we can still use it from the branch.

PR #980, opened 2018-12-31
Redesigned send/receive screens for Ether and ERC20

hboon:
Doesn't include NTFs. The screen looks a little bare for ERC20 tokens since we aren't able to show transactions specifically for the token.

(For user testing 7-11 Jan?)


James-Sangalli:
Hi @hboon buttons look good but I will keep this as a WIP until completed, can use for testing like you mentioned also

hboon:
The transaction list for ERC20 in the screen is empty now. We can figure out from if a transaction is for ERC20 token transfer based on the `input` field, but I'll delay it till after the user testing. We aren't capturing the input data yet. To do it, we'll have to do a database migration. Makes it really hard to install another TestFlight build once we increase the version.

James-Sangalli:
Actually, hold that thought till UI flaws are cleared

hboon:
Haha. Yeah. I’ll push to this branch/PR though

hboon:
Be better if we rebase master instead of merging it into this branch though. It produces a cleaner history and also easier to maintain while we are working on several branches. But I’ll clean it up when I work on this branch again. Don’t worry.

James-Sangalli:
@hboon there was a merge conflict with master, hence why I merged it in

hboon:
Have rebased master and force push. Also push the version of this branch before rebase as `redesigned-send-receive-v1.4-build-250`. We can keep it temporarily in case we need to reproduce the build.

hboon:
Fixed #1008 with 5ad3d757f6b9fad71b0dd27a8c3373538e718aff (might break the commit hash with future rebase)

hboon:
Oh. You merged. Thanks.

Issue #875, opened 2018-11-03
Improve DApp Browser UI with our own landing page instead of stateofdapps


zhangzhongnan928:

farrahfazirah:
The storyboarding reference:  https://invis.io/Z5P16OVRM42

The design handover for developers (iOS): https://zpl.io/a3lmrPW

JamesSmartCell:
Looks very clean and fits with app UI, I wonder though if it would be good to have some dapps the user can click on on the page itself?

Maybe not; most search engine front pages before google were full of things to click on, now looking at yahoo it's almost painful.

Still google do have a 'daily feature'. Maybe we can have just one app featured on the page somehow?

Are users ready for the 'google' approach? Maybe we have to go through a stage where we feature a dapp each day that we know works well with the wallet?

zhangzhongnan928:
The DApp Browser home page is like the home page on your mobile.
You can add your own Dapps on that page.
Want to see suggested DApps, click the Discover DApps.




Regards,
[image: AlphaWallet] <https://awallet.io/>

*Victor Zhang*

[image: AlphaWallet Github] <https://github.com/alpha-wallet> [image:
AlphaWallet Twitter] <https://twitter.com/Alpha_wallet> [image: AlphaWallet
Reddit] <https://www.reddit.com/r/AlphaWallet> [image: AlphaWallet Facebook]
<https://www.facebook.com/AlphaWallet/> [image: AlphaWallet Youtube]
<https://www.youtube.com/channel/UCPP12PQeQmQ4_6qXzNvZw-w/> [image:
AlphaWallet Linkedin] <https://www.linkedin.com/company/%CE%B1wallet/> [image:
AlphaWallet Weibo] <https://weibo.com/AlphaWallet>
Join our Telegram group. <https://t.me/AlphaWalletGroup>

JamesSmartCell:
OK that makes sense. Would we want to feature a 'dapp of the day' picked by someone in the team because they liked it?

zhangzhongnan928:
Haha, it's good idea. Android is fine, you can even show the suggested
DApps directly in the home page.
But for iOS no, Apple not allow, we need to hide the suggested Dapps.


Regards,
[image: AlphaWallet] <https://awallet.io/>

*Victor Zhang*

[image: AlphaWallet Github] <https://github.com/alpha-wallet> [image:
[...]

PR #999, opened 2019-01-17
Implement redesigned Dapp browser and new Dapp home screens

hboon:
For #875 

Can't split it into smaller atomic commits unfortunately.

I have kept the existing database structure, so My Dapps and History are still stored  per-wallet with this PR.


hboon:
@James-Sangalli thanks!

PR #1009, opened 2019-01-18
rename references from eth to xdai for sending

James-Sangalli:
for #997 

Note: TokenObject is added into the database as Ethereum, therefore I had to patch the switch on token type to ensure that xdai was used when appropriate.

### DIFF
--- a/AlphaWallet/Localization/es.lproj/Localizable.strings
+++ b/AlphaWallet/Localization/es.lproj/Localizable.strings
@@ -324,3 +324,4 @@
 "dappBrowser.myDapps.edit.url.label" = "Address";
 "dappBrowser.clearMyDapps" = "Remove Dapp?";
 "recent.transactions" = "Recent Transactions";
+"blockchain.XDAI" = "xDAI Chain";
--- a/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
+++ b/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
@@ -92,21 +92,12 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         ethCostLabelLabel.translatesAutoresizingMaskIntoConstraints = false
         dollarCostLabelLabel.translatesAutoresizingMaskIntoConstraints = false
         dollarCostLabel.translatesAutoresizingMaskIntoConstraints = false
-
         pricePerTokenField.translatesAutoresizingMaskIntoConstraints = false
-        if config.chainID == 100 {
-            self.pricePerTokenField.cryptoToDollarRate = 1
-        } else {
-            ethPrice.subscribe { [weak self] value in
-                if let value = value {
-                    self?.pricePerTokenField.cryptoToDollarRate = value
-                }
-            }
-        }
+
+        setPricesForCoinType(config: config)
+
         pricePerTokenField.delegate = self
-
         ethCostLabel.translatesAutoresizingMaskIntoConstraints = false
-
         quantityStepper.translatesAutoresizingMaskIntoConstraints = false
         quantityStepper.minimumValue = 1
         quantityStepper.value = 1
@@ -258,6 +249,20 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
 
     @objc func quantityChanged() {
         updateTotalCostsLabels()
+    }
+
+    private func setPricesForCoinType(config: Config) {
+        switch config.server {
+        case .xDai:
+            //TODO fetch price
+            self.pricePerTokenField.cryptoToDollarRate = 1
+        case .rinkeby, .ropsten, .main, .custom, .callisto, .classic, .kovan, .sokol, .poa:
+            ethPrice.subscribe { [weak self] value in
+                if let value = value {
+                    self?.pricePerTokenField.cryptoToDollarRate = value
+                }
+            }
+        }
     }
 
     private func updateTotalCostsLabels() {
--- a/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
+++ b/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
@@ -54,13 +54,13 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
             config: Config,
             storage: TokensDataStore,
             paymentFlow: PaymentFlow,
-            ethPrice: Subscribable<Double>,
+            cryptoPrice: Subscribable<Double>,
             viewModel: EnterSellTokensCardPriceQuantityViewControllerViewModel
     ) {
         self.config = config
         self.storage = storage
         self.paymentFlow = paymentFlow
-        self.ethPrice = ethPrice
+        self.ethPrice = cryptoPrice
         self.viewModel = viewModel
 
         let tokenType = OpenSeaNonFungibleTokenHandling(token: viewModel.token)
@@ -93,9 +93,11 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         dollarCostLabelLabel.translatesAutoresizingMaskIntoConstraints = false
         dollarCostLabel.translatesAutoresizingMaskIntoConstraints = false
         pricePerTokenField.translatesAutoresizingMaskIntoConstraints = false
-
-        setPricesForCoinType(config: config)
-
+        cryptoPrice.subscribe { [weak self] value in
+            if let value = value {
+                self?.pricePerTokenField.cryptoToDollarRate = value
+            }
+        }
         pricePerTokenField.delegate = self
         ethCostLabel.translatesAutoresizingMaskIntoConstraints = false
         quantityStepper.translatesAutoresizingMaskIntoConstraints = false
@@ -249,20 +251,6 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
 
     @objc func quantityChanged() {
         updateTotalCostsLabels()
-    }
-
-    private func setPricesForCoinType(config: Config) {
-        switch config.server {
-        case .xDai:
-            //TODO fetch price
-            self.pricePerTokenField.cryptoToDollarRate = 1
-        case .rinkeby, .ropsten, .main, .custom, .callisto, .classic, .kovan, .sokol, .poa:
-            ethPrice.subscribe { [weak self] value in
-                if let value = value {
-                    self?.pricePerTokenField.cryptoToDollarRate = value
-                }
-            }
-        }
     }
 
     private func updateTotalCostsLabels() {
--- a/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
+++ b/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
@@ -94,9 +94,13 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         dollarCostLabel.translatesAutoresizingMaskIntoConstraints = false
 
         pricePerTokenField.translatesAutoresizingMaskIntoConstraints = false
-        ethPrice.subscribe { [weak self] value in
-            if let value = value {
-                self?.pricePerTokenField.ethToDollarRate = value
+        if config.chainID == 100 {
+            self.pricePerTokenField.cryptoToDollarRate = 1
+        } else {
+            ethPrice.subscribe { [weak self] value in
+                if let value = value {
+                    self?.pricePerTokenField.cryptoToDollarRate = value
+                }
             }
         }
         pricePerTokenField.delegate = self
--- a/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
@@ -413,9 +413,9 @@ extension TokensCoordinator: TokensViewControllerDelegate {
     func didSelect(token: TokenObject, in viewController: UIViewController) {
         switch token.type {
         case .nativeCryptocurrency:
-                    delegate?.didPress(for: .send(type: .nativeCryptocurrency(config: session.config, destination: .none)), in: self)
+            show(fungibleToken: token, transferType: .nativeCryptocurrency(config: session.config, destination: .none))
         case .xDai:
-            delegate?.didPress(for: .send(type: .xDai(config: session.config, destination: .none)), in: self)
+            show(fungibleToken: token, transferType: .xDai(config: session.config, destination: .none))
         case .erc20:
             show(fungibleToken: token, transferType: .ERC20Token(token))
         case .erc721:
--- a/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
@@ -340,7 +340,7 @@ class TokensCoordinator: Coordinator {
                         callCompletionFailed()
                     }
                 }
-            case .ether:
+            case .ether, .xDai:
                 break
             }
         }
@@ -377,7 +377,7 @@ class TokensCoordinator: Coordinator {
 
     private func makeCoordinatorReadOnlyIfNotSupportedByOpenSeaERC721(coordinator: TokensCardCoordinator, token: TokenObject) {
         switch token.type {
-        case .ether, .erc20, .erc875:
+        case .ether, .erc20, .erc875, .xDai:
             break
         case .erc721:
             switch OpenSeaNonFungibleTokenHandling(token: token) {
@@ -414,6 +414,12 @@ extension TokensCoordinator: TokensViewControllerDelegate {
         switch token.type {
         case .ether:
             show(fungibleToken: token, transferType: .ether(config: session.config, destination: .none))
+        case .ether, .xDai:
+            if token.name == "xDai" {
+                delegate?.didPress(for: .send(type: .xDai(config: session.config, destination: .none)), in: self)
+            } else {
+                delegate?.didPress(for: .send(type: .ether(config: session.config, destination: .none)), in: self)
+            }
         case .erc20:
             show(fungibleToken: token, transferType: .ERC20Token(token))
         case .erc721:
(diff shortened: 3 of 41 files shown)


AlphaWallet/Localization/es.lproj/Localizable.strings
```diff
@@ -322,3 +322,5 @@
 "dappBrowser.myDapps.edit.title.label" = "Title";
 "dappBrowser.myDapps.edit.url.label" = "Address";
 "dappBrowser.clearMyDapps" = "Remove Dapp?";
+
+"blockchain.XDAI" = "blockchain.XDAI";
```
hboon: Shall we just use the English value, so it'll look nicer if the user sees it? Or does it have to be like that for the localisation service to work with it?

i.e.

```
"blockchain.XDAI" = "xDAI Chain";
```

AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
```diff
@@ -94,9 +94,13 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         dollarCostLabel.translatesAutoresizingMaskIntoConstraints = false
 
         pricePerTokenField.translatesAutoresizingMaskIntoConstraints = false
-        ethPrice.subscribe { [weak self] value in
-            if let value = value {
-                self?.pricePerTokenField.ethToDollarRate = value
+        if config.chainID == 100 {
```
hboon: Better to use a  `switch` on `config.server` instead of if-else and `config.chainID`. Bit more code, but it's strong type checking (harder to key in or change the `100` to an invalid value) and it's easier to spot special handling of `RPCServer.xDai` by looking up references rather than search for `100`.

AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
```diff
@@ -94,9 +94,13 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
...
+            self.pricePerTokenField.cryptoToDollarRate = 1
```
hboon: Looks like it's possible to get the USD price of DAI with `https://api.coinmarketcap.com/v1/ticker/dai/`. Maybe we can do it in a separate PR?

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
```diff
@@ -391,8 +391,12 @@ class TokensCoordinator: Coordinator {
 extension TokensCoordinator: TokensViewControllerDelegate {
     func didSelect(token: TokenObject, in viewController: UIViewController) {
         switch token.type {
-        case .ether:
-            delegate?.didPress(for: .send(type: .ether(config: session.config, destination: .none)), in: self)
+        case .ether, .xDai:
+            if token.name == "xDai" {
```
hboon: Why not break up into 2 cases? :)

```
case .ether:
//...
case .xDai:
//...
```

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -270,22 +270,23 @@ extension TokensViewController: UITableViewDelegate {
...
         switch token.type {
-        case .ether:
+        case .ether, .xDai:
             let cellViewModel = EthTokenViewCellViewModel(
                     token: token,
                     ticker: viewModel.ticker(for: token),
                     currencyAmount: session.balanceCoordinator.viewModel.currencyAmount,
-                    currencyAmountWithoutSymbol: session.balanceCoordinator.viewModel.currencyAmountWithoutSymbol
+                    currencyAmountWithoutSymbol: session.balanceCoordinator.viewModel.currencyAmountWithoutSymbol,
+                    chainId: session.config.chainID
```
hboon: Can we pass in `session.config.server` instead of `session.config.chainID`. Stronger type check, etc.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -270,22 +270,23 @@ extension TokensViewController: UITableViewDelegate {
...
             )
             return cellViewModel.cellHeight
         case .erc20:
             let cellViewModel = TokenViewCellViewModel(token: token)
             return cellViewModel.cellHeight
         case .erc721:
-            let cellViewModel = NonFungibleTokenViewCellViewModel(token: token)
+            let cellViewModel = NonFungibleTokenViewCellViewModel(token: token, chainId: session.config.chainID)
```
hboon: Can we pass in session.config.server instead of session.config.chainID. Stronger type check, etc.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -319,14 +320,15 @@ extension TokensViewController: UITableViewDataSource {
...
+        case .ether, .xDai:
             let cell = tableView.dequeueReusableCell(withIdentifier: EthTokenViewCell.identifier, for: indexPath) as! EthTokenViewCell
             cell.configure(
                     viewModel: .init(
                             token: token,
                             ticker: viewModel.ticker(for: token),
                             currencyAmount: session.balanceCoordinator.viewModel.currencyAmount,
-                            currencyAmountWithoutSymbol: session.balanceCoordinator.viewModel.currencyAmountWithoutSymbol
+                            currencyAmountWithoutSymbol: session.balanceCoordinator.viewModel.currencyAmountWithoutSymbol,
+                            chainId: session.config.chainID
```
hboon: Can we pass in session.config.server instead of session.config.chainID. Stronger type check, etc.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -336,11 +338,11 @@ extension TokensViewController: UITableViewDataSource {
             return cell
         case .erc721:
             let cell = tableView.dequeueReusableCell(withIdentifier: NonFungibleTokenViewCell.identifier, for: indexPath) as! NonFungibleTokenViewCell
-            cell.configure(viewModel: .init(token: token))
+            cell.configure(viewModel: .init(token: token, chainId: session.config.chainID))
```
hboon: Can we pass in session.config.server instead of session.config.chainID. Stronger type check, etc.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -336,11 +338,11 @@ extension TokensViewController: UITableViewDataSource {
...
             return cell
         case .erc875:
             let cell = tableView.dequeueReusableCell(withIdentifier: NonFungibleTokenViewCell.identifier, for: indexPath) as! NonFungibleTokenViewCell
-            cell.configure(viewModel: .init(token: token))
+            cell.configure(viewModel: .init(token: token, chainId: session.config.chainID))
```
hboon: Can we pass in session.config.server instead of session.config.chainID. Stronger type check, etc.

AlphaWallet/Tokens/ViewModels/EthTokenViewCellViewModel.swift
```diff
@@ -10,17 +10,20 @@ struct EthTokenViewCellViewModel {
     private let currencyAmount: String?
     private let currencyAmountWithoutSymbol: Double?
     private let ticker: CoinTicker?
+    private let chainId: Int
```
hboon: Can we pass in `RPCServer` instead of `Int`? Stronger type check, etc.

AlphaWallet/Tokens/ViewModels/EthTokenViewCellViewModel.swift
```diff
@@ -36,7 +39,12 @@ struct EthTokenViewCellViewModel {
     }
 
     var blockChainName: String {
-        return R.string.localizable.blockchainEthereum()
+        if chainId == 100 {
+            //xdai network, can create switch statement in future if supporting many chains
```
hboon: I think it's better to use a `switch` right away. Much easier to discover this piece of code rather than remember to search for `100` later.

AlphaWallet/Tokens/ViewModels/NonFungibleTokenViewCellViewModel.swift
```diff
@@ -6,9 +6,11 @@ import BigInt
 
 struct NonFungibleTokenViewCellViewModel {
     private let token: TokenObject
+    private let chainId: Int
 
-    init(token: TokenObject) {
+    init(token: TokenObject, chainId: Int) {
```
hboon: Can we pass in `RPCServer` instead of `Int`? Stronger type check, etc.

AlphaWallet/Tokens/ViewModels/EthTokenViewCellViewModel.swift
```diff
@@ -10,17 +10,20 @@ struct EthTokenViewCellViewModel {
...
     private let ticker: CoinTicker?
+    private let chainId: Int
 
     init(
         token: TokenObject,
         ticker: CoinTicker?,
         currencyAmount: String?,
-        currencyAmountWithoutSymbol: Double?
+        currencyAmountWithoutSymbol: Double?,
+        chainId: Int
```
hboon: Can we pass in `RPCServer` instead of `Int`? Stronger type check, etc.

AlphaWallet/Tokens/ViewModels/NonFungibleTokenViewCellViewModel.swift
```diff
@@ -39,7 +41,12 @@ struct NonFungibleTokenViewCellViewModel {
     }
 
     var blockChainName: String {
-        return R.string.localizable.blockchainEthereum()
+        if chainId == 100 {
```
hboon: I think it's better to use a switch right away. Much easier to discover this piece of code rather than remember to search for 100 later.

AlphaWallet/Transfer/ViewModels/SendHeaderViewViewModel.swift
```diff
@@ -8,13 +8,19 @@ struct SendHeaderViewViewModel {
     var currencyAmount: String?
     var currencyAmountWithoutSymbol: Double?
     var showAlternativeAmount = false
+    var chainId = Config().chainID
```
hboon: `RPCServer` instead of `Int`

AlphaWallet/Transfer/ViewModels/SendHeaderViewViewModel.swift
```diff
@@ -8,13 +8,19 @@ struct SendHeaderViewViewModel {
...
 
     var issuer: String {
         return ""
     }
 
     var blockChainName: String {
-        return R.string.localizable.blockchainEthereum()
+        if chainId == 100 {
```
hboon: I think it's better to use a switch right away. Much easier to discover this piece of code rather than remember to search for 100 later.

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -11,6 +11,7 @@ class AmountTextField: UIControl {
     enum Currency: String {
         case eth = "ETH"
         case usd = "USD"
+        case xDai = "xDAI"
```
hboon: Think it should be "DAI" instead of "xDAI"?

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -81,7 +82,12 @@ class AmountTextField: UIControl {
     }()
 
     init() {
-        currentPair = Pair(left: .eth, right: .usd)
+        let chainId = Config().chainID
+        if chainId == 100 {
```
hboon: I think it's better to use a switch right away. Much easier to discover this piece of code rather than remember to search for 100 later.

Can you help look for this line in `AmountTextField.swift` and see if it's broken? I'm already regretting not using a switch here, otherwise, you'd found it. (and search for explicit references to `.eth` and `.usd` in the same file for similar bugs..

```
if currentPair.left != .eth {
```

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -195,18 +201,22 @@ class AmountTextField: UIControl {
             alternativeAmountLabel.text = atLeastOneWhiteSpaceToKeepTextFieldHeight
         } else {
             switch currentPair.left {
-            case .eth:
+            case .eth, .xDai:
                 alternativeAmountLabel.text = "~ \(amount) USD"
             case .usd:
-                alternativeAmountLabel.text = "~ \(amount) ETH"
+                if Config().chainID == 100 {
+                    alternativeAmountLabel.text = "~ \(amount) xDAI"
```
hboon: `DAI` instead of `xDAI`

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -195,18 +201,22 @@ class AmountTextField: UIControl {
             alternativeAmountLabel.text = atLeastOneWhiteSpaceToKeepTextFieldHeight
         } else {
             switch currentPair.left {
-            case .eth:
+            case .eth, .xDai:
                 alternativeAmountLabel.text = "~ \(amount) USD"
             case .usd:
-                alternativeAmountLabel.text = "~ \(amount) ETH"
+                if Config().chainID == 100 {
```
hboon: Better to switch on `RPCServer` instead of if-else on `Int`

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -217,12 +227,14 @@ class AmountTextField: UIControl {
...
     private func convertToAlternateAmountNumeric() -> Double? {
-        if let ethToDollarRate = ethToDollarRate, let string = textField.text, let amount = Double(string) {
+        if let ethToDollarRate = cryptoToDollarRate, let string = textField.text, let amount = Double(string) {
             switch currentPair.left {
             case .eth:
                 return amount * ethToDollarRate
             case .usd:
                 return amount / ethToDollarRate
+            case .xDai:
+                return amount
```
hboon: Better to combine with `.eth`:

```
case .eth, .xDai:
    return amount * ethToDollarRate
```

And probably rename `ethToDollarRate` to `cryptoToDollarRate` too.

AlphaWallet/UI/Views/AmountTextField.swift
hboon: Another way to do this is to change `Currency` to:

```
    enum Currency {
        case cryptoCurrency(String)
        case usd(String)
    }
```

But this implies that only the display value is different ("DAI" vs "ETH"), but handling is the same. If that's valid, this might be cleaner.

AlphaWallet/Localization/es.lproj/Localizable.strings
James-Sangalli: It is using the english version, you are correct about the localisation service, appcode added that in for me once I added it into the English file

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
James-Sangalli: See the TODO, realm stores the users address in the database as an ethereum type, hence it never hits xDai here and I have to check it inside the same case.

AlphaWallet/UI/Views/AmountTextField.swift
James-Sangalli: @hboon yes it functions the same way as a native currency, putting this in.

James-Sangalli:
@hboon all patched up, will add a seperate issue for price fetching on dai

James-Sangalli:
@hboon after merging with master it seems the send/receive ui has gone?

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -34,26 +35,28 @@ class AmountTextField: UIControl {
     var ethCost: String {
         get {
             switch currentPair.left {
-            case .eth, .xDai:
+            case .cryptoCurrency(_):
```
hboon: FYI. We can just use `case .cryptoCurrency:` if you don't need the associated values. (I switch and unused variable to `_` and then forget to remove it all the time. No biggie)

hboon:
> @hboon also solves #1010 however, the label is not updating despite the prices being fetched, see [d54a7e0](https://github.com/AlphaWallet/alpha-wallet-ios/commit/d54a7e0df4dabc4b98eb1d477f05789a129f39e2)

I pushed https://github.com/AlphaWallet/alpha-wallet-ios/pull/1009/commits/c8a95299cc06f76772ce0b3acf1cc6f8616bafbb which fixes this. Feel free to amend. There are 2 problems:

A. The symbols don't match "DAI" vs "xDai" which is fixed by `mapSymbolToVersionInRates()` doing a simple translation.
B. There are 2 switches where we had use `default` earlier, so the new `.xDai` wasn't handled correctly. Do you think you can help remove the default and break on the other enums? It's more verbose, but would have helped. (but read the note below first).

I realised something that you might want to explore. Maybe it's not necessary to add a `.xDai` case to `TransferType` at all? What if we renamed `.ether` to something like `.nativeCurrency` or `.cryptoCurrency`? I took a quick look and the behaviour seems the same for `.ether` and `xDai` (then you would just need A above and not B)

AlphaWallet/UI/Views/AmountTextField.swift
hboon: > @hboon not DAI, DAI is on the mainchain, since the xdai chain runs on POA, the native currency is xDAI rather than regular DAI. I make the distinction clear to the user.

Ah. I see. Thanks.

hboon:
> @hboon after merging with master it seems the send/receive ui has gone?

@James-Sangalli You should have it now that you have merge the send/receive UI branch into master.

But still, better to rebase master rather than merging master in though.

`git checkout xdai-renaming`
`git rebase master`

It's easier to resolve merge issues for you too (and cleaner history).

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
hboon: Hmm.. ok. A better way to do this if we don't do a migration: Push the logic into `TokenObject`, so `TokenObject.type` is always correct and we wouldn't forget to special case elsewhere.

```
    var type1: TokenType {
        get {
            if name == "xDai" {
                return .xDai
            } else {
                return TokenType(rawValue: rawType)!
            }
        }
        set {
            rawType = newValue.rawValue
        }
    }
```

PR #1030, opened 2019-01-21
Fix: tapping the Send button for ether and token card when on a watch-account doesn't show error message correctly

hboon:
The send button in this screen:

simulator screen shot - iphone x - 2019-01-21 at 11 22 50


James-Sangalli:
@hboon seems ethereum blockchain is also cut

hboon:
Yeah. Farrah created an issue.

PR #1041, opened 2019-01-22
Add support to display etherscan-like page for wallet address and fetch/display transaction list for POA Network

hboon:
Test address: 0x52fE2ba14dB890b5cf98FBB0Eac74AF7A38Aca5F.

We can probably use Blockscout for POA Sokol, Ethereum Classic too


hboon:
@James-Sangalli no, they are the same, we already use them for xDai.

Issue #960, opened 2018-12-12
Can't connect to Ethereum Classic [ETC] Dapps - suspect faulty RPC node.


antsankov:
I tried to connect to:
- [P3C.io](https://p3c.io/use.html)
- [Saturn.network](https://saturn.network)

These are currently the 2 largest ETC dapps. Neither of them worked with Alphawallet. I suspect this has to do with the node being chosen [here](https://github.com/AlphaWallet/alpha-wallet-ios/blob/910a0b8f80ed521db2cb09f4c425fa12439c24dc/AlphaWallet/Settings/Types/Config.swift#L89) - can you change this to one of the nodes [here](https://github.com/kvhnuke/etherwallet/blob/9051f3240fe1297d2b783aee7224cc7a4fb6b5db/app/scripts/nodes.js#L133) specifically `https://web3.gastracker.io` which is the most stable.

James-Sangalli:
Hi @antsankov, Thank you for finding this.

Turns out this is caused by an API change, we will look into adopting a new node that can support eth classic (thanks for listing some alternatives)

antsankov:
Yes! I am happy to help, if you need any beta testing or have any questions
about ETC.

PR #1048, opened 2019-01-24
Add working ETC Node

antsankov:
This is the same node that has been confirmed to work with MyEtherWallet see [here](https://github.com/MyEtherWallet/MyEtherWallet/blob/57690d95b88053fda8f507f2c72083d2ab5883ad/src/networks/nodes/etc-gastracker.js#L5).

Fixes issues connecting to ETC Dapps (Issue [960](https://github.com/AlphaWallet/alpha-wallet-ios/issues/960))


antsankov:
Fixes issue https://github.com/AlphaWallet/alpha-wallet-ios/issues/960

Issue #1072, opened 2019-02-01
Use payment master for native cryptocurrency links


James-Sangalli:
Related to #1053

PR #1089, opened 2019-02-10
add route and handling for currency links

James-Sangalli:
For #1072

### DIFF
--- a/AlphaWallet/Market/UniversalLinkHandler.swift
+++ b/AlphaWallet/Market/UniversalLinkHandler.swift
@@ -55,7 +55,6 @@ public class UniversalLinkHandler {
         let link = (message + signature).hexa2Bytes
         let binaryData = Data(bytes: link)
         let base64String = binaryData.base64EncodedString()
-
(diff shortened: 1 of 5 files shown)


AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -127,8 +126,8 @@ public class UniversalLinkHandler {
         let expiry = Array(bytes[16...19])
         let contractAddress = Array(bytes[20...39])
         let v = String(bytes[104], radix: 16)
-        let r = Data(bytes: Array(bytes[72...103])).hex()
-        let s = Data(bytes: Array(bytes[40...71])).hex()
+        let r = Data(bytes: Array(bytes[40...71])).hex()
+        let s = Data(bytes: Array(bytes[72...103])).hex()
```
hboon: Is v, r, s packed the same way as normal and spawnable MagicLinks now? Maybe we can reuse `getVRSFromLinkBytes()` like the other 2 type of links, maybe passing in an offset?)

Issue #1091, opened 2019-02-10
Display free dai offering instead of token card on currency links


James-Sangalli:
<img width="320" alt="screen shot 2019-02-11 at 10 57 02 am" src="https://user-images.githubusercontent.com/16630514/52541680-e390e880-2deb-11e9-9692-87484fa40658.png">

PR #1094, opened 2019-02-11
display amount of currency you can import

James-Sangalli:
For #1091

@hboon slightly hacky but requires a kind of TBML implementation to get it right exactly with the UI

### DIFF
--- a/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
@@ -203,8 +203,9 @@ class UniversalLinkCoordinator: Coordinator {
             let contractAsAddress = Address(string: signedOrder.order.contractAddress)!
             if signedOrder.order.nativeCurrencyDrop {
                 let amt: Decimal
-                if let amount = Decimal(string: signedOrder.order.count.description) {
-                    amt = amount / 10000
+                let szabosPerEth = 10000
+                if let amount = Decimal(exactly: signedOrder.order.count) {
+                    amt = amount / szabosPerEth
                 } else {
                     amt = 0
                 }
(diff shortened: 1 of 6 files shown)


AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -192,7 +202,25 @@ class UniversalLinkCoordinator: Coordinator {
             guard let recoverAddress = Address(string: ethereumAddress.address) else { return false }
             let contractAsAddress = Address(string: signedOrder.order.contractAddress)!
             if signedOrder.order.nativeCurrencyDrop {
-                self.makeTokenHolder([""], signedOrder.order.contractAddress)
+                let amt: Decimal
+                if let amount = Decimal(string: signedOrder.order.count.description) {
```
hboon: You also simplify this:

```
if let amount = Decimal(string: signedOrder.order.count.description) {
```

with:

```
if let amount = Decimal(exactly: signedOrder.order.count)
```

Looks a bit odd though. Do you really mean to read the value of `count`?

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -192,7 +202,25 @@ class UniversalLinkCoordinator: Coordinator {
...
+                    amt = amount / 10000
```
hboon: Is there an existing constant we can use for `10000`? If not do add something like this before the line and use the variable. It's at least a "executable" comment.

```
let someSensibleName = 10000
```

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
James-Sangalli: Yes, count records the amount in szabo

Issue #1095, opened 2019-02-12
Don't allow users to claim currency links without a payment server


James-Sangalli:
Not using a payment server is risky because it opens the user up to front running

PR #1096, opened 2019-02-12
require paymaster for certain currency links 

James-Sangalli:
For #1095

### DIFF
--- a/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
@@ -159,13 +159,15 @@ class UniversalLinkCoordinator: Coordinator {
         let requiresPaymaster = requiresPaymasterForCurrencyLinks(signedOrder: signedOrder)
         if signedOrder.order.price == 0 {
             self.checkPaymentServerSupportsContract(contractAddress: signedOrder.order.contractAddress) { supported in
+                //Currency links on mainnet/classic/xdai without a paymaster should be rejected for security reasons (front running)
+                guard supported || !requiresPaymaster else {
+                    self.showImportError(errorMessage: R.string.localizable.aClaimTokenFailedServerDown())
+                    return
+                }
                 if supported {
                     self.usePaymentServerForFreeTransferLinks(signedOrder: signedOrder)
-                } else if !requiresPaymaster {
-                    self.handlePaidImports(signedOrder: signedOrder)
                 } else {
-                    //Currency links on mainnet/classic/xdai without a paymaster should be rejected for security reasons (front running)
-                    self.showImportError(errorMessage: R.string.localizable.aClaimTokenFailedServerDown())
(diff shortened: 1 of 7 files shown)


AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -156,19 +156,35 @@ class UniversalLinkCoordinator: Coordinator {
 
     func completeOrderHandling(signedOrder: SignedOrder)
     {
+        let requiresPaymaster = requiresPaymasterForCurrencyLinks(signedOrder: signedOrder)
         if signedOrder.order.price == 0 {
             self.checkPaymentServerSupportsContract(contractAddress: signedOrder.order.contractAddress) { supported in
```
hboon: Here's a version using guard to make it clearer that is the precondition you are looking out for:

```
            self.checkPaymentServerSupportsContract(contractAddress: signedOrder.order.contractAddress) { supported in
                //Currency links on mainnet/classic/xdai without a paymaster should be rejected for security reasons (front running)
                guard supported || !requiresPaymaster else {
                    self.showImportError(errorMessage: R.string.localizable.aClaimTokenFailedServerDown())
                    return
                }

                if supported {
                    self.usePaymentServerForFreeTransferLinks(signedOrder: signedOrder)
                } else {
                    self.handlePaidImports(signedOrder: signedOrder)
                }
            }
```

Checking against classic means we don't allow classic drops until payment server supports them right?

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -156,19 +156,35 @@ class UniversalLinkCoordinator: Coordinator {
...
+                    //Currency links on mainnet/classic/xdai without a paymaster should be rejected for security reasons (front running)
+                    self.showImportError(errorMessage: R.string.localizable.aClaimTokenFailedServerDown())
                 }
             }
         } else {
             self.handlePaidImports(signedOrder: signedOrder)
         }
     }
 
+    private func requiresPaymasterForCurrencyLinks(signedOrder: SignedOrder) -> Bool {
```
hboon: Here's a version using guard to make it clearer that is the precondition you are looking out for:

```
        guard signedOrder.order.nativeCurrencyDrop else { return false }
        guard signedOrder.order.price == 0 else { return false }

        switch config.server {
        case .xDai, .classic, .main:
            return true
        default:
            return false
        }
```

Since there's a switch based on `config.server`, we can actually move that as a property into `RPCServer` too, like how `RPCServer.name` works

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
James-Sangalli: Any network with value should only allow paymaster

James-Sangalli:
@hboon done

hboon:
> @hboon done

@James-Sangalli Good job!

Issue #1050, opened 2019-01-26
Handle magic links with network specification, prevent user from importing on the wrong network


zhangzhongnan928:
Switch to the right network automatically when importing through MagicLink 
https://github.com/AlphaWallet/alpha-wallet-android/issues/521

James-Sangalli:
@colourful-land this would require a change to the link format itself, by specifying the chainID.

colourful-land:
Let's say what @James-Sangalli  asks for is done, what should happen?

Imagine this:

1. Alice is on the main chain. She has 3 ethers. She receives a magic link from Bob.
2. The link is about an asset on DAI network. It asks for 1Dai for a Dai-Kitten.
3. Alick clicks "Yes", wallet says "insufficient balance. you have 0."
4. Alice is in shock, thinking that her Ethers are gone. She goes to the main wallet interface.
5. Because the wallet is *automatically switched to* dai, the balance is 0 on main wallet interface. She confirmed that and freaks out.

Can we get @farrahfazirah's attention to work on an acceptable user flow on this and various other multi-change scenarios? The best way to move forward is to tell stories graphically. Farrah, we need to speed things up so can you produce (for design-space exploration purpose) wireframes only and not the full design, refine the detail only when the design is agreed on?

colourful-land:
Thank you, James. How urgent it is? I will work with you to get this working.

farrahfazirah:
![switchandhint](https://user-images.githubusercontent.com/20555752/52694590-2fd35880-2fa5-11e9-8002-9c36cc27b1f2.png)

1. Alice is on the Ethereum network
2. She receives a Magiclink from Bob asking for 1 Dai 1 Kitty
3. Alice opens the link and is prompted by a message saying that the wallet will automatically switch from Ethereum to xDai as the payment is in xDai

**If Alice has xDai**
4A. Network is switched to xDai. If Alice has xDai, it will automatically transition after 3 secs (give the user a bit more time to realise they're on xDai/see their balance) to the import/purchase page.

**If Alice has insufficient xDai**
4B. Network is switched to xDai. If Alice has insufficient xDai, it will automatically transition after 3 secs (give the user a bit more time to realise they're on xDai/see their balance) to a dialog box telling her she has insufficient balance. She confirms that she is aware of this and clicks Ok.

**Visual hints for which network user is on**
[...]

PR #1101, opened 2019-02-13
filter and handle magic links by DNS specified network

James-Sangalli:
For #1050

@hboon can merge once we have the server up and running with different subdomains

### DIFF
--- a/AlphaWallet/Localization/es.lproj/Localizable.strings
+++ b/AlphaWallet/Localization/es.lproj/Localizable.strings
@@ -329,3 +329,5 @@
 "a.claim.token.failed.notEnoughXDAI.title" = "a.claim.token.failed.notEnoughXDAI.title";
 "photos" = "Photos";
 "a.claim.token.failed.serverDown" = "a.claim.token.failed.serverDown";
+
+"a.claim.token.wrongNetworkLink" = "a.claim.token.wrongNetworkLink";
--- a/AlphaWallet/Market/Coordinators/UniversalLinkInPasteboardCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkInPasteboardCoordinator.swift
@@ -4,6 +4,7 @@ import UIKit
 
 protocol UniversalLinkInPasteboardCoordinatorDelegate: class {
     func importUniversalLink(url: URL, for coordinator: UniversalLinkInPasteboardCoordinator)
+    func showImportError(errorMessage: String, cost: ImportMagicTokenViewControllerViewModel.Cost?)
 }
 
 class UniversalLinkInPasteboardCoordinator: Coordinator {
@@ -12,9 +13,13 @@ class UniversalLinkInPasteboardCoordinator: Coordinator {
     
     func start() {
         guard let contents = UIPasteboard.general.string?.trimmed else { return }
-        guard contents.hasPrefix(UniversalLinkHandler().urlPrefix) else { return }
-        guard contents.count > UniversalLinkHandler().urlPrefix.count else { return }
         guard let url = URL(string: contents) else { return }
+        let isLegacy = contents.hasPrefix(Constants.legacyMagicLinkPrefix)
+        guard contents.hasPrefix(UniversalLinkHandler().urlPrefix) || isLegacy else {
+            let actualNetwork = Config().server.magicLinkNetwork(url: url.description)
+            delegate?.showImportError(errorMessage: R.string.localizable.aClaimTokenWrongNetworkLink(actualNetwork), cost: nil)
+            return
+        }
         UIPasteboard.general.string = ""
         delegate?.importUniversalLink(url: url, for: self)
     }
--- a/AlphaWallet/Market/UniversalLinkHandler.swift
+++ b/AlphaWallet/Market/UniversalLinkHandler.swift
@@ -46,16 +46,21 @@ extension Array {
 
 public class UniversalLinkHandler {
 
-    public let urlPrefix = "https://app.awallet.io/"
+    public let urlPrefix: String
+
+    init() {
+        urlPrefix = Config().magicLinkPrefix.description
+    }
 
     //message is with 32 bytes each of price and expiry and is shortened for link
     func createUniversalLink(signedOrder: SignedOrder) -> String {
+        let prefix = Config().magicLinkPrefix.description
         let message = formatMessageForLink(signedOrder: signedOrder)
         let signature = signedOrder.signature
         let link = (message + signature).hexa2Bytes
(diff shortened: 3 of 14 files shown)


AlphaWallet/Settings/Types/RPCServers.swift
```diff
@@ -33,6 +33,30 @@ enum RPCServer: Hashable {
         }
     }
 
+    func magicLinkNetwork(url: String) -> String {
+        if url.contains(Constants.mainnetMagicLinkPrefix) {
```
hboon: You can probably use something like and switch instead of if-else:

```
let url = URL(string: url) //optional though
url.host == "poa.awallet.io"
```

AlphaWallet/Settings/Types/Constants.swift
```diff
@@ -10,6 +10,19 @@ public struct Constants {
         Array(arrayLiteral: 0x58, 0x44, 0x41, 0x49, 0x44, 0x52, 0x4F, 0x50)
     ).hex()
 
+    // Magic link networks
+    public static let mainnetMagicLinkPrefix = "https://app.awallet.io/"
```
hboon: Shall we add a https://eth.awallet.io/ and make it compatible with "https://app.awallet.io/" but no longer generate `app.awallet.io` links at some point (after all mobile apps are updated)? That could free up the `app.awallet.io` sub-domain later.

James-Sangalli:
@hboon this is now compatible with the paymaster, @JamesSmartCell Please take note of this and update accordingly

James-Sangalli:
@lyhistory also take note here

AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -46,10 +46,9 @@ extension Array {
 
 public class UniversalLinkHandler {
 
-    public let urlPrefix = "https://app.awallet.io/"
-
     //message is with 32 bytes each of price and expiry and is shortened for link
     func createUniversalLink(signedOrder: SignedOrder) -> String {
+        let urlPrefix = Config().magicLinkPrefix.description
```
hboon: Might be good to hardcode to use the legacy "https://app.awallet.io/" when creating for mainnet for compatibility for existing app versions? (we can't change the `magicLinkPrefix` value because it's used elsewhere). If so create an issue so we don't forget to remove it?

James-Sangalli: I think it best to simply move on

James-Sangalli:
@hboon see new commit, note that we can accept old magic links in this version but older versions cannot take new links.

AlphaWallet/Localization/es.lproj/Localizable.strings
```diff
@@ -329,3 +329,5 @@
 "a.claim.token.failed.notEnoughXDAI.title" = "a.claim.token.failed.notEnoughXDAI.title";
 "photos" = "Photos";
 "a.claim.token.failed.serverDown" = "a.claim.token.failed.serverDown";
+
+"a.claim.token.wrongNetworkLink" = "a.claim.token.wrongNetworkLink";
```
hboon: Remember the %@ in other Localisable.strings files :)

AlphaWallet/Market/Coordinators/UniversalLinkInPasteboardCoordinator.swift
```diff
@@ -12,9 +12,9 @@ class UniversalLinkInPasteboardCoordinator: Coordinator {
     
     func start() {
         guard let contents = UIPasteboard.general.string?.trimmed else { return }
-        guard contents.hasPrefix(UniversalLinkHandler().urlPrefix) else { return }
-        guard contents.count > UniversalLinkHandler().urlPrefix.count else { return }
         guard let url = URL(string: contents) else { return }
+        guard contents.hasPrefix(UniversalLinkHandler().urlPrefix)
+                      || contents.hasPrefix(Constants.legacyMagicLinkPrefix) else { return }
```
hboon: This means it will only auto-detect MagicLinks for the existing network. Which is fine, if that's the intention.

hboon:
> @hboon see new commit, note that we can accept old magic links in this version but older versions cannot take new links.

@James-Sangalli Right, but that means if Alice generates a mainnet link on the new release and sends to Bob who is running an old release, he can't import. So maybe it is better to still generate legacy links for a while?

AlphaWallet/Localization/es.lproj/Localizable.strings
James-Sangalli: It is referencing it directly, not using the text?

AlphaWallet/Market/Coordinators/UniversalLinkInPasteboardCoordinator.swift
James-Sangalli: Hmm, might be best to port the logic which checks displays an error if a correct link on the wrong network

James-Sangalli:
> > @hboon see new commit, note that we can accept old magic links in this version but older versions cannot take new links.
> 
> @James-Sangalli Right, but that means if Alice generates a mainnet link on the new release and sends to Bob who is running an old release, he can't import. So maybe it is better to still generate legacy links for a while?

Perhaps

AlphaWallet/Localization/es.lproj/Localizable.strings
hboon: Hmm.. Maybe it'd be harmless actually. If you switch to say `es` and access the R.generated method with an arg, does it still just display "a.claim.token.wrongNetworkLink" verbatim?

I was thinking of changing it to "a.claim.token.wrongNetworkLink %@", but it'd still look like garbage. So as long as it runs and doesn't crash, that's good.

AlphaWallet/Market/Coordinators/UniversalLinkInPasteboardCoordinator.swift
hboon: @James-Sangalli yeah, for completeness.

hboon:
@James-Sangalli almost forgot, there's 2 more things to do:

- [ ] iOS repo — Update AlphaWallet.entitlements to include the new (sub)domain names. This is best done in Xcode as that will update both of Apple's side (if necessary) as well as the .entitlements file.
- [ ] Server — see if the `apple-app-site-association files` are served correctly. eg. https://ropsten.aw.app/apple-app-site-association

Adding associated domains to Entitlements:

1. In Xcode, in the navigator, click on AlphaWallet project (instead of Pods)
2. Under Targets, click on AlphaWallet
3. Click on Capabilities tap
4. Update the Domains section.

Issue #1113, opened 2019-02-21
Scan EIP 681 QR code while in Send screen and fill in address + amount


hboon:
See https://github.com/AlphaWallet/alpha-wallet-android/issues/601

JamesSmartCell:
The schema will support different token contract types and even functions.

When a function call is given, assume it's a smart contract call.

Validate:
- check if current Token matches the 'address' field of the code. If wrong, prompt with Token name.
- Optional/TODO: Check if contract supports the function (parse constructor code). If not, prompt.

This will require modifications to UI, so we can produce a prototype run probably on Android and then get formal UI for handling smart contract calls.

This way of doing it is a little clumsy - there should be an ability to scan QR codes near the front of the app, and the app can have a heuristic to work out what to do:

I suggest this action tree:
1. Address:
- is there a token contract at this address? If so, add the token contract to the watch list.
- no contract? Take user to 'send' screen (This has an issue the token contract could be on a different chain - do we scan all chains?).

2. EIP681 code:
- take user to send screen and handle as already implemented. If Token is a known Non-Fungible then take user to the relevant Token Detail page. In future TBML will take over from here.
- If token is not known, then app should read the token in and take user to detail page. There is a possibility the user was 'air-dropped' the token so has a balance there but doesn't know it.

PR #1119, opened 2019-02-25
support eip681 qr codes

James-Sangalli:
For #1113 

Note: Ignores the chainID for now

Edit: chain ids are checked

### DIFF
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -298,23 +298,38 @@ extension SendViewController: QRCodeReaderDelegate {
         reader.dismiss(animated: true) { [weak self] in
             self?.activateAmountView()
         }
+        guard let result = QRURLParser.from(string: result) else { return }
+        guard checkAndFillEIP861Details(result: result) else { return }
+    }
 
-        guard let result = QRURLParser.from(string: result) else {
-            return
+    func checkAndFillEIP861Details(result: ParserResult) -> Bool {
+        //TODO error display on returns
+        //Note: not checking the 'transferType' since erc20 is implied by whether it has uint256 and address.
+        //The contract address can be compared to the one in the token card and if it matches will proceed else fail
+        //this protects the user from sending funds to the wrong address
+        if let chainId = result.params["chainId"] {
+            guard self.config.chainID == Int(chainId) else { return false }
         }
-        targetAddressTextField.value = result.address
-
-        if let dataString = result.params["data"] {
-            data = Data(hex: dataString.drop0x)
+        //if erc20 (eip861 qr code)
+        if let recipient = result.params["address"], let amt = result.params["uint256"] {
+            guard recipient != "0" && amt != "0" else { return false }
+            //address will be set as contract address if erc token, therefore need to ensure the QR code has set the same contract address
+            //as the user is using
+            guard transferType.contract().eip55String.sameContract(as: result.address) else { return false }
+            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(amt) ?? BigInt(), units: .ether)
+            targetAddressTextField.value = recipient
+            return true
         } else {
-            data = Data()
+            targetAddressTextField.value = result.address
         }
-
-        if let value = result.params["amount"] {
-            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(value) ?? BigInt(), units: .ether)
+        //if ether transfer (eip861 qr code)
+        if let value = result.params["value"], let amountToSend = Double(value) {
+            guard value != "0" else { return false }
+            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(amountToSend), units: .ether)
         } else {
             amountTextField.ethCost = ""
         }
+        return true
     }
 }
 
--- a/AlphaWallet/Foundation/QRURLParser.swift
+++ b/AlphaWallet/Foundation/QRURLParser.swift
@@ -10,7 +10,8 @@ struct ParserResult: Equatable {
 
 struct QRURLParser {
     static func from(string: String) -> ParserResult? {
-        let parts = string.components(separatedBy: ":")
+        let result = string.replacingOccurrences(of: "pay-", with: "")
+        let parts = result.components(separatedBy: ":")
         if parts.count == 1, let address = parts.first, CryptoAddressValidator.isValidAddress(address) {
             return ParserResult(
                 protocolName: "",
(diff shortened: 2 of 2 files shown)


James-Sangalli:
@hboon hold up

hboon:
@James-Sangalli got it

hboon:
@James-Sangalli I remember James included a test case in a separate PR. I was hoping we can use that (or make it into a separate JSON and somehow feed it into both iOS and Android). I haven't check it out yet if the test data is what I was hoping to create. But we can do it next time. Useful for things like Universal link, asset definition parsing, etc too.

hboon:
For ERC20, I think we should either:

A. ignore the QR code or inform the user that it is not supported when we are in the ERC20 send screen
B. support it correctly in the ERC20 send screen by reading for `function_name`="transfer": ethereum:<target_address>/transfer?address=<contract_address>&uint256=12

Also check that `chain_id` is either missing or matches the current chain (even if we don't show an error message for now), in case someone generates a testnet link and senders unwittingly transfers on mainnet.

We can create issues to add these later if we aren't doing them in this PR:

Value has to support scientifc notation `value=2.014e18`
* chain_id
* ENS support

PS: Force pushes are good! 👍

James-Sangalli:
ERC20 QR data example:ethereum:0xedd6d7ba0ff9f4bc501a12529cb736ca76a4fe7e@11/transfer?address=0xedd6d7ba0ff9f4bc501a12529cb736ca76a4fe7e&uint256=1

Source: https://brunobar79.github.io/eip681-link-generator/#

hboon:
This doesn't affect us now, but just a note for myself because I'll forget about it:

The EIP doesn't state clearly whether the value of a non-empty `function_name` can be something other than `"transfer"` and whether it's designed to call other function signatures. But it seems to suggest it ought to. URL query string parameters don't rely on order, but EIP681 does. Our implementation (which uses a dictionary to hold the parsed parameters) will have to change in the future to handle ordering and duplicate parameter names correctly.

e.g. URL that will cause problems with our implementation:

ethereum:0xedd6d7ba0ff9f4bc501a12529cb736ca76a4fe7e@11/someFunctionToCall?address=0xedd6d7ba0ff9f4bc501a12529cb736ca76a4fe7e&uint256=1&uint256=2

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -298,23 +298,36 @@ extension SendViewController: QRCodeReaderDelegate {
         reader.dismiss(animated: true) { [weak self] in
             self?.activateAmountView()
         }
+        guard let result = QRURLParser.from(string: result) else { return }
+        guard checkEIP861Requirements(result: result) else { return }
+    }
 
-        guard let result = QRURLParser.from(string: result) else {
-            return
+    func checkEIP861Requirements(result: ParserResult) -> Bool {
```
hboon: Would be good to rename `checkEIP861Requirements()` because it not only checks, it fills the textboxes according to the URL data for both EIP861 and a raw wallet address string (which isn't EIP861).

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -298,23 +298,36 @@ extension SendViewController: QRCodeReaderDelegate {
...
+        //this protects the user from sending funds to the wrong address
+        if let chainId = result.params["chainId"] {
+            guard config.chainID == Int(chainId) else { return false }
         }
-        targetAddressTextField.value = result.address
-
-        if let dataString = result.params["data"] {
-            data = Data(hex: dataString.drop0x)
+        //if erc20 (eip861 qr code)
+        if let recipient = result.params["address"], let amt = result.params["uint256"] {
```
hboon: Should need to check if receipt and amt are empty too.

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -298,23 +298,36 @@ extension SendViewController: QRCodeReaderDelegate {
...
+            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(amt) ?? BigInt(), units: .ether)
+            targetAddressTextField.value = recipient
+            //address will be set as contract address if erc token, therefore need to ensure the QR code has set the same contract address
+            //as the user is using
+            guard transferType.contract().eip55String == result.address else { return false }
```
hboon: Should use `sameContract(as:)` to handle case differences.

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -298,23 +298,36 @@ extension SendViewController: QRCodeReaderDelegate {
...
+            return true
         } else {
-            data = Data()
+            targetAddressTextField.value = result.address
         }
-
-        if let value = result.params["amount"] {
-            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(value) ?? BigInt(), units: .ether)
+        //if ether transfer (eip861 qr code)
+        if let value = result.params["value"], let amountToSend = BigInt(value, radix: 16) {
```
hboon: Have to handle when it's ERC20, the value isn't decimals=18.

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
James-Sangalli: Done

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
James-Sangalli: ok

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
James-Sangalli: Actually, even in the case of erc20 it is 18 decimals, you can artificially lower this but it is up to the qr code maker to get this right. the default is 18 and the units in the QR code are set in whole ethers, meaning for the qr code maker to mess it up they would have to input a value too low to be accepted like 10 decimals points 0.00... when the smart contract accepts 5, however each smart contract still has the 18 decimal point default and the qr maker should be able to get this correct (we cannot know unless we query the smart contract to ask, it might not even be set in the smart contract either). I think it is safe to leave this as is.

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
James-Sangalli: Done

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
hboon: EIP681 [1] addresses this:

> UNIT is a URL-encoded unicode string. If UNIT is ETH, it always means a multiplier of 10^18. If it is something else AND the addressed contract has a symbol field exactly matching this string AND the contract has a decimals field, then 10 to that power is used as a multiplier. Otherwise, the payment request is deemed invalid. Applications that have no access to the blockchain should refuse accepting requests with a non-empty UNIT, if it is not ETH.

So a correct implementation should fetch `decimals` to verify. https://brunobar79.github.io/eip681-link-generator generates with decimal=0 for ERC20, but yeah, I guess it's not too dangerous to assume decimals=18 since the value sent will be smaller than expected; so we can leave it for now.

[1] https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -298,23 +298,38 @@ extension SendViewController: QRCodeReaderDelegate {
...
-        if let dataString = result.params["data"] {
-            data = Data(hex: dataString.drop0x)
+        //if erc20 (eip861 qr code)
+        if let recipient = result.params["address"], let amt = result.params["uint256"] {
+            guard recipient != "0" && amt != "0" else { return false }
+            amountTextField.ethCost = EtherNumberFormatter.full.string(from: BigInt(amt) ?? BigInt(), units: .ether)
+            targetAddressTextField.value = recipient
+            //address will be set as contract address if erc token, therefore need to ensure the QR code has set the same contract address
+            //as the user is using
+            guard transferType.contract().eip55String.sameContract(as: result.address) else { return false }
```
hboon: I missed this earlier. The guard for contract should be moved up so that it returns false and doesn't set the textbox(es).

Also a similar check when link is for ether and SendViewController is not?

PR #1126, opened 2019-03-11
checks if the currency link is claimed and refactor

James-Sangalli:
For #1125

### DIFF
--- a/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
@@ -192,6 +192,80 @@ class UniversalLinkCoordinator: Coordinator {
         return !config.server.isTestnet
     }
 
+    private func handleSpawnableLink(signedOrder: SignedOrder, tokens: [BigUInt]) {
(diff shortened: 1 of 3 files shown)


AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -310,6 +328,25 @@ class UniversalLinkCoordinator: Coordinator {
...
 
+    private func checkIfLinkClaimed(r: String, completionHandler: @escaping (Bool) -> Void) {
+        let parameters: Parameters = [ "r": r ]
+        Alamofire.request(
+                Constants.paymentServerClaimedToken,
+                method: .get,
+                parameters: parameters
+        ).responseJSON { result in
+            if let response = result.response {
+                if response.statusCode == 208 || response.statusCode > 299 {
```
hboon: Does it generate a response code of `208` if it's true?

And this part looks like it should be omitted:

```
|| response.statusCode > 299
```

James-Sangalli: @hboon yes and the second check is for if there is an error, if there is an error (above 299) then we assume we cannot proceed.

Issue #1131, opened 2019-03-17
Update link to community forum


hboon:
From https://github.com/AlphaWallet/TokenScript/issues/57, change link in app from https://community.alphawallet.com/c/feedback/ios-feedback to the corresponding https://www.tokenscript.org/c/AlphaWallet/ios-feedback

PR #1132, opened 2019-03-17
Update link to community forums

hboon:
For #1131


James-Sangalli:
@hboon alphawallet feedback is on the tokenscript site?

hboon:
@James-Sangalli yeah. It is now. Weiwu merged them so there’s just 1 Discourse site.

PR #1133, opened 2019-03-18
Multi chain concurrently

hboon:


hboon:
@James-Sangalli I have forced pushed to fix a display bug. See screenshot for separator and issuer. Look for the test ticket.

In case you need it, the last build you did was for 1.60 (264).

Simulator Screen Shot - iPhone X - 2019-03-22 at 10 46 07

hboon:
Fixed and done #1157, #1158, #1159

hboon:
Fixed #1156

hboon:
Forced pushed. Previous version was 1.60/267.

hboon:
@James-Sangalli when you think it's good, let me know and I'll squash the commits into one (or a few bigger commits and we can merge into master.

hboon:
> Working fine on my side

@James-Sangalli thanks. I've squashed it into a single gigantic commit. Ready to merge.

James-Sangalli:
@hboon travis is acting strange, all the PR tests currently open run perfectly fine on my machine but fail on travis, what about for you?

hboon:
@James-Sangalli there's 1 test which fails on my machine if ran as part of the entire suite, but works if ran standalone. Travis is failing for seemingly entirely different reasons though. My sometimes failing test should be removed with #1161

James-Sangalli:
> @James-Sangalli there's 1 test which fails on my machine if ran as part of the entire suite, but works if ran standalone. I thought it's only me. However, Travis is failing for seemingly entirely different reasons though. My sometimes failing test should be removed with #1161

Likewise for #1135 but #1135 shouldn't fail to begin with?

hboon:
#1135 is failing at:

1. ClaimOrderCoordinatorTests.testClaimOrder()
2. ConfigTests.testWeb3StillLoadsAfterSwitchingLocale()

(1) is probably what triggered us to implement #1161 in the first place :)
(2) is the one which fails locally for me. I figure we can leave it since #1161 will get rid of it.

hboon:
Forced pushed to include approach (1) of https://github.com/AlphaWallet/alpha-wallet-ios/issues/1166

hboon:
@James-Sangalli Want to remove the last 2 commits for versioning and merge into master first?

Issue #1135, opened 2019-03-18
When auto detecting partner tokens, we should check for the token type and use the appropriate function to get balance instead of assuming it's ERC20


hboon:
This explains why 0xa16b70E8fAd839E62aBBa2d962E4ca5a28aF9e76 gets added despite a balance of 0.

Look in TokensDataStore.autoDetectTokens() for GetBalanceCoordinator

PR #1143, opened 2019-03-20
check token type before getting the balance for partner contracts

James-Sangalli:
For #1135


James-Sangalli:
Hmm Travis, maybe there calls are slow? This test doesn't fail on my machine, does it fail on yours? testClaimOrder

hboon:
@James-Sangalli works locally on mine too. Try `git rebase master` to include #1152. I wonder if it's fixed by that. i.e. when `Web3Swift.request()` gets invoked, `isLoaded` is still false, and then `self` (the `Web3Swift` instance) is destroyed, so the early return kicks in and the request is not executed.

But this will only happen if the `wait()` in the test case doesn't retain the test case itself, which is a bit strange.

Don't you wish for tracing garbage collectors (like Java's) sometimes?

PR #1146, opened 2019-03-21
with new api route for infura

James-Sangalli:
WIP until error fixed or we move node provider 

For #731

### DIFF
--- a/AlphaWallet/EtherClient/EtherServiceRequest.swift
+++ b/AlphaWallet/EtherClient/EtherServiceRequest.swift
@@ -19,7 +19,7 @@ struct EtherServiceRequest<Batch: JSONRPCKit.Batch>: APIKit.Request {
     }
 
     var path: String {
-        return "/"
+        return ""
     }
 
     var parameters: Any? {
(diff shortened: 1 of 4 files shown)


hboon:
@James-Sangalli good job!

James-Sangalli:
@hboon that should happen since the id is there and actually the ether balances won't show in the wallet key view with that link. It boggles my mind since everything is correct.

hboon:
@James-Sangalli Oh. try changing `EtherServiceRequest.path` from:

```
    var path: String {
        return "/"
    }
```

to:

```
    var path: String {
        return ""
    }
```

There are a few places which accesses `EtherServiceRequest`. Not sure if this is compatible with all of them


This came from the docs (which works):

```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "id": 1, "method": "eth_blockNumber", "params": []}' "https://mainnet.infura.io/v3/da3717f25f824cc1baa32d812386d93f"
```

Our equivalent has a trailing slash:

```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "id": 1, "method": "eth_blockNumber", "params": []}' "https://mainnet.infura.io/v3/da3717f25f824cc1baa32d812386d93f/"
```

James-Sangalli:
@hboon excellent spotting

Issue #1149, opened 2019-03-22
Support OpenSea on Ropsten


hboon:

hboon:
Looks like OpenSea only supports Rinkeby, not Ropsten at the moment.

James-Sangalli:
Not supported.

PR #1154, opened 2019-03-24
add support for ropsten and rinkeby opensea api calls

James-Sangalli:
For #979 & #1149


James-Sangalli:
@hboon yea thought so, all good now.

PR #1161, opened 2019-03-31
replace webkit web3 encoding calls with native web3swift

James-Sangalli:
For #1160 

A step forward for decentralisation!


AlphaWallet/InCoordinator.swift
```diff
@@ -420,7 +420,7 @@ class InCoordinator: Coordinator {
         let s = "0x" + signature.substring(with: Range(uncheckedBounds: (64, 128)))
         guard let wallet = keystore.recentlyUsedWallet else { return }
 
-        ClaimOrderCoordinator(web3: web3).claimOrder(
+        ClaimOrderCoordinator(web3: web3, config: self.config).claimOrder(
```
hboon: I haven't look over the whole thing yet. But code following this line should be changed too:

```
        ClaimOrderCoordinator(web3: web3, config: self.config).claimOrder(
```

The old version of `claimOrder()` (despite the name) only encodes the function call and prepares the transaction payload and the payload is then used to populate a transaction in the completion block passed to `claimOrder()`, starting at line 434.

Since we are now encoding and then calling `callSmartContract()` (this is good!) in `claimOrder()`, the completion block passed to it should be much simplified.

In other words,

* Old version of `claimOrder()` only encodes the function call
* New version of `claimOrder()` encodes + makes the function call

James-Sangalli: Sussed. Tests pass on your machine?

AlphaWallet/InCoordinator.swift
```diff
@@ -429,91 +429,9 @@ class InCoordinator: Coordinator {
...
-                        completion(true)
-                        print(res)
-                    case .failure(let error):
-                        completion(false)
-                        print(error)
-                    }
-                }
+            case .success(_):
+                completion(true)
             case .failure: break
```
hboon: Should do this when it fails:

```
completion(false)
```

AlphaWallet/Tokens/Coordinators/ClaimOrderCoordinator.swift
```diff
@@ -93,26 +122,26 @@ class ClaimOrderCoordinator {
...
+            ] as [AnyObject]
+            callSmartContract(withConfig: config, contract: Address(string: signedOrder.order.contractAddress)!, functionName: function.name, abiString: function.abi, parameters: parameters).done { result in
+                if let res = result["0"] as? String {
+                    completion(.success(res))
+                } else {
+                    completion(.failure(AnyError(Web3Error(description: "Error"))))
+                }
             }
+        } catch {
+
```
hboon: Should have a:

```
completion(.failure(AnyError(Web3Error(description: "Error"))))
```

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -123,88 +124,103 @@ class TransactionConfigurator {
...
-                            gasPrice: self.calculatedGasPrice,
-                            gasLimit: GasLimitConfiguration.maxGasLimit,
-                            data: data
-                    )
-                    completion(.success(()))
-                case .failure(let error):
-                    completion(.failure(error))
-                }
+            return completion(.success(()))
+        case .ERC20Token(_):
```
hboon: Some of these values should be read from `ContractERC721Transfer`, `ContractERC875Transfer `, `ContractERC20Transfer`.

hboon:
@James-Sangalli that's all I have unless you make other changes.

James-Sangalli:
@hboon for some reason the callSmartContract is not happy with the params although there is little insight into why. If you use optionals this can trigger it but it is working for erc20 and erc875 transfers but failing on 721 and 875 trade deals. Also, the encoders are no longer needed except for the 875 magic links since the abi is not referenced for tx data creation.

James-Sangalli:
For 721 and 875 it is not working properly, see here: https://etherscan.io/tx/0x0974dcbf9f840692fd5a747b3d075d47e3969d14687a97a3457662dbff2ccf32

hboon:
@James-Sangalli 

I ran this snippet to figure out why it's not working. You can paste them into AppDelegate and call both `foo()` and `bar()` to compare:

```
    private func foo() {
        let encoder = ABIEncoder()
        let parameters = [
            try! ABIValue(Address(string: "0x0000000000000000000000000000000000000011")!, type: ABIType.address),
            try! ABIValue(Address(string: "0x0000000000000000000000000000000000000022")!, type: ABIType.address),
            try! ABIValue(BigUInt("3")!, type: ABIType.uint(bits: 256))
        ] as [Any]
        try! encoder.encode(signature: "transferFrom(address,address,uint256)")
        try! encoder.encode(Data(fromArray: parameters), static: false)
        NSLog("xxx data:")
        dump(encoder.data.hexEncoded)
    }

    private func bar() {
        let encoder = ABIEncoder()
        try! encoder.encode(signature: "transferFrom(address,address,uint256)")
        try! encoder.encode(Address(string: "0x0000000000000000000000000000000000000011")!)
        try! encoder.encode(Address(string: "0x0000000000000000000000000000000000000022")!)
        try! encoder.encode(BigUInt("3")!)
        NSLog("xxx data:")
        dump(encoder.data.hexEncoded)
    }
```

The results are:

foo:

```
0x23b872dd
0000000000000000000000000000000000000000000000000000000000000060
8072dd01006000200000000000000000000000000000000020e9bd0801000000
4055dc01006000200000000000000000000000000000000020e9bd0801000000
0054dc01006000000000000000000000000000000000000020e9bd0801000000
```

bar:

```
0x23b872dd
0000000000000000000000000000000000000000000000000000000000000011
0000000000000000000000000000000000000000000000000000000000000022
0000000000000000000000000000000000000000000000000000000000000003
```

In short, there are 3 args (3 scalar values), so call `encoder.encode()` 3 times, not 1 array (which we would then call `encoder.encode()` once).

hboon:
@James-Sangalli BTW, if you look in the `web3swift` pod (the one we are switching the calls to), there's also a `ABIEncoder` and friends types. Are they the same as the one we are vendor-ing, which this PR is using? I don't know if you are intentionally using the vendor-ed types and not the ones in `web3swift` pod.

I think it's not a bad idea to use the vendor-ed version anyway since it creates less dependency on the `web3swift` pod.

PR #1164, opened 2019-04-03
Fix: category names in Discover Dapps

hboon:
Make Games and Tools plural and fix spelling variations: Financial vs Finance.


hboon:
cc @JamesSmartCell in case this affects you too.

PR #1170, opened 2019-04-07
Add support for TokenScript

hboon:
This handles the canonicalized version of the TokenScript file as `0x0c8d487dd27d7c4a027d9f92915659139e0b2ff2.tsml`. The file can be AirDropped onto an iOS device.

Because `.tsml` is an uncommon or unused extension, there might not be other apps that claim to support it, so after the AirDrop is complete, the AlphaWallet app will be launched/resumed if it is not already active, otherwise it will be silent. If it is currently showing the token instance iconified view or token instance view, they will live-reload. If the action view is current displayed, it has to be closed (tap the back button) and tap `Enter` again to reload.

It doesn't do these:

1. ~~Handle separate files.~~ (update: Done)
2. ~~Check the namespace of the TokenScript file and warn user to upgrade their app~~ (update: Done)
3. Whitelisting of URLs. I commented out the whitelisting feature so all URLs can be accessed, since there's no way to specify the whitelist in the TokenScript at the moment and maybe workshop attendees will want to put in their own URLs

I'll try to figure out (1) and (2), but this looks like a good baseline to do a TestFlight and test it out.

Attached a few screenshots:

&nbsp; | &nbsp;
--- | ---
Simulator Screen Shot - iPhone X - 2019-04-07 at 18 36 45 | Simulator Screen Shot - iPhone X - 2019-04-07 at 18 36 51
Simulator Screen Shot - iPhone X - 2019-04-07 at 18 37 01 | Simulator Screen Shot - iPhone X - 2019-04-07 at 18 37 05
Simulator Screen Shot - iPhone X - 2019-04-07 at 18 37 07 |

cc @colourful-land  @JamesSmartCell


hboon:
Added a commit for (1):

Have made it such that you can AirDrop files like `token.en.shtml` by naming it in the format: `contract-entityName.tsml` eg. `0x0c8d487dd27d7c4a027d9f92915659139e0b2ff2-token.en.tsml`. The TokenScript client then replaces the entity with the file contents verbatim in `contract.tsml` (the parser doesn't do that unfortunately).

So this is possible:

1. Tap on token in Wallet tab to show token instance iconified view (without TokenScript file)
2. AirDrop `0x0c8d487dd27d7c4a027d9f92915659139e0b2ff2.tsml` and token instance iconified view reloads
3. Tap on iconified view to show token instance view.
4. AirDrop `0x0c8d487dd27d7c4a027d9f92915659139e0b2ff2-token.en.tsml` and token instance view reloads
5. AirDrop `0x0c8d487dd27d7c4a027d9f92915659139e0b2ff2-style.tsml` and token instance reloads with updated CSS styles

In addition, you can also go to Settings tab > TokenScript Overrides and tap on a overridden file to share it to another iOS device. See screenshots.

@James-Sangalli another TestFlight build please? v1.6.x. Good for the workshop?

&nbsp; | &nbsp;
--- | ---
IMG_1128 | IMG_1129

hboon:
Supports #1171

hboon:
@James-Sangalli Added a commit to switch repo server to https://repo.tokenscript.org in case I'm away and it's handy for testing the new repo.

Feel free to `git reset a73735cc51976e1ec496b957a6b9c6834ebd3152` locally to move to the previous commit before building if you want to skip it.

hboon:
With the last few commits, un-canonicalized TokenScript files (.xml, not .tsml) and the files referred to by entities are assumed and expected to be in the same directory (i.e. what used to be `cards/` have to be moved 1 level up). These files as well as the canonicalized TokenScript files (.tsml) can be AirDrop-ed directly to iOS devices without having to rename the files.

And here's how they look like in the `TokenScript Overrides` screen:

Simulator Screen Shot - iPhone X - 2019-04-11 at 14 21 47

hboon:
Force pushing. Last build was for 274.

hboon:
Rebased master

hboon:
Forced pushed. Last build was 275.

hboon:
Squashed history and forced push to the branch.

* de1c8b06d25cff6fd094ac256def317c61e7d54f (for now. The hash will change) is "Add support for TokenScript schema v2019/04". I'll base all changes to add support for schema v2019/05. This commit itself will only be modified when `git rebase master`. Handy to look back at this commit to restore 2019/04.
* Commits after this are for schema v2019/05. WIP for now and will be squash/edited until we finalise work on v2019/05

hboon:
The build at this point bundles the XDAI bridge in the app. It might or might not include some quirky changes I made during development but should be overridable by dropping another file into `assetDefinitionsOverrides/` or via AirDrop. But let me know if the built-in one looks wrong.

We'll need the final and signed version at some point.

hboon:
@James-Sangalli Mostly there! I'll come back to drop in an updated xDai bridge TokenScript file, but there should be minimal code changes now other than fixes.

hboon:
@James-Sangalli heads-up. Some fixes.

hboon:
Ah. I forgot to add this. For posterity:

As part of 2019/05, I've added an Address class. This is commonly seen as AlphaWallet.Address. It's tucked inside an otherwise empty enum called AlphaWallet to use as a namespace until Swift supports explicit namespaces (with modules being implicit namespaces). AlphaWallet.Address is restricted to the TokenScript parts of the codebase for this PR. After this PR we can go about removing most traces of Address (TrustKeystore) and EthereumAddress (web3swift). We can't remove EthereumAddress completely because both libraries' public interfaces use them. But a good goal is to strive for is as few `import TrustKeystore` and `import web3swift` as possible.

Issue #1173, opened 2019-04-09
ENS name should come from resolver contract not owner 


James-Sangalli:

James-Sangalli:
@JamesSmartCell Is this correct on android?

PR #1174, opened 2019-04-11
get ENS address from resolver rather than owner

James-Sangalli:
For #1173 

@hboon do we need EthereumAddress and Address? 

@JamesSmartCell Please ensure that Android does this correctly.


hboon:
> @hboon do we need EthereumAddress and Address?

@James-Sangalli Unfortunately yes, the former is part of the Web3Swift pod and latter is part of TrustKeyStore; they have be used when interacting with the respective pods.

I was working on a branch a long time ago to introduce a third Address type (the horror!) which will replace web3swift.EthereumAddress/TrustKeyStore.Address for most of the cases so we can be less tightly coupled TrustKeyStore (good clue is to search for how many files include "import TrustKeystore" just to use the `Address` type). This new Address type will then provide a way to convert to/from EthereumAddress and TrustKeyStore's Address when we really need them. It's a general approach that is useful to be less dependent on third party libraries. Our codebase has changed a lot from implementing multi-chain and TokenScript.. so will have to restart that effort again.

Part of switching to our own Address type should also include converting all String representations of addresses to the new type. But this is all part of another issue/PR.

James-Sangalli:
@hboon the test works properly with the new method

hboon:
Good job!

Issue #1182, opened 2019-04-15
Support Görli Tesnet


JamesSmartCell:
Here's the details:

GOERLI_NETWORK_NAME = "Görli (Test)";
GOERLI_SYMBOL = "GÖETH";
GOERLI_ID = 5;
GOERLI_RPC_URL = "https://goerli.infura.io/v3/<your infura key>";
Etherscan API: "https://api-goerli.etherscan.io/"
Etherscan TX page URL: "https://goerli.etherscan.io/tx/"
Colour: #BBAE9A

Everything else works same as Rinkeby/Ropsten testnets.

James-Sangalli:
Goerli or Gorli?

PR #1186, opened 2019-04-16
with new goerli testnet and infura spec

James-Sangalli:
For #1182

### DIFF
--- a/AlphaWallet/Settings/Types/RPCServers.swift
+++ b/AlphaWallet/Settings/Types/RPCServers.swift
@@ -15,6 +15,7 @@ enum RPCServer: Hashable, CaseIterable {
     case classic
     case callisto
     case xDai
+    case goerli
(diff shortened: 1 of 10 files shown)


AlphaWallet/Settings/Types/RPCServers.swift
```diff
@@ -221,6 +230,7 @@ enum RPCServer: Hashable, CaseIterable {
             case .poa: return "https://blockscout.com/poa/core/api"
             case .xDai: return "https://blockscout.com/poa/dai/api"
             case .sokol: return "https://blockscout.com/poa/sokol/api"
+            case .goerli: return "https://api-goerli.etherscan.io/"
```
hboon: Without the trailing `/` perhaps? Like:

```
case .goerli: return "https://api-goerli.etherscan.io"
```

Issue #1188, opened 2019-04-17
Support ICAP addresses


James-Sangalli:
To be compatible with ICAP and IBAN, we can have it that the wallet only generates keys that start with 00. This will mean all the keys in our wallet will be compatible with the ICAP standard.

https://ethereum.stackexchange.com/questions/1085/what-is-an-icap-address

PR #1189, opened 2019-04-17
create an ICAP address on generation

James-Sangalli:
For #1188 

@hboon hold off on this one, might not actually be worth it...


James-Sangalli:
Nobody seems to use this format anymore, closing but we can always reference it later if it becomes relevant.

colourful-land:
If we are the only one using this feature, it is a security issue. The addresses using ICAP might be considered an AlphaWallet address and be targeted for AlphaWallet weakness.

This feature is considered if there are multiple wallets supporting it (are there?)

Either way, I expect ICAP related features to be in the same tier of rich functions that requires a smart contract (wallet contract) to support, therefore it's more like a wallet contract address generation consideration.

PR #1195, opened 2019-04-23
return ETH symbol rather than Goerli

James-Sangalli:


James-Sangalli:
@hboon you mean with the special o char?

hboon:
> @hboon you mean with the special o char?

No. From "0.15 Görli (GÖETH)" to "0.15 Görli (ETH)".

The hardcoded string which you updated is used in the app except when displaying the list of tokens in the wallet tab. The symbol is saved into the database, once, upon creation and read to display in the Wallets tab.

But then Görli support isn't available on the app store yet, only 2-3 of us will have this problem, so maybe it's not worth doing a migration for this..

James-Sangalli:
@hboon yes, not yet released. Probably best we simply do it. Will it cause a crash or simply the wrong symbol to be displayed?

hboon:
Simply wrong symbol for the 2-3 of us until reinstall. No crash or error otherwise.

Probably not worth the effort to do a migration. Up to you.

PR #1202, opened 2019-04-24
compat with swift 4

James-Sangalli:
For #1201 

Try it on the latest xcode, ensuring you clean the build first :)


James-Sangalli:
@hboon it's all good now

Issue #1211, opened 2019-04-29
Receiving Dai doesn't show up in recent transactions 


zhangzhongnan928:
But balance is updated correctly.
![CB1F8496-9F1B-44A2-B080-5320571136CD](https://user-images.githubusercontent.com/33795543/56887809-1b333680-6aa5-11e9-8f77-cf2a0a6c92bf.jpeg)

PR #1217, opened 2019-05-10
Get received transactions

James-Sangalli:
This PR:

- Enables an event lookup for erc20 transactions meaning you can get incoming transactions and pick up contracts you haven't transacted with directly (from Etherscan)

- Adds a flag to the Transaction object to prevent getting the block number from one of these event transactions and thus miss some of the older transactions you might have done 

- Adds a function to call for generic events 

Solves #1211 and issues related to it 

@JamesSmartCell This is relevant for Android too.

### DIFF
--- a/AlphaWallet/Core/Initializers/MigrationInitializer.swift
+++ b/AlphaWallet/Core/Initializers/MigrationInitializer.swift
@@ -16,7 +16,7 @@ class MigrationInitializer: Initializer {
     }
 
     func perform() {
-        config.schemaVersion = 2
+        config.schemaVersion = 3
         config.migrationBlock = { migration, oldSchemaVersion in
             if oldSchemaVersion < 2 {
                 //Fix bug created during multi-chain implementation. Where TokenObject instances are created from transfer Transaction instances, with the primaryKey as a empty string; so instead of updating an existing TokenObject, a duplicate TokenObject instead was created but with primaryKey empty
@@ -27,6 +27,15 @@ class MigrationInitializer: Initializer {
                         migration.delete(newObject)
                         return
                     }
+                }
+            }
+        }
+        config.migrationBlock = { migration, oldSchemaVersion in
+            if oldSchemaVersion < 3 {
+                migration.enumerateObjects(ofType: Transaction.className()) { oldObject, newObject in
+                    guard let oldObject = oldObject else { return }
+                    guard let newObject = newObject else { return }
+                    newObject["isERC20Interaction"] = false 
                 }
             }
         }
--- a/AlphaWallet/Tokens/ViewModels/TokenViewControllerViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/TokenViewControllerViewModel.swift
@@ -3,13 +3,16 @@
 import Foundation
 import UIKit
 import TrustKeystore
+import web3swift
+import Alamofire
+import PromiseKit
 
-struct TokenViewControllerViewModel {
+class TokenViewControllerViewModel {
     private let transferType: TransferType
     private let session: WalletSession
     private let tokensStore: TokensDataStore
     private let transactionsStore: TransactionsStorage
-    let recentTransactions: [Transaction]
+    var recentTransactions: [Transaction] = []
 
     init(transferType: TransferType, session: WalletSession, tokensStore: TokensDataStore, transactionsStore: TransactionsStorage) {
         self.transferType = transferType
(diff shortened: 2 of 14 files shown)


AlphaWallet/Core/Initializers/MigrationInitializer.swift
```diff
@@ -30,5 +30,14 @@ class MigrationInitializer: Initializer {
                 }
             }
         }
+        config.migrationBlock = { migration, oldSchemaVersion in
+            if oldSchemaVersion < 3 {
+                migration.enumerateObjects(ofType: Transaction.className()) { oldObject, newObject in
+                    guard let oldObject = oldObject else { return }
+                    guard let newObject = newObject else { return }
+                    newObject["isERC20Interaction"] = true
```
hboon: Should be false, like this:

```
newObject["isERC20Interaction"] = false
```

AlphaWallet/Tokens/ViewModels/TokenViewControllerViewModel.swift
```diff
@@ -3,13 +3,16 @@
 import Foundation
 import UIKit
 import TrustKeystore
+import web3swift
+import Alamofire
+import PromiseKit
 
-struct TokenViewControllerViewModel {
+class TokenViewControllerViewModel {
```
hboon: Can you restore this to `struct` and the `recentTransactions` to `let` if you don't need them to be mutable anymore? I am a bit wary because there is 1 case in our codebase where there is a bug triggered simply by changing the view model from class -> struct, or vice versa. I don't remember which one, but I added a comment to that class/struct.

So just to be safe :)

AlphaWallet/Transactions/Coordinators/SingleChainTransactionDataCoordinator.swift
```diff
@@ -140,7 +152,8 @@ class SingleChainTransactionDataCoordinator: Coordinator {
 
         let startBlock: Int
         let sortOrder: AlphaWalletService.SortOrder
-        if let newestCachedTransaction = storage.completedObjects.first {
+        //Do not use the transactions block number if it has been retrieved by erc20 event logs
+        if let newestCachedTransaction = storage.completedObjects.first, !newestCachedTransaction.isERC20Interaction {
```
hboon: I think you'll have to do a loop of some sort (not efficient) or create a new property (like completedObjects, but add a filter for isERC20Interaction so this line is simply changed to (with a better name than `completedObjectsThatAreNotERC20Interaction` hopefully :) ):

```
if let newestCachedTransaction = storage.completedObjectsThatAreNotERC20Interaction.first, !newestCachedTransaction.isERC20Interaction {
```

Otherwise, if the last transaction in the database is `isERC20Interaction=true`, then it'll (wrongly) fetch the oldest transaction that it can (blockNumber 1)

AlphaWallet/Tokens/ViewModels/TokenViewControllerViewModel.swift
hboon: Ah. It's: `TokensViewModel`

James-Sangalli:
@hboon all good now, please build it on your device just to double check the migration is ok :)

AlphaWallet/Transactions/Storage/TransactionsStorage.swift
```diff
@@ -28,6 +28,10 @@ class TransactionsStorage {
         return objects.filter { $0.state == .completed }
     }
 
+    var transactionObjectsThatDoNotComeFromEventLogs: [Transaction] {
+        return objects.filter { !$0.isERC20Interaction }
```
hboon: Can you change this to (untested code):

```
        return objects
            .filter("state == \(TransactionState.completed.rawValue)")
            .filter("isERC20Interaction == false")
```

The non-string version is ran as normal Swift collection, so the filter() call forces the entire transactions dataset to be loaded in memory and every Transaction instance runs the block against it.

The string version is run natively within Realm without having to load the entire transactions dataset.

(I know the `completedObjects` still use the block version, we inherited them and we haven't fix it yet)

hboon: @James-Sangalli Updated the property to:

```
    var transactionObjectsThatDoNotComeFromEventLogs: Results<Transaction> {
        return realm.objects(Transaction.self)
                .sorted(byKeyPath: "date", ascending: false)
                .filter("chainId = \(self.server.chainID)")
                .filter("id != ''")
                .filter("state == \(TransactionState.completed.rawValue)")
                .filter("isERC20Interaction == false")
    }
```

Some duplicate in the filtering, but I didn't want to make you have to make additional changes to the existing `object` property in the PR.

PR #1218, opened 2019-05-10
remove redundant todo and reference where it is being done

James-Sangalli:


hboon:
@James-Sangalli I think `TokensDataStore.refreshBalanceForERC721Tokens()` might already be doing this when it's fetching the balance. The TODO is outdated?

James-Sangalli:
@hboon touche, we make this PR simply remove the todo and reference where it is done. Haha.

PR #1225, opened 2019-05-20
sets the ethereum window object

James-Sangalli:
This may have been causing issues on multiple dapp sites.


hboon:
Related to: #1225

Issue #987, opened 2019-01-06
DApp https://showcase.funfair.io/account


zhangzhongnan928:
Not working

Issue #901, opened 2018-11-13
Support EIP1102 aka. Privacy Mode or at least test to verify it doesn't break dapp browser


hboon:
https://eips.ethereum.org/EIPS/eip-1102

https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

hboon:
https://github.com/James-Sangalli/xcontract/issues/6 would provide a good test case.

James-Sangalli:
@hboon this is supported in a branch on xcontract but the web3 beta api seems to be riddled with bugs, goes hand in hand with issue #877

James-Sangalli:
@hboon this is dependent on the dapp itself, currently cryptokitties has this feature and it works without asking on our wallet (we probably need to upgrade web3 to get that to ask for permission)

James-Sangalli:
@hboon another note, unless privacy mode is enabled on the client, the default way will continue. We could potentially add this feature in later

hboon:
@James-Sangalli re-opening

PR #1226, opened 2019-05-22
Add enable function to dapp browser

James-Sangalli:
Should solve issues like #987 but doesn't seem to.


hboon:
Also part of #901

PR #1234, opened 2019-06-08
Upgrade fstream for security

James-Sangalli:


James-Sangalli:
@hboon I believe it is for crowdin, the language service.

Issue #1244, opened 2019-06-18
Provide a Clear Browser Cache function in the setting page


zhangzhongnan928:
https://github.com/AlphaWallet/alpha-wallet-ios/issues/1224

hboon:
- [ ] Implement
- [ ] Verify #1224 works for Victor (hopefully without having deleting the app since)

PR #1246, opened 2019-06-18
Add clear dapp browser cache action to Settings tab

hboon:
For #1244


hboon:
A simple way to test manually if it works.

1. Visit a particular URL in dapp browser
2. Open web inspector and run `localStorage.x` = "I set this"
3. Run `localStorage.x` to verify it's there
4. Tap Settings > Clear Dapp Brower cache
5. Repeat (3) and it should be nil.

PR #1269, opened 2019-06-28
Only show the Redeem action button for FIFA ticket tokens and not for other ERC875 tokens

hboon:
Before this PR, all ERC875 tokens including those with TokenScript files marketing `interface="erc875"` includes a redeem action.

After applying this PR, only FIFA tokens (with the 2 hardcoded contract addresses) will have the redeem action.


hboon:
Yeah. We'll need to make the Usher app available and support it though.

Issue #1265, opened 2019-06-23
HD wallet support


colourful-land:
## Which HD library to use?

This relates to #1118 - we must evaluate the path to support HD wallet and the possibility of using Trust Wallet Core should be investigated. I filed it in iOS because @JamesSmartCell is away and we are on the edge of starting backup/recovery project by Tomek.

## Support only English seed phrase with the vision of supporting more down the way

We enforce the use of English seed phrases. Despite that seed phrases are [polyglot](https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md
). Reason:

- We assume that Chinese input methods always send the combination to the server (e.g. when the user writes an ideograph in a drawing area)
- we are concerned putting the keyboard to password mode prevents typing of umlauts like ë

Tomek is going to develop a safe seed phrase recovery tool which will allow non-English. But it will be a separate task.

JamesSmartCell:
Existing web3j issue for HD wallet support https://github.com/web3j/web3j/issues/533

JamesSmartCell:
We could implement this in web3j and fulfil our community service! However it has to be done very carefully.

colourful-land:
> We could implement this in web3j and fulfil our community service! However it has to be done very carefully.

We prefer paying them to do so than using our own bandwidth. We will probably fail the goal of becoming the wallet of next 1M blockchain users even if we didn't split our focus.

James-Sangalli:
@colourful-land @JamesSmartCell why not use Trust's implementation?

colourful-land:
> @colourful-land @JamesSmartCell why not use Trust's implementation?

Good question

James-Sangalli:
@colourful-land @hboon @JamesSmartCell 

He uses a seed wallet and derives multiple coins by choosing paths far away from each other. Supporting this kind of scheme is much more practical for the future when there are other coins.

Example code (iOS):

1. https://github.com/TrustWallet/trust-keystore/blob/master/Tests/KeyStoreTests.swift (notice the derivationPath: label to derive the path)
2. https://github.com/TrustWallet/trust-keystore/blob/master/Sources/KeyStore.swift#L45
3. https://github.com/TrustWallet/trust-keystore/blob/master/Sources/KeyStore.swift#L67
4. (legacy keystore importing): https://github.com/TrustWallet/trust-keystore/blob/master/Sources/KeyStore.swift#L81

[...]

PR #1272, opened 2019-07-01
Removing Keystore files and adding HD Wallet support

hboon:
Implements #1265 

Current State
---
We only support raw Ethereum private keys and store them as keystore files with the AES encryption key being stored in the keychain

Goals
---
There are 3 goals:

- [x] Store raw private keys directly in the keychain, doing away with writing keystore files to the file system and app-level AES encryption of the private key
- [x] Support multiple HD Wallets, each limited to just the Ethereum address at account index 0
- [x] Support new backup alerts UI

Wallets are still restricted to Ethereum only. Relevant UI (such as import/export wallets) has to be updated.

Added a 2 more items:

- Store seed and raw private keys using symmetric encryption based on Diffie-Hellman with ECDSA keys stored in Secure Enclave
- Screen prompting user to confirm locking the seed/private key


hboon:
After the discussion on 20190704, we might still use AES after all, but still write the encrypted private key directly to the keychain instead of writing keystore files on disk.

hboon:
Keeping a list of additions/changes here on top of the specs/visual design:

* Added a "Watched" section alongside "HD Wallets" and "Keystore Wallets"
* Retained the wording "Backup Encrypted Private Key" instead of "Export Keystore JSON" because we thought mentioning keystore and JSON might be confusing, but it's a bit inconsistent. So someone file issues this doesn't look right.

hboon:
Sorry for the huge PR @James-Sangalli.

The changes in logic are in 2 commits:

* e85a27c - which adds HD wallet support (account 0 only) and stores private keys directly in the keychain instead encrypted keystore files
* fc78038 - makes the keychain entries for seed phrase and private keys require user presence to unlock

Others commit are UI related to HD wallet support and the new backup alerts.

hboon:
Rebased master so it's easier to merge.

hboon:
Rebased master so it's easier to merge.

hboon:
Note that Face ID/Touch ID is not accessible when the PIN (i.e iOS Passcode) is disabled.

2 notable scenarios that was coded for:

A.

1. With passcode enabled, create account
2. Disable passcode
3. Access key (sign and back up)
4. Enable passcode
5. Access key (sign and back up)

B.

1. With passcode disabled, create account
2. Access key (sign and back up)
3. Enable passcode
4. Access key (sign and back up)

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+import Result
+import TrustWalletCore
+
+enum EtherKeystoreError: LocalizedError {
+    case protectionDisabled
+}
+
+///We use ECDSA keys, achieving symmetric encryption based on Diffie-Hellman to encrypt the HD wallet seed and raw private keys and store the ciphertext in the keychain
+///
+///The ECDSA key is stored in the secure enclave, not requiring user presence
```
n1v0lg: Why is the user presence requirement added to the ciphertext and not the ECDSA key? We are weakening security by allowing usage of the ecdsa key without authentication rather than the ciphertext. If possible, require user presence for both; if only one is possible, key is preferable to ciphertext.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+            let address = AlphaWallet.Address(fromPrivateKey: privateKey)
+            let hasEthereumAddressAlready = wallets.map({ $0.address }).contains {
+                $0.sameContract(as: address)
+            }
+            guard !hasEthereumAddressAlready else {
+                return .failure(.duplicateAccount)
+            }
+            let seed = HDWallet.computeSeedWithChecksum(fromSeedPhrase: mnemonicString)
+            if isUserPresenceCheckPossible {
+                let isSuccessful = saveSeedForHdWallet(seed, forAccount: .init(address: address), withUserPresence: false)
```
n1v0lg: Could the flow be simplified? In particular, instead of storing without user presence first, and then trying to store with user presence, why not check for user presence first and then store accordingly?

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+                let isSuccessful = savePrivateKeyForNonHdWallet(privateKey, forAccount: .init(address: address), withUserPresence: false)
+                guard isSuccessful else { return .failure(.failedToCreateWallet) }
+                let _ = savePrivateKeyForNonHdWallet(privateKey, forAccount: .init(address: address), withUserPresence: true)
+            } else {
+                let isSuccessful = savePrivateKeyForNonHdWallet(privateKey, forAccount: .init(address: address), withUserPresence: false)
+                guard isSuccessful else { return .failure(.failedToCreateWallet) }
+            }
+            addToListOfEthereumAddressesWithPrivateKeys(address)
+            return .success(Wallet(type: .real(.init(address: address))))
+        case .mnemonic(let mnemonic, _):
```
n1v0lg: Unlike in the Android version, the non-user authenticated ciphertext does not get deleted after the authenticated version gets stored. Is this intentional? Presumably, if the user imported the seed, he has access to it and we therefore don't need to worry about maintaining a non-user presence protected backup anymore. 

Is the reason that the importWallet method is also used when a new seed is created? In that case, I suggest using two separate methods, or using a flag to indicate that the non-user-presence backup should be removed if the user is importing a wallet from an existing seed (as opposed to creating a new one).

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+                //Read it back, forcing iOS to check for user-presence
+                isSuccessful = getPrivateKeyFromNonHdWallet(forAccount: account, prompt: prompt, withUserPresence: true) != nil
+            }
+        }
+        if isSuccessful {
+            addToListOfEthereumAddressesProtectedByUserPresence(account.address)
+            let secureEnclave = SecureEnclave()
+            if isHdWallet(account: account) {
+                secureEnclave.deletePrivateKeys(withName: encryptionKeyForSeedLabel(fromAccount: account, withUserPresence: false))
+                keychain.delete("\(Keys.ethereumSeedUserPresenceNotRequiredPrefix)\(account.address.eip55String)")
```
n1v0lg: Can secureEnclave.deletePrivateKeys throw an exception? In that case, we will never execute `keychain.delete`. Ideally, the second deletion should not be contingent on the previous one succeeding, i.e., we should delete as much as possible rather than "aborting" if a preceding operation fails.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+        }
+        let data = keychain.getData("\(prefix)\(account.address.eip55String)", prompt: prompt)
+                .flatMap { decryptHdWalletSeed(fromCipherTextData: $0, forAccount: account, withUserPresence: withUserPresence) }
+                .flatMap { String(data: $0, encoding: .utf8) }
+        //We copy the record that doesn't require user-presence make a new one which requires user-presence and read from that. We don't want to read the one without user-presence unless absolutely necessary (e.g user has disabled passcode)
+        if data == nil && withUserPresence && shouldWriteWithUserPresenceIfNotFound && keychain.lastResultCode == errSecItemNotFound {
+            if let seedWithoutUserPresence = getSeedForHdWallet(forAccount: account, prompt: prompt, withUserPresence: false, shouldWriteWithUserPresenceIfNotFound: false) {
+                saveSeedForHdWallet(seedWithoutUserPresence, forAccount: account, withUserPresence: true)
+            }
+            return getSeedForHdWallet(forAccount: account, prompt: prompt, withUserPresence: true, shouldWriteWithUserPresenceIfNotFound: false)
```
n1v0lg: A little confused by the flow here. Why are we reading, writing, and then re-reading and returning what we just wrote? Is that to ensure the writing was successful? A more natural flow would be to simply return the seed we read the first time around.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+        case .watch(let address):
+            removeAccountFromBookkeeping(.init(address: address))
+        }
+        (try? LegacyFileBasedKeystore())?.delete(wallet: wallet)
+        return .success(())
+    }
+
+    private func deletePrivateKeysFromSecureEnclave(forAccount account: EthereumAccount) {
+        let secureEnclave = SecureEnclave()
+        secureEnclave.deletePrivateKeys(withName: encryptionKeyForPrivateKeyLabel(fromAccount: account, withUserPresence: true))
```
n1v0lg: Can exceptions be thrown here? This would prevent deletion in the subsequent lines. Make all deletions not contingent on each other, i.e., delete as much as possible rather than "aborting" if a preceding operation fails.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+        secureEnclave.deletePrivateKeys(withName: encryptionKeyForPrivateKeyLabel(fromAccount: account, withUserPresence: false))
+        secureEnclave.deletePrivateKeys(withName: encryptionKeyForSeedLabel(fromAccount: account, withUserPresence: true))
+        secureEnclave.deletePrivateKeys(withName: encryptionKeyForSeedLabel(fromAccount: account, withUserPresence: false))
+    }
+
+    private func deleteKeysAndSeedCipherTextFromKeychain(forAccount account: EthereumAccount) {
+        keychain.delete("\(Keys.ethereumRawPrivateKeyUserPresenceNotRequiredPrefix)\(account.address.eip55String)")
```
n1v0lg: Can exceptions be thrown here? This would prevent deletion in the subsequent lines. Make all deletions not contingent on each other, i.e., delete as much as possible rather than "aborting" if a preceding operation fails.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+    func isKeystore(wallet: Wallet) -> Bool {
+        switch wallet.type {
+        case .real(let account):
+            return ethereumAddressesWithPrivateKeys.contains(account.address.eip55String)
+        case .watch:
+            return false
+        }
+    }
+
+    func isWatched(wallet: Wallet) -> Bool {
```
n1v0lg: What exactly does isWatched mean?

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+    func signPersonalMessage(_ message: Data, for account: EthereumAccount) -> Result<Data, KeystoreError> {
+        let prefix = "\u{19}Ethereum Signed Message:\n\(message.count)".data(using: .utf8)!
+        return signMessage(prefix + message, for: account)
+    }
+
+    func signHash(_ hash: Data, for account: EthereumAccount) -> Result<Data, KeystoreError> {
+        guard let key = getPrivateKeyForSigning(forAccount: account) else { return .failure(.accountMayNeedImportingAgainOrEnablePasscode) }
+        do {
+            var data = try EthereumSigner().sign(hash: hash, withPrivateKey: key)
+            // TODO: Make it configurable, instead of overriding last byte.
```
n1v0lg: Fix TODO.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+        guard let key = getPrivateKeyForSigning(forAccount: account) else { return .failure(.accountMayNeedImportingAgainOrEnablePasscode) }
+
+        do {
+            var messageHashes = [Data]()
+            for i in 0...data.count - 1 {
+                let hash = data[i].sha3(.keccak256)
+                messageHashes.append(hash)
+            }
+            var data = try EthereumSigner().signHashes(messageHashes, withPrivateKey: key)
+            // TODO: Make it configurable, instead of overriding last byte.
```
n1v0lg: Fix TODO.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -0,0 +1,647 @@
...
+                transaction.data,
+                v, r, s,
+            ])!
+            return .success(data)
+        } catch {
+            return .failure(.failedToSignTransaction)
+        }
+    }
+
+    //TODO should and can we check if the keychain has that entry without prompting for user-presence?
```
n1v0lg: Fix TODO.

AlphaWallet/KeyManagement/SecureEnclave.swift
```diff
@@ -0,0 +1,176 @@
...
+            return 0
+        }
+    }
+
+    private func encrypt(plainTextData: Data, withPublicKey publicKey: SecKey) throws -> Data {
+        guard SecKeyIsAlgorithmSupported(publicKey, .encrypt, algorithm) else {
+            throw Error.encryptionNotSupported(algorithm: algorithm)
+        }
+        var error: Unmanaged<CFError>?
+        guard let cipherTextData = SecKeyCreateEncryptedData(publicKey, algorithm, plainTextData as CFData, &error) as? Data else {
```
n1v0lg: What happens if the seed is larger than the blocksize with which we are encrypt using ECDSA? Is an error thrown or should this be checked explicitly before the call? Some comment goes for `decrypt` call.

AlphaWallet/KeyManagement/EtherKeystore.swift
James-Sangalli: User is observing an address for which he does not have a key.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: Both the ECDSA key and ciphertext now requires user presence.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: We do check if user presence is available first (but not ask user to verify user presence, otherwise creating each wallet as well as silently migrating old keystore wallets before this PR will require user to verify user-presence).

And if user presence is available (i.e. at least the iOS passcode is set), we save the set that doesn't require user presence first. We don't expect it to fail (but to be sure, we check the result `isSuccessful` the first time to be defensive), and if it works, then we proceed to save the version that requires user-presence. It doesn't matter if the set that requires user presence fails at this point.

This looks a bit tricky but it is because we don't get notified via a hook when user-presence is enabled/disabled. i.e the code handles this case:

1. User deletes iOS passcode
2. Creates a wallet (only k' can be written since there's no passcode)
3. Use the wallet, will attempt to run like (5), but ends up accessing k', as iOS wouldn't/can't prompt to verify user presence because there's no passcode
4. User enables iOS passcode
5. Use the wallet. We know user presence check is available. Tries to access k, fails because it's absent, so we use k' to write k and read k again. This time k works and user is prompted for user presence (as k is read).

For (5), We can't use the result from k' for wallet access right away because we want the user to be prompted to verify user presence, hence we read k again (after writing it, but we are writing k because we don't have it)

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: Yes.

Upon importing a set of keys, the wallet key should still be accessible without user-presence (in case the user disables biometrics). It is only after the user choose to elevate security to make it go to "green bar" (refer to https://github.com/AlphaWallet/TokenScript/wiki/General-Nomenclature#green-bar-for-wallet-security-level for the odd, but unambiguous terminology)

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: No, `SecureEnclave.deletePrivateKeys` will not throw an exception that can be caught (like Java). If it does, it will have to marked with the `throws` keyword in its signature, like:

```
func deletePrivateKeys(withName name: String) throws {}
```

And the compiler will force it to be handled (do-catch, like Java's try-catch) or explicitly ignored with a `try?` like:

```
let result = try? secureEnclave.deletePrivateKeys(...) //ignores error if there's one, `result` will be nil
```

Or power through and crash upon error with `try!`:

```
let result = try! secureEnclave.deletePrivateKeys(...) //crash if error is thrown
```

Or propogate it upstream with `try`:

```
let result = try secureEnclave.deletePrivateKeys(...) //have to handle it with do-catch locally or mark function with `throws`
```

It is also possible that a Swift function generates a runtime error (corruption of the stack for example), but those typically result in a crash and we can't recover from them programmatically other than grabbing a crash log.

So Swift is more of a very safe C, but not as safe as VM-based languages like Java in this aspect.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: I'll reuse the notation k to refer to the ECDSA key that requires user presence to access and k' to refer to the one that doesn't require user presence and just omit the corresponding ciphertext (they are accessed at the same time as k and k'):

We always try to access the wallet via k. This attempt to access the data will cause iOS the prompt the user to verify user presence.

In the line preceding the if-else, k is accessed. The `if` checks that access has failed because the k doesn't exist (instead of the user cancelling the operation or failing to prove user presence), so we attempt to read k' and write k then try to read k again.

The `else` only gets accessed if user-presence is not required.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: No. As above.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: No. As above.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: It's a leftover from old code before the PR. I think we'll always have this hardcoded for Ethereum signatures. I'll fix it outside of the PR.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: It's a leftover from old code before the PR. I think we'll always have this hardcoded for Ethereum signatures. I'll fix it outside of the PR.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: Have removed the TODO. Not necesssary anymore

AlphaWallet/KeyManagement/SecureEnclave.swift
hboon: The docs says that `SecKeyCreateEncryptedData` and `SecKeyCreateDecryptedData` will return nil upon error. We check that and throw a Swift exception (which we then handle in the UI client code) when it fails. But I tested encrypting a long plaintext and decrypting the result and at least it works empirically.

hboon:
Hi @n1v0lg. I've added 3 commits to resolve the issues you mentioned. Please check again.

@James-Sangalli don't merge yet. I'll rebase this PR and the other related PRs when it's done.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -11,10 +11,13 @@ enum EtherKeystoreError: LocalizedError {
...
 ///
-///The ECDSA key is stored in the secure enclave, not requiring user presence
-///The ciphertext is stored in the keychain, not requiring user presence at first, but once user has backed up, requires user presence
+///There are 2 sets of (ECDSA key and ciphertext) for each Ethereum raw private key or HD wallet seed. 1 set is stored requiring user presence for access and the other doesn't. The second set is needed to ensure the user has does not lose access to the Ethereum raw private key (or HD wallet seed) when they delete their iOS passcode. Once the user has verified that they have backed up their wallet, they can choose to elevate the security of their wallet which deletes the set of (ECDSA key and ciphertext) that do not require user-presence.
+///
+///Technically, having 2 sets of (ECDSA key and ciphertext) for each Ethereum raw private key or HD wallet seed may not be required for iOS. But it is done:
+///(A) to be confident that we cause the user to lose access to their wallets and
```
jot2re: Don't you mean "to be confident that we _don't_ cause the user to lose access"? ☺️

AlphaWallet/KeyManagement/EtherKeystore.swift
jot2re: Thanks for clarifying Boon. The only other comment I have is that it might make sense to change the new comments you added to the code to not talk about a PR, since that is a bit out of context for a code comment.

AlphaWallet/KeyManagement/EtherKeystore.swift
jot2re: Thanks for clarifying

AlphaWallet/KeyManagement/EtherKeystore.swift
jot2re: Thanks for clarifying, Boon. The only other comment I have is that it might make sense to change the new comments you added to the code to not talk about a PR, since that is a bit out of context for a code comment.

AlphaWallet/KeyManagement/EtherKeystore.swift
jot2re: Again, thanks for clarifying.

AlphaWallet/KeyManagement/EtherKeystore.swift
jot2re: Thanks for explaining. This explanation might also be valuable to add as a code comment.

hboon:
@James-Sangalli good to merge.

hboon:
Thanks @jot2re and @n1v0lg !

James-Sangalli:
Yay

PR #1278, opened 2019-07-07
help with UX by using UserExperior

James-Sangalli:
For https://github.com/AlphaWallet/web-service/issues/46


AlphaWallet/Info.plist
```diff
@@ -2,6 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>App Transport Security Settings</key>
```
hboon: The key/value is invalid here. But you can remove them because it's already included at https://github.com/AlphaWallet/alpha-wallet-ios/blob/894d99f4ca06b9e95573fac41b47b2d71f99fd80/AlphaWallet/Info.plist#L81

hboon:
Build is failing at Travis

James-Sangalli:
Slightly concerned about this because it may drain the users data and it may be a privacy problem. We would obviously block the recorder for sensitive screens but maybe we don't even want to see what else they get up to in the browser for example. 

We will discuss this further before we release this, if we even do.

hboon:
Good point. We can also consider to limit this to special TestFlight builds if we ultimately decide not to put it on the App store.

James-Sangalli:
Consumes too much resources and is a little creepy.

PR #1284, opened 2019-07-11
Don't block UI while fetching ERC20 transactions

hboon:
Noticed this while working on #1283 and comparing `getErc20Interactions()` and `getContractList()`.

This PR makes `getErc20Interactions()` process the JSON result in a background queue like `getContractList()`. Before this PR, `getErc20Interactions()` might cause a slowdown in the UI depending on how many transactions the web API returns as well as how many networks are enabled in the app. (this call is made once per enabled network).

@James-Sangalli this might be 1 of the cause for slowdown in Victor's phone, but I don't think it's the only one. Can you try it again after merging?


hboon:
@James-Sangalli Thanks!

PR #1304, opened 2019-07-17
guards against null in address filter

James-Sangalli:
This unwrapped a null when I was using the phone. Should be guarded.


James-Sangalli:
@hboon no way to know because that issue has zero detail in it!

PR #1330, opened 2019-08-05
update to July schema

James-Sangalli:
Uses July schema but doesn't fully implement vars like bytes (this is ok for now)

Will enable compound to work.


hboon:
@James-Sangalli I’ll check it out, but can we not use the default: label? Spelling all the cases out is verbose, but using default always ends up causing a bug at some point when we add another enum case or want to change the behavior for one of the existing enum cases because the compiler can’t help to look for it.

We can use case .a, .b, .c: if they are handled the same way to reduce the verbosity a little.

AlphaWallet/TokenScriptClient/Models/TokenScript.swift
```diff
@@ -14,8 +14,8 @@ extension TokenScript {
     static let repoClientName = "AlphaWallet"
     static let repoPlatformName = "iOS"
     static let tokenScriptNamespacePrefix = "http://tokenscript.org/"
-    static let oldNoLongerSupportedTokenScriptNamespaceVersions = ["2019/04"].map { "\(tokenScriptNamespacePrefix)\($0)/tokenscript" }
-    static let supportedTokenScriptNamespaceVersion = "2019/05"
+    static let oldNoLongerSupportedTokenScriptNamespaceVersions = ["2019/05"].map { "\(tokenScriptNamespacePrefix)\($0)/tokenscript" } + ["2019/04"].map { "\(tokenScriptNamespacePrefix)\($0)/tokenscript" }
```
hboon: This can be:

```
static let oldNoLongerSupportedTokenScriptNamespaceVersions = ["2019/04", "2019/05"].map { "\(tokenScriptNamespacePrefix)\($0)/tokenscript" }
```

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -367,7 +367,7 @@ private class PrivateXMLHandler {
             }
         case .unsupportedTokenScriptVersion(let isOld):
             if isOld {
-                assertImpossibleCodePath()
+                //assertImpossibleCodePath()
```
hboon: Does the code reach here at runtime? Sounds like a bug.

AlphaWallet/TokenScriptClient/Models/TokenIdOrigin.swift
```diff
@@ -31,10 +31,14 @@ struct TokenIdOrigin {
...
         case .e18:
             return EtherNumberFormatter().number(from: String(number)).flatMap { .uint(BigUInt($0)) }
+        case .e8:
+            return EtherNumberFormatter().number(from: String(number)).flatMap { .uint(BigUInt($0)) }
         case .bool:
             return .bool(number != 0)
         case .void:
             return nil
+        default:
+            return nil
```
hboon: Should we support functions returning `.bytes`, etc?

AlphaWallet/TokenScriptClient/Models/AssetAttributeValueUsableAsFunctionArguments.swift
```diff
@@ -47,6 +47,8 @@ enum AssetAttributeValueUsableAsFunctionArguments {
             return coerceToUInt(forFunctionType: functionType)
         case .void:
             return nil
+        default:
```
hboon: Needs to coerce them the arguments to the correct type. It might need to be 1 or 2 different types (likely 1), depending on what the function call is used for based on `functionType`. `coerceToAddress` is an example of it being coerced into 2 different types. The reason is because we, unfortunately, use 2 different encoders for calls and transactions.

AlphaWallet/TokenScriptClient/Models/CallForAssetAttributeCoordinator.swift
```diff
@@ -83,6 +83,9 @@ class CallForAssetAttributeCoordinator {
                     case .void:
                         //Don't expect to reach here
                         seal.fulfill(.bool(false))
+                    default:
+                        //Don't expect to reach here
+                        seal.fulfill(.bool(false))
```
hboon: Need to convert `.bytes` etc to the right return value.

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
James-Sangalli: Yes, if it is old or too new it will hit the assert and crash the app

AlphaWallet/TokenScriptClient/Models/TokenIdOrigin.swift
James-Sangalli: Yes, I can add that in now

James-Sangalli:
@hboon casted the bytes but need to be wary that bytes and string are very similar (dynamic length) but string is for utf8.

Also noticed that uint is handled but not all the uintN types, same for bytesN, we should add those in too.

James-Sangalli:
@hboon should be good to go.

hboon:
@James-Sangalli can you replace the `default:` labels with explicit ones? More verbose, but much more future-proof.

James-Sangalli:
Let's revisit this once we have a newer schema to address the concerns we had with bytes.

PR #1331, opened 2019-08-07
Fixed filenames to compile on case-sensitive drives. 

jzaki:
(omitted Podfile.lock diffs)


jzaki:
> @jzaki thanks for the PR. Is it not compiling before the PR because the Xcode project file spells the 2 files the way it does after your PR is applied?

The harddrive I use is formatted as case-sensitive, and so the project files referred to (correct spelling) did not refer to files that existed (files had incorrect spelling). Once I renamed the files to what the xcode project referred to, it finds them and compiles successfully (Xcode 10.3)

Notice the error occurs after all tests pass: https://travis-ci.com/AlphaWallet/alpha-wallet-ios/builds/122256356#L11685-L11694
Do all tests pass for you locally? I'm in the process of getting xcode 10.1 to test as Jenkins does, some of the pods receive an old version (although the Podfiles doesn't seem to force it), and also the post_install to refer some pods to swift version 4 doesn't affect the file that Xcode 10.3 took issue with (locally I manually pointed SWXMLHash to swift version 4.0).

Will update with how Xcode 10.1 builds/tests locally.

hboon:
Ah, it's just Travis messing with us as usual. It passes once I re-ran the tests. Thanks!

hboon:
@jzaki Can you rebase master and force push the branch and we can merge. Something like:

```
git checkout fix-character-case-in-names
git fetch upstream master:master #Replace "upstream" with the remote name you have for this https://github.com/AlphaWallet/alpha-wallet-ios
git rebase master
git force push upstream fix-character-case-in-names #replace "upstream"
```

Issue #1314, opened 2019-07-19
Invalid signature on native token cards 


James-Sangalli:

hboon:
@James-Sangalli can you elaborate on this? A reproducible case?

James-Sangalli:
@hboon simply open up the ethereum, xdai etc native cards and you will see "invalid signature"

hboon:
@James-Sangalli  I know you mentioned previously that it fails for debug TSMLs, but when I try with the official Dai TSML, I see this. Looks good to me. Does it fail to validate when you test it?

<img width="482" alt="Screenshot 2019-07-23 at 11 00 08 PM" src="https://user-images.githubusercontent.com/56189/61722912-b5053980-ad9d-11e9-988c-5c671d51a327.png">

PR #1332, opened 2019-08-08
Fix: TokenScript files with invalid signatures are wrongly matched, always to mainnet

hboon:
For #1314


James-Sangalli:
@hboon test fails for attribute extraction

hboon:
@James-Sangalli Test should be fixed now. Also rebased master.

James-Sangalli:
Thanks

hboon:
> Thanks

Thanks for pointing it out! I keep looking at the wrong place.

Issue #976, opened 2018-12-26
Enhance UI/UX of the currency/asset search bar 


farrahfazirah:
When a user clicks on the search bar, an empty space appears in between the search bar and cards.

<div align="center"><img src="https://user-images.githubusercontent.com/20555752/50440652-28db8280-0932-11e9-9602-44c7d2beadec.jpg" height="350px" width="200px"></div>

<br>

Here is my suggestion below to improve the look and experience of the search bar. Ideally, the search bar should remain static when interacted by the user.

<br>

<div align="center"><img src="https://user-images.githubusercontent.com/20555752/50440869-0ac25200-0933-11e9-82ce-c6839a60cca0.png" height="550px" width="250px"></div>

<br>

Modifications:
- "Cancel" is removed and replaced with "x" icon. The "x" icon appears on the far right, inside the search bar when user clicks on the search bar.
- Replaced the search icon
- Changed the font type and colour
- Changed the UX copy of search bar
- Changed border radius of search bar

<a href="https://invis.io/ZPPRHSMUXQV">Please see this InVision board for the prototyping preview</a>

hboon:
Done with #1337

PR #1318, opened 2019-07-23
Add back up this wallet action to settings screen

hboon:
Part of the redesign in https://community.tokenscript.org/t/design-manage-wallets-keys/239

Based on #1317, so after #1317 is merged, we can rebase and merge this PR.

PR #1337, opened 2019-08-11
Improve search UI

hboon:
For #976. This PR removes the blue patch of empty space below the search bar. Had to do quite a bit of work to make it work correctly. No UI design change otherwise since it might be redesigned.


hboon:
Building on #1318 because there are changes that affect the UI changes introduced in HD wallet support; so holding this as a draft PR till the former is merged.

hboon:
@James-Sangalli good to merge after #1318

Issue #1347, opened 2019-08-14
Auto correct should be disabled when exporting keystore JSON


James-Sangalli:

hboon:
Folded into #1272

PR #1353, opened 2019-08-14
Fix: disable autocorrect when entering password for Keystore JSON export

hboon:
For #1347 

Based on #1272. Not pushing to that PR for now since a review is ongoing. Will clean up and rebase the 3 affected PRs once we clear security review.


hboon:
Included in #1272

PR #1357, opened 2019-08-16
fix server not initialized

James-Sangalli:


hboon:
@James-Sangalli great spotting! It has to set the server both before and after validation though. Have added a commit. If that looks good, feel free to rebase and merge.

PR #1365, opened 2019-08-21
Fix: deleted tokens are automatically added back when we detect ERC20 transactions involving them

hboon:
This is because `TransactionsStorage.tokens(from:)` performs an update/addition to the tokens database with:

```
let operation = transaction.localizedOperations.first,
let contract = operation.contractAddress,
```

updating/adding a token with `contract` as the contract address.

@James-Sangalli  But maybe we need to check both `to` and `operation?.contractAddress` when filtering? Is the former still needed to handle when there's an Ether transfer to the contract?

ie. the PR changes:

```
guard let addressToCheck = AlphaWallet.Address(string: $0.to) else { return false }
return !contractsToAvoid.contains(addressToCheck)
```

to:

```
guard let addressToCheck = $0.operation?.contractAddress else { return false }
return !contractsToAvoid.contains(addressToCheck)
```

Maybe we need to check both `AlphaWallet.Address(string: $0.to)` and `$0.operation?.contractAddress` instead?

### DIFF
--- a/AlphaWallet/Transactions/Coordinators/SingleChainTransactionDataCoordinator.swift
+++ b/AlphaWallet/Transactions/Coordinators/SingleChainTransactionDataCoordinator.swift
@@ -96,17 +96,30 @@ class SingleChainTransactionDataCoordinator: Coordinator {
     }
 
     private func update(items: [Transaction]) {
+        let transactionsToPullContractsFrom = filterTransactionsToPullContractsFrom(items)
+        storage.add(transactions: items, transactionsToPullContractsFrom: transactionsToPullContractsFrom)
+        delegate?.handleUpdateItems(inCoordinator: self)
+    }
+
+    private func filterTransactionsToPullContractsFrom(_ transactions: [Transaction]) -> [Transaction] {
         let alreadyAddedContracts = tokensStorage.enabledObject.map { $0.contractAddress }
         let deletedContracts = tokensStorage.deletedContracts.map { $0.contractAddress }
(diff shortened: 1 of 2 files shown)


James-Sangalli:
@hboon not sure what you mean when referring to the ether transfer?

hboon:
> @hboon not sure what you mean when referring to the ether transfer?

@James-Sangalli oh, I was thinking if you send Ether to a contract, the `to` of that transaction will be the smart contract address.

But then it wouldn't appear in this list of transactions because it's not an ERC20 transaction, so it doesn't have to be checked.

hboon:
@James-Sangalli wait a minute. Let me think about it a bit  more.

hboon:
@James-Sangalli I have changed the filtering to check both `to` and `contractAddress`:

```
return transactions.filter {
    if let toAddressToCheck = AlphaWallet.Address(string: $0.to) {
        if contractsToAvoid.contains(toAddressToCheck) {
            return false
        }
    }
    if let contractAddressToCheck = $0.operation?.contractAddress {
        if contractsToAvoid.contains(contractAddressToCheck) {
            return false
        }
    }
    return true
}
```

This is because this function is used for both ERC20 transactions and the main transaction list that we pulled. So we have to handle ERC20 transfers (the smart contract addres is in `transaction.operation?.contractAddress`) as well as when someone sends Ether to a smart contract (`transaction.to`).

Does that sound right?

James-Sangalli:
@hboon yes, etherscan API fills the to param, often this is a contract (contractAddress is only in the api for creation of a contract)

PR #1378, opened 2019-08-25
Extract recid/v + 27

### LINKED ISSUE #1369: Extract function: data[64] += 27
Extract function: data[64] += 27

https://github.com/AlphaWallet/alpha-wallet-ios/blob/master/AlphaWallet/KeyManagement/EtherKeystore.swift#L438

This same code appears a few times and I want to make sure it's more intention revealing and easier to find if we ever need to tweak it by extracting it into a 1-line function. @colourful-land can you help suggest a name for the function or variable?

(This looks like the v of (r, s, v))

hboon:
Closes #1369 

Simple change. But have to be careful during review @James-Sangalli

### DIFF
--- a/AlphaWallet/EtherClient/TransactionSigning.swift
+++ b/AlphaWallet/EtherClient/TransactionSigning.swift
@@ -55,7 +55,7 @@ struct HomesteadSigner: Signer {
         precondition(signature.count == 65, "Wrong size for signature")
         let r = BigInt(sign: .plus, magnitude: BigUInt(signature[..<32]))
         let s = BigInt(sign: .plus, magnitude: BigUInt(signature[32..<64]))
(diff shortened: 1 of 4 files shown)


hboon:
@James-Sangalli we aren't. And actually there might be a few more possibilities. The ideal way is to refactor so that the +=27 is applied when we sign, but the existing code needs some refactoring [1] to do that safely and to add some tests for it. So this PR is really only extract the magic number `27` and be able to discover them to refactor or fix bugs.

> guard against that scenario before adding it?

We should set up another issue to refactor if we want. This PR doesn't change the logic.

[1] Because there is at least 1 spot where we also apply the += 27 as a fix to a signature we get from MagicLink, as a patch.

PR #1399, opened 2019-08-27
Fix: make confirm prompt for signing, handle long messages

### LINKED ISSUE #1142: Make UI dialog prompting to sign message display long messages correctly (both tall and short phones)
Eg. when creating account on https://playground.counterfactual.com/

hboon:
Fixes #1142 

Before PR | After PR 1 | After PR 2
-|-|-
 |  | 

This is useful for testing typed messages: https://danfinlay.github.io/js-eth-personal-sign-examples/

Should probably have another PR for fixing signing compatibility against the URL above. But have to be careful to keep compatible with Android version.


hboon:
@JamesSmartCell maybe relevant for Android

PR #1401, opened 2019-08-27
Fix: Hide current networks native balance in dapp browser when switching networking screen is shown

hboon:
Fixes broken git merge in #1381


hboon:
@James-Sangalli bah. Git merge fails sometimes. No biggie.

PR #1430, opened 2019-09-07
Fix: TokenScript files should override behavior of ERC721 tokens fetched from OpenSea too

### LINKED ISSUE #1428: TokenScripts should override realm and services like OpenSea

hboon:
Fixes #1428


hboon:
@James-Sangalli yes, it will show it as an ERC721 collectible under the Collectibles tab in Wallet tab like it used to. The difference is when it is tapped, if a TokenScript file is available, it will use the iconified view (even if it’s an empty string) of available as well as offer the TokenScript actions abs action views specified in the file.

When a TokenScript file is applied to an ERC721, there isn’t a way to specify to use the non-TokenScript iconified view [1], nor is there a direct way to pick up and use the token data (like traits) from OpenSea that is already stored in our local database.

Both are probably desirable, especially the former (fallback views). The latter is harder to standardise access in TokenScript.

[1] we briefly talked about having it. Maybe when we specify an empty string or not supplying a tag. But there was no conclusion.

PR #1434, opened 2019-09-07
Add `tokenscript` as a special search keyword in the Wallet tab search bar to filter for tokens with TokenScript file

hboon:
Just like we can now with "ERC 20", "ERC 721" and "ERC 875" before this PR

### DIFF
--- a/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
@@ -5,6 +5,7 @@ import UIKit
 
 //Must be a class, and not a struct, otherwise changing `filter` will silently create a copy of TokensViewModel when user taps to change the filter in the UI and break filtering
 class TokensViewModel {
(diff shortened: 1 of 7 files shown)


AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -111,6 +113,9 @@ class TokensViewModel {
                         return $0.type == .erc721
                     } else if keyword.lowercased() == "erc875" || keyword.lowercased() == "erc 875" {
                         return $0.type == .erc875
+                    } else if keyword.lowercased() == "tokenscript" {
```
James-Sangalli: @hboon better to ignore casing? https://stackoverflow.com/questions/30532728/how-to-compare-two-strings-ignoring-case-in-swift-language

hboon: Yes. It is already using keyword.lowercased()

PR #1437, opened 2019-09-10
ensure overflowed integers on 721 balance count doesn't break

James-Sangalli:
This broke on Victor's phone, check if the max value is reached and if it has fail rather than crash (my try-catch block didn't work on this exception for some reason)

This only effects 721 tokens that are not supported by OpenSea


hboon:
Can you try if comparing against ‘Int.max’ works?

Does this crash happen if we include the recent PR with the migration? (Either way this fix is good to have)

James-Sangalli:
@hboon done. Not sure about the migration (didn't happen on my phone) but it crashed victors

PR #1470, opened 2019-09-24
Fix: navigation bar and status bar color is wrong (black) for Wallet tab when search bar is visible in iOS 13.1

### LINKED ISSUE #1423: Navigation bar in Wallet tab has wrong background color in iOS 13.1 beta
Might just be an issue with the beta. But should revisit and verify.

hboon:
Fixes #1423 

Issue might only be for iOS 13.1 beta (but it has happened for all 4 betas) and iOS 13.1 public release is imminent.

Hackish, but seems safe for other OS versions.


hboon:
@James-Sangalli would you do a submission after this please? The issue is reproducible on the app store version on iOS 13.1 beta.

PR #1473, opened 2019-09-26
update to October schema

James-Sangalli:
- cast bytes & bytesN types
- add missing cases for variables

TODO: handle return types and encoding for bytesN variables


AlphaWallet/TokenScriptClient/Models/TokenIdOrigin.swift
```diff
@@ -21,18 +21,29 @@ struct TokenIdOrigin {
...
+        case .e8:
+            return EtherNumberFormatter().number(from: String(number), decimals: 8).flatMap { .uint(BigUInt($0)) }
+        case .e4:
+            return EtherNumberFormatter().number(from: String(number), decimals: 4).flatMap { .uint(BigUInt($0)) }
+        case .e2:
+            return EtherNumberFormatter().number(from: String(number), decimals: 2).flatMap { .uint(BigUInt($0)) }
         case .bool:
             return .bool(number != 0)
+        case .bytes, .bytes1, .bytes2, .bytes3, .bytes4, .bytes5, .bytes6, .bytes7, .bytes8, .bytes9, .bytes10, .bytes11, .bytes12, .bytes13, .bytes14, .bytes15, .bytes16, .bytes17, .bytes18, .bytes19, .bytes20, .bytes21, .bytes22, .bytes23, .bytes24, .bytes25, .bytes26, .bytes27, .bytes28, .bytes29, .bytes30, .bytes31, .bytes32:
+            let bytesEncodedNumber: BigUInt = (bitmask & tokenId) << bitShift
```
hboon: We don't have to apply the bitmask here again (and it's shifting in the wrong direction :) Should be >>), because it's already done above. Just use `number` like in the `.uint` case above.

Also mapping it to a .address isn't right. It should be just:

```
return .bytes(BigUInt(number))
```

or 

```
return .bytes(String(number, 16))
```

AlphaWallet/TokenScriptClient/Models/TokenIdOrigin.swift
```diff
@@ -21,18 +21,29 @@ struct TokenIdOrigin {
     }
 
     func extractValue(fromTokenId tokenId: TokenId) -> AssetInternalValue? {
-       let number: BigUInt = (bitmask & tokenId) >> bitShift
+        let number: BigUInt = (bitmask & tokenId) >> bitShift
```
hboon: We'll have to do a bitmask and shift in `FunctionOrigin.extractValue()` like we do here in `TokenIdOrigin` here before `resultSubscribable.value = value` to change the value on the RHS.

AlphaWallet/TokenScriptClient/Models/Origin.swift
```diff
@@ -8,9 +8,111 @@ import Kanna
 enum OriginAsType: String {
     case address
     case uint
+    case uint8
```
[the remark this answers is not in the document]
James-Sangalli: @hboon we are supporting String in schema 2019/10 so I will keep that in, I can trim the other extra types off as they are leftover from 2019/07

AlphaWallet/TokenScriptClient/Models/AssetInternalValue.swift
```diff
@@ -7,6 +7,7 @@ import PromiseKit
 enum AssetInternalValue: Codable {
     case address(AlphaWallet.Address)
     case string(String)
+    case bytes(String)
```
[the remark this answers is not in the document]
James-Sangalli: Agree that it's better to use Data, UInt could easily overflow as the maximum is 64 bits.

AlphaWallet/TokenScriptClient/Models/Origin.swift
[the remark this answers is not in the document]
James-Sangalli: Besides, String is very similar to bytes, just an automatic utf8 imposition. Don't see a good reason to not support it in this schema

AlphaWallet/TokenScriptClient/Models/CallForAssetAttributeCoordinator.swift
```diff
@@ -71,6 +71,9 @@ class CallForAssetAttributeCoordinator {
                     case .bool:
                         let result = value as? Bool ?? false
                         seal.fulfill(.bool(result))
+                    case .bytes, .bytes1, .bytes2, .bytes3, .bytes4, .bytes5, .bytes6, .bytes7, .bytes8, .bytes9, .bytes10, .bytes11, .bytes12, .bytes13, .bytes14, .bytes15, .bytes16, .bytes17, .bytes18, .bytes19, .bytes20, .bytes21, .bytes22, .bytes23, .bytes24, .bytes25, .bytes26, .bytes27, .bytes28, .bytes29, .bytes30, .bytes31, .bytes32:
+                        let result = value as? String ?? ""
```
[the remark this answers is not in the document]
James-Sangalli: TODO make sure about this one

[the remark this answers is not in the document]
James-Sangalli: Can confirm they use Data

James-Sangalli:
@hboon ready for another review, although all the builds have suddenly hit a DB exception on my computer, even if I go back to master... Does it work on your side?

AlphaWallet/TokenScriptClient/Models/AssetAttributeSyntax.swift
```diff
@@ -74,6 +74,8 @@ enum AssetAttributeSyntax: String {
             return .string(address.eip55String)
         case .string:
             return value
+        case .bytes:
+            return value
```
hboon: The return value for `.bytes` has to be converted to string first and then wrapped as: `.string(stringConvertedFromData)`

AlphaWallet/TokenScriptClient/Models/AssetAttributeMapping.swift
```diff
@@ -42,6 +42,8 @@ struct AssetAttributeMapping {
             return address.eip55String
         case .string(let string):
             return string
+        case .bytes(let data):
+            return data.hexString
```
hboon: Probably want to use `hexEncoded` instead, which adds a "0x" prefix.

AlphaWallet/TokenScriptClient/Models/AssetAttributeToJavaScriptConvertor.swift
```diff
@@ -10,6 +10,8 @@ struct AssetAttributeToJavaScriptConvertor {
             return "\"\(address.eip55String)\""
         case .string(let string):
             return "\"\(string)\""
+        case .bytes(let bytes):
+            return "\"\(bytes.hexString)\""
```
hboon: Probably want to use hexEncoded instead, which adds a "0x" prefix.

AlphaWallet/TokenScriptClient/Models/FunctionOrigin.swift
```diff
@@ -80,11 +80,13 @@ struct FunctionOrigin {
     private let originContractOrRecipientAddress: AlphaWallet.Address
     private let attributeId: AttributeId
     private let functionType: FunctionType
+    private let bitmask: BigUInt
```
hboon: I think we should either:

A) make the `bitmask` property optional, and we try to read it before applying it, either fallback to `0xF...F`, or even better skip the bitmask operation entirely
B) In the 3 `.init()` fallback to `0xF..F` (then we can't skip the bitmask operation like in A above)

The 2 bitmask value in `XMLHandler.createFunctionOriginFrom()` shouldn't be `BigUInt(0)`, they should be nil then A or B above take care of it.

AlphaWallet/TokenScriptClient/Models/UserEntryOrigin.swift
```diff
@@ -10,24 +10,40 @@ struct UserEntryOrigin {
...
         case .uint:
-            return BigUInt(userEntry).flatMap { .uint($0) }
-        case .utf8:
+            guard let userEntryNumber = BigUInt(userEntry, radix: 16) else { return BigUInt(userEntry).flatMap { .uint($0) } }
+            let number: BigUInt = (bitmask & userEntryNumber) >> bitShift
+            return .uint(number)
+        case .utf8, .string:
             return String(data: Data(bytes: userEntry.hexa2Bytes), encoding: .utf8).flatMap { .string($0) }
+        case .bytes:
+            guard let userEntryNumber = BigUInt(userEntry, radix: 16) else { return .bytes(Data(bytes: userEntry.hexa2Bytes)) }
```
hboon: Can you see if `userEntry.hexa2Bytes` handles "0xF" correctly? Maybe we need to strip off the "0x" if it's there (there's a `drop0x` for that).

AlphaWallet/TokenScriptClient/Models/AssetAttributeValueUsableAsFunctionArguments.swift
```diff
@@ -126,6 +141,8 @@ enum AssetAttributeValueUsableAsFunctionArguments {
             return address.eip55String as AnyObject
         case .string(let string):
             return string as AnyObject
+        case .bytes(let bytes):
+            return bytes as AnyObject
```
hboon: `return bytes as AnyObject` has to be converted to a string with like `0xFF`

James-Sangalli:
@hboon thanks for that, rebased and ready to go.

hboon:
Good work!

PR #1475, opened 2019-09-26
Fix: when displaying token in import MagicLink UI, the section attribute is displayed wrongly

### LINKED ISSUE #1474: Display issue with attribute in import MagicLink screen

hboon:
Fixes #1474 

(Would not have happened if this had a snapshot test)

Before PR|After PR
-|-
 |


hboon:
@James-Sangalli up to you if we need to resubmit

PR #1479, opened 2019-09-26
Adding Fuse Studio DApp

bejavu:


James-Sangalli:
Hi @bejavu, unfortunately the app store forbade us from putting references to dapps, so this is no longer used anymore. This may change in future but in the meantime this will not be visible

James-Sangalli:
@bejavu please also revert the putting into master commit as this is not necessary

PR #1495, opened 2019-10-11
sets a default bitmask for 2019/10 TokenScripts

### LINKED ISSUE #1493: TokenScript properties and updates to webview don't load in 2019/10

James-Sangalli:
Closes #1493

### DIFF
--- a/AlphaWallet/TokenScriptClient/Models/FunctionOrigin.swift
+++ b/AlphaWallet/TokenScriptClient/Models/FunctionOrigin.swift
@@ -147,8 +147,11 @@ struct FunctionOrigin {
         let resultSubscribable = Subscribable<AssetInternalValue>(nil)
         subscribable.subscribe { value in
             guard let value = value else { return }
-            guard let bitmask = self.bitmask else { return }
-            resultSubscribable.value = self.castReturnValue(value: value, bitmask: bitmask)
+            if let bitmask = self.bitmask {
+                resultSubscribable.value = self.castReturnValue(value: value, bitmask: bitmask)
+            } else {
+                resultSubscribable.value = value
+            }
         }
         return .subscribable(resultSubscribable)
     }
--- a/AlphaWallet/TokenScriptClient/Models/Origin.swift
+++ b/AlphaWallet/TokenScriptClient/Models/Origin.swift
@@ -80,22 +80,21 @@ enum Origin {
     }
 
     init?(forTokenIdElement tokenIdElement: XMLElement, xmlContext: XmlContext) {
-        guard let bitmask = XMLHandler.getBitMask(fromTokenIdElement: tokenIdElement) else { return nil }
+        let bitmask = XMLHandler.getBitMask(fromTokenIdElement: tokenIdElement) ?? TokenScript.defaultBitmask
         guard let asType = tokenIdElement["as"].flatMap({ OriginAsType(rawValue: $0) }) else { return nil }
-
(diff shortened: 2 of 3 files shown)


AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -810,9 +810,12 @@ extension XMLHandler {
         }
     }
 
-    static func getBitMask(fromTokenIdElement tokenIdElement: XMLElement) -> BigUInt? {
-        guard let bitmask = tokenIdElement["bitmask"] else { return nil }
-        return BigUInt(bitmask, radix: 16)
+    static func getBitMask(fromTokenIdElement tokenIdElement: XMLElement) -> BigUInt {
```
hboon: Nice!

How about keeping `getBitMask()` as returning `BigUInt?` (i.e. optional), as well as storing the bitmask as an optional in the various origins types (e.g. `FunctionOrigin`), and then when applying, we can check if the bitmask is nil, just return `value` instead of "`value AND bitmask` + shift" (actually I thought that's what we did). This is better because:

1. we potentially skip an `AND` operation for every attribute with the initial load and every refresh.
2. we don't have to hardcode a constant (bitmask)
3. Being nil in the origin types (`FunctionOrigin`) etc also matches the schema where bitmask is optional, although having a default value helps with that too, but refer to [2].

If we have to add a constant for the bitmask, it'll fit better in `TokenScript.swift` because it's better to think of the TokenScript client as a separate framework/pod, which we should probably work on at some point. We might have a single cross-platform code base for the meat of TokenScript, but we'll still have quite a bit of platform-specific TokenScript client code. This is minor and easy to discover and refactor later of course.

James-Sangalli: @hboon see latest commit. A default must be applied in the origins otherwise it will return nil and never resolve the value. I put this as a constant in TokenScript.swift and only use it in Origins init

AlphaWallet/TokenScriptClient/Models/FunctionOrigin.swift
```diff
@@ -147,8 +147,11 @@ struct FunctionOrigin {
         let resultSubscribable = Subscribable<AssetInternalValue>(nil)
         subscribable.subscribe { value in
             guard let value = value else { return }
-            guard let bitmask = self.bitmask else { return }
-            resultSubscribable.value = self.castReturnValue(value: value, bitmask: bitmask)
+            if let bitmask = self.bitmask {
+                resultSubscribable.value = self.castReturnValue(value: value, bitmask: bitmask)
```
James-Sangalli: @hboon removes unnecessary shifting and masking if there is no bitmask

AlphaWallet/TokenScriptClient/Models/Origin.swift
```diff
@@ -80,22 +80,21 @@ enum Origin {
     }
 
     init?(forTokenIdElement tokenIdElement: XMLElement, xmlContext: XmlContext) {
-        guard let bitmask = XMLHandler.getBitMask(fromTokenIdElement: tokenIdElement) else { return nil }
+        let bitmask = XMLHandler.getBitMask(fromTokenIdElement: tokenIdElement) ?? TokenScript.defaultBitmask
```
James-Sangalli: Needed to prevent a nil return and thus no value being resolved

PR #1498, opened 2019-10-14
(trying to work with Travis:) Fix: crash when etherscan API returns malformed ERC20 transfer transaction data, with missing `to

hboon:


hboon:
@James-Sangalli I have no idea!

PR #1505, opened 2019-10-23
checks if view is defined and if not, sets the default opensea view

### LINKED ISSUE #1501: Keep default layout for 721 tokens when you have a tokenscript for it
When creating the ENS TokenScript I noticed that the OpenSea layout disappears when I try to add extra functions to the token. We should keep this default layout IF the TokenScript developer leaves a blank view for the card.

James-Sangalli:
closes #1501


hboon:
Strange. I was fixing this and thought it had to be more complicated. Let me check this out.

PR #1511, opened 2019-10-24
Use OpenSea 721 layout even if there is a TokenScript file, if it does not contain an appropriate view

### LINKED ISSUE #1501: Keep default layout for 721 tokens when you have a tokenscript for it

hboon:
Supersedes #1505. Fixed #1501 

Sorry, was working on this halfway and was distracted. #1505 is correct, but we have to handle both `view` and `view-iconified`. Also took the chance to remove a few additional classes that were originally used for displaying OpenSea 721 tokens. They aren't necessary since this PR makes them (721 display and non-721 TokenScript display) swappable.


hboon:
@James-Sangalli and of course Travis passes once you re-run the test suite..

PR #1512, opened 2019-10-25
with latest cocopods

James-Sangalli:


James-Sangalli:
@hboon travis needs it to, adding now

PR #1515, opened 2019-10-26
allows setting of user-entry from token instance

### LINKED ISSUE #1513: Should be able to fill attribute-types with web3.tokens.data.currentInstance vars

James-Sangalli:
Closes #1513

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/TokenInstanceActionViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/TokenInstanceActionViewController.swift
@@ -150,6 +150,7 @@ class TokenInstanceActionViewController: UIViewController, TokenVerifiableStatus
         let userEntryIds = action.attributes.values.compactMap { $0.userEntryId }
         let fetchUserEntries = userEntryIds
                 .map { "document.getElementById(\"\($0)\").value" }
+                .map { "web3.tokens.data.currentInstance.\"\($0)\"" }
                 .compactMap { tokenScriptRendererView.inject(javaScript: $0) }
         let xmlHandler = XMLHandler(contract: contract, assetDefinitionStore: assetDefinitionStore)
         let tokenLevelAttributeValues = xmlHandler.resolveAttributesBypassingCache(withTokenId: tokenId, server: server, account: session.account)


AlphaWallet/Tokens/ViewControllers/TokenInstanceActionViewController.swift
```diff
@@ -150,6 +150,7 @@ class TokenInstanceActionViewController: UIViewController, TokenVerifiableStatus
         let userEntryIds = action.attributes.values.compactMap { $0.userEntryId }
         let fetchUserEntries = userEntryIds
                 .map { "document.getElementById(\"\($0)\").value" }
+                .map { "web3.tokens.data.currentInstance.\"\($0)\"" }
```
hboon: @James-Sangalli Implementation-wise, this wouldn't work because it'll effectively generate JavaScript like this:

```
web3.tokens.data.currentInstance.document.getElementById("someId").value
```

These 2 maps have to be combined into a single `.map()` so that for each element in `userEntryIds`, it tries `getElementById()` then falls back to `web3.tokens.data.currentInstance` if the former fails. The latter can fail too and might need to be handled too.

	.map { "document.getElementById(\"\($0)\").value" }
	.map { "web3.tokens.data.currentInstance.\"\($0)\"" }

Also, accessing the property in currentInstance should be changed from `web3.tokens.data.currentInstance.foo` to `web3.tokens.data.currentInstance["foo"]`. The latter will work when the `user-entry` has an identifier that is not a valid JavaScript variable name. E.g. "my-id".

---

But on a higher level, I'm not sure if this is the right thing to add support for. The `user-entry` origin is meant for:

1. user input
2. and only fetching unmodified user input

Because we are using HTML and JavaScript as the frontend, it means that we can programmatically generate a hidden `input` as a workaround for (1) so it might not come from the user anyway, and also attach JavaScript to modify the input so it breaks (2) too. Would you check with @colourful-land if he's OK with this?

Are we generating state programatically (sometimes from processing user input, sometimes from web API calls) in the TokenScript file that needs to be passed to the TokenScript client? If it's processing user input, what are we doing with it?

Otherwise the implementation (with the changes mentioned above) seems ok.

James-Sangalli: @hboon thanks for that, noted.

I do use hidden inputs right now but there are a few issues without using the token instance.

For example: opensea provides the ENS name from it's call, there is currently no way to feed this value from the instance straight into the attribute-type except for defining it as user-input and reading it from there (this is NOT entered by the user at all & cannot be retrieved from the smart contract directly except via event logging). This has created the problem that we cannot load the ensName straight away into other attributes which depend on it to get things like the price to renew the domain etc. We will need to workout a solution in both iOS and android which can handle reading this straight away so that we can get the values we need onto the interface and into the tokenscript. In android it currently reads the values on load and applies things like the bitmask to it but this causes an issue in the javascript code because it is represented as a byte buffer and throws an error because the JS code is expecting a string and we need to show a string value to the user not a byte buffer (of course we can convert this buffer into a string but that would need to be coded into the JS which is not nice, better to do it from the client) 

Often we also need to do an operation on user-input for it to be valid, such as doing a namehash operation on the subdomain that a user enters, I work around this by either modifying the existing element before it gets confirmed by the users click or creating a hidden element, it would be more elegant to simply get this from the props.

James-Sangalli: @hboon flag this PR, it's only sensible that we use event support to get this rather than creating a hacky fix like this

Issue #1438, opened 2019-09-11
FIFA normal transfer is not working


zhangzhongnan928:
error msg as below

![IMG_6653](https://user-images.githubusercontent.com/33795543/64696559-05695f80-d4e2-11e9-8be7-f74ff4f151cc.PNG)

hboon:
@James-Sangalli can you help verify if it's still happening?

James-Sangalli:
@zhangzhongnan928 @hboon not an issue for me in the app store version https://etherscan.io/tx/0x555dca89db4146743253305bf1923ddc908b4b341e3807b96e679039f59de3e9

hboon:
@James-Sangalli Thanks!

PR #1519, opened 2019-10-28
fix encoding of params for NFT encodings

### LINKED ISSUE #1518: ENS 721 token transfer function fail， build2.12 300
D3EDFF6E-79DD-4AFA-8C1B-CFE0A3568497

James-Sangalli:
Fixes #1518 & https://github.com/AlphaWallet/alpha-wallet-ios/issues/1438

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -207,7 +207,7 @@ class TransactionConfigurator {
                     parameters = [TrustKeystore.Address(address: transaction.to!), BigUInt(transaction.tokenId!)!]
                 } else {
                     function = Function(name: "safeTransferFrom", parameters: [.address, .address, .uint(bits: 256)])
-                    parameters = [TrustKeystore.Address(address: self.account.address), transaction.to!, BigUInt(transaction.tokenId!)!]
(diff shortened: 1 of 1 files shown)


hboon:
@James-Sangalli there you go. I only went as far as stepping past the encoding. See if it makes sense and please test it :)

James-Sangalli:
@hboon the encoding works but the contract address check is wrong (checking the recipient but not the contract address) this is bizarre but I will solve it here too.

James-Sangalli:
@hboon fixed, see commit

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -198,11 +198,11 @@ class TransactionConfigurator {
                 completion(.failure(AnyError(Web3Error(description: "malformed tx"))))
             }
 
-        case .ERC721Token:
+        case .ERC721Token(let token):
             do {
                 let function: Function
                 let parameters: [Any]
-                if transaction.to!.isLegacy721Contract {
```
James-Sangalli: @hboon this is the recipient address which is wrong

hboon: So bizzarre.

PR #1524, opened 2019-10-29
Remove references to cryptokitties

hboon:
This PR makes 2 changes to comply with Apple app review guidelines:

* Replace the full dapp list (which we were already not using in the app to comply with Apple guidelines previously) from the project, with an empty version, so that the cryptokitties.co URL (as well as URLs for other dapps/tokens) is no longer included in the binary
* Rename several var/folder/type name that mentions "cryptokitt" just so we don't get raise a false alarm with app review think that we had added special support for CryptoKitties.


hboon:
@James-Sangalli Would you help run this on the binary after building and pre-submission? There should be zero matches:

```
strings AlphaWallet.app/AlphaWallet | grep -i cryptokitt
```

James-Sangalli:
@hboon no matches, thanks!

PR #1525, opened 2019-10-29
clean up some warnings

James-Sangalli:


AlphaWallet/UI/WhereIsWalletAddressFoundOverlayView.swift
```diff
@@ -79,11 +79,11 @@ extension WhereIsWalletAddressFoundOverlayView: DialogDelegate {
     }
 }
 
-fileprivate protocol DialogDelegate: class {
+private protocol DialogDelegate: class {
```
hboon: Not too important, but is Xcode complaining about use of `fileprivate`? Because the few usage here are legit — the types are only used within the file.

James-Sangalli: Yup, private is accessible from extensions in swift 4, https://stackoverflow.com/questions/39027250/what-is-a-good-example-to-differentiate-between-fileprivate-and-private-in-swift

hboon: Yes, but why change from fileprivate to private? Because they are only accessed within the same file.

James-Sangalli: fileprivate is too liberal, can be restricted further as private?

hboon: Haha. Actually, maybe there is no difference between fileprivate and private if it's for the outermost type (in this case the DialogDelegate class), since they all fall under the same/one project module. Anyway, the PR is good.

PR #1539, opened 2019-11-06
add support for spawnable paymaster queries

James-Sangalli:
Needed for spawnable UEFA tickets which will depend on the paymaster only

### DIFF
--- a/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
+++ b/AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
@@ -88,7 +88,7 @@ class UniversalLinkCoordinator: Coordinator {
     private func createHTTPParametersForCurrencyLinksToPaymentServer(
             signedOrder: SignedOrder,
             recipient: AlphaWallet.Address
-    ) -> Parameters {
+    ) -> (Parameters, String) {
         let signature = signedOrder.signature.drop0x
         let parameters: Parameters = [
             "prefix": Constants.xdaiDropPrefix,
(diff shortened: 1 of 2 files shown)


AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -127,8 +130,16 @@ class UniversalLinkCoordinator: Coordinator {
         if isForTransfer {
             parameters.removeValue(forKey: "price")
         }
+
+        if signedOrder.order.spawnable {
```
hboon: It looks like we will no longer hit `Constants.currencyDropServer` when `signedOrder.order.nativeCurrencyDrop` == true. Is this what we want?

If not, might be better to restore the assignment for `query` back into `getParametersAndQuery()` and add a check for `signedOrder.order.spawnable` there and assign accordingly.

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -431,6 +437,11 @@ class UniversalLinkCoordinator: Coordinator {
         return indices.map(String.init).joined(separator: ",")
     }
 
+    private func stringEncodeTokenIds(_ tokenIds: [BigUInt]?) -> String? {
+        guard let tokens = tokenIds else { return nil }
+        return tokens.map({ $0.serialize().hexString }).joined()
```
hboon:
```
return tokens.map({ $0.serialize().hexString }).joined(separator: ",")
```

James-Sangalli: Not necessary here because the paymaster knows each token is 32 bytes

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
James-Sangalli: @hboon thats handled by createHTTPParametersForCurrencyLinksToPaymentServer in the switch statement?

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
James-Sangalli: Actually, should split by commas, shorter url and more consistent

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
hboon: Ah, right.

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
hboon: Yes, but also because it's correct? Because we want something like "0x01,0x02" rather than "0x010x02".

James-Sangalli: Was originally thinking of a full padded 32 bytes per token so you could separate but that's long

PR #1553, opened 2019-11-13
Add support for ERC 20, and improve existing support for native crypto, for EIP 681 payment links

### LINKED ISSUE #1526: Support ERC20 sends with EIP681
Native crypto transfers is already supported with:

ethereum:0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359?value=2.014e18

Support this:

```
ethereum:0x744d70fdbe2ba4cf95131626614a1763df805b9e/transfer?address=0x3d597789ea16054a084ac84ce87f50df9198f415&uint256=314e17
```

hboon:
Closes #1526.

* Stricter parsing
* Set up foundation so that we can add support for other types of `ethereum:` links, other than EIP 681
* When an ERC 20 token is not found in the local database, it's fetched first
* When user is at the send native crypto/ERC20 screen and scans a payment link for another token, it will update accordingly. It will only switch when the server is not specified in the link or is the same as the current server

Limitations/TODO:

* ~~Assumes decimals=18 for ERC 20 tokens. It's still safe for the user since the amount is displayed~~
* Balance displayed if the ERC20 token is not in the database is 0. #1554


James-Sangalli:
@hboon compound etc uses decimals 8, will the number displayed be massive?

hboon:
> @hboon compound etc uses decimals 8, will the number displayed be massive?

We'll have to fix it to read decimals then. Hold on first.

hboon:
> @hboon compound etc uses decimals 8, will the number displayed be massive?

Fixed. Good for review again. Thanks

PR #1555, opened 2019-11-14
refactor magic link handlers

James-Sangalli:


AlphaWallet/Market/OrderHandler.swift
```diff
@@ -55,20 +55,18 @@ public class OrderHandler {
     func signOrders(orders: [Order], account: EthereumAccount) throws -> [SignedOrder] {
         var signedOrders = [SignedOrder]()
         var messages = [Data]()
-
-        for i in 0...orders.count - 1 {
+        let _ = orders.map({
```
hboon: Better to use for-each if you don't expect a mapped output:

```
for each in orders {
```

AlphaWallet/Market/UniversalLinkHandler.swift
```diff
@@ -206,72 +206,53 @@ public class UniversalLinkHandler {
...
-        //contract address
-        messageWithSzabo.append(contentsOf: message[64...83])
+        messageWithSzabo.append(contentsOf: signedOrder.order.contractAddress.data.bytes)
         messageWithSzabo.append(contentsOf: indices)
-        messageWithSzabo.insert(LinkFormat.normal.rawValue, at: 0)
         return Data(bytes: messageWithSzabo).hex()
     }
-    
+
     private func formatTo4Bytes(_ array: [UInt8]) -> [UInt8] {
```
hboon: Maybe call it `padTo4Bytes()` for consistency?

PR #1561, opened 2019-11-18
add ERC721 ticket support 

James-Sangalli:


James-Sangalli:
@hboon always returns 1 balance even if there is no balance

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -495,9 +495,9 @@ private class PrivateXMLHandler {
             actions = [.nonFungibleTransfer]
         case .erc875:
             if contractAddress.isFifaTicketcontract {
-                actions = [.erc875Redeem, .erc875Sell, .nonFungibleTransfer]
+                actions = [.nftRedeem, .nftSell, .nonFungibleTransfer]
```
hboon: But 721 is an NFT too.

hboon: Ah, you are right because they are applicable for ERC721 tickets too.

hboon:
@James-Sangalli I missed this earlier: do you want to add an .erc721ForTicketTokens case to TokenInterfaceType?

hboon:
> @James-Sangalli I missed this earlier: do you want to add an .erc721ForTicketTokens case to TokenInterfaceType?

@James-Sangalli Sorry! My bad, James. It's parsed and extracted from the TokenScript file. So we don't need (and can't have) `erc721ForTicketTokens`/`erc721Ticket`. So your original way of checking the contract is correct.

hboon:
All good otherwise. Merge when you are done with removing the directory :P

PR #1567, opened 2019-11-20
add compatibility for erc721 ticket magic links

James-Sangalli:


AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -390,6 +390,7 @@ private class PrivateXMLHandler {
         }
         return Token(
                 id: tokenId,
+                tokenType: TokenType.erc875,
```
James-Sangalli: @hboon slightly put off by this

hboon: @James-Sangalli Good call.

How about adding a `tokenType: TokenType` arg to this function:

`func getToken(name: String, symbol: String, fromTokenId tokenId: TokenId, index: UInt16, inWallet account: Wallet, server: RPCServer, callForAssetAttributeCoordinator: CallForAssetAttributeCoordinator) -> Token {`

I noticed that callers either have a `TokenObject` available (which has a `tokenType` property) or should at least have the contract and TokensDataStore (or ServerDictionary<TokenDataStore> which you can use an RPCServer as the key to get the correct TokensDataStore) instance, and then the `TokenObject` to read its `tokenType` property from.

We might have to re-examine `Token` later to see if it's always used correctly or whether it's necessary at all.

James-Sangalli:
@hboon ready to go

AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -489,13 +490,14 @@ class UniversalLinkCoordinator: Coordinator {
 
     private func makeTokenHolderImpl(name: String, symbol: String, bytes32Tokens: [String], contractAddress: AlphaWallet.Address) {
         //TODO pass in the wallet instead
+        guard let tokenHolderStrong = tokenHolder else { return }
```
hboon: This wouldn't work, because `self.tokenHolder` is nil. (An unfortunate part of using `TokenHolder?` as a type here). We'll need to use something like:

```
let tokensDatastore = tokensDatastores[server]
let token = tokensDatastore.token(forContract: someContract)
//token.tokenType
```

James-Sangalli: Haha that's what I thought my happen since it is assigned below, one second.

James-Sangalli:
@hboon sussed

PR #1569, opened 2019-11-20
checks if erc165 getBalances(address) hash is supported

James-Sangalli:


AlphaWallet/RPC/Commands/web3swift-pod/GetInterfaceSupported165Encode.swift
```diff
@@ -0,0 +1,7 @@
+// Copyright © 2019 Stormbird PTE. LTD.
+import Foundation
+
+struct GetInterfaceSupported165Encode {
```
hboon: `GetInterfaceSupported165Encode` and `GetIsERC721` looks identical. Maybe rename the latter to the former? The new coordinator still looks necessary though.

AlphaWallet/Settings/Types/Constants.swift
```diff
@@ -57,7 +57,10 @@ public struct Constants {
     static let ticketContractAddressRopsten = AlphaWallet.Address(string: "0xD8e5F58DE3933E1E35f9c65eb72cb188674624F3")!
     
     // UEFA hardcoded addresses
-    static let uefaRopsten = AlphaWallet.Address(string: "0x439bf1B20841Fa5498C3471C202F438147b7e447")!
+    static let uefaRopsten = AlphaWallet.Address(string: "0xc2f901070AdcfA07Bc5c46cc2390063248eb9E81")!
+
+    //UEFA 721 balances function hash
+    static let balances165Hash721Ticket = "0xc84aae17"
```
hboon: How about removing this and just relying on the ERC165 check? It's one thing less to maintain though performance will be slightly better (but only once while the contract is added) and only for this contract.

hboon: It'll probably affect a few other places where the constant is use, which is good. Those usages will cause problem in the future :)

PR #1576, opened 2019-11-23
Fix: TokenScript client sometimes stop displaying attributes in web views. Requires restart to fix

hboon:
Symptoms are identical to what Victor reported reported a month or two ago.

This PR disables reading from the cache for token attributes.

This is good for function origin-based attributes since we don't have a way to specify TTL for them yet. But this will worse performance for scrolling in the wallet tab. It should not affect users wouldn't many token IDs (not tokens) with TokenScript tokenID attributes, but we should bring back caching, maybe this time only for token ID-based attributes.


hboon:
> Hmm this again... Thanks

@James-Sangalli Related, but not the same issue. Previous issue was in the action view with the value being stuck (because it was cached). This time something is wrong with the cache itself.

PR #1578, opened 2019-11-23
filters zero balance tokens from balances array

James-Sangalli:
If you have a zeroed out array value it will still add to your balance and give you a falsely high number, this PR prevents this for erc721 tickets


James-Sangalli:
> More idiomatic to use ‘isZero‘

Doesn't exist in BigUInt

hboon:
@James-Sangalli oh, it's not public. Oh well.

PR #1583, opened 2019-11-27
Separate out Infura key into a separate file so it is easy to specify a different key (file) locally when building

hboon:
As a separate step, we can replace this new `Constants+Credentials.swift` file with a local version that is ignored by git.


hboon:
@James-Sangalli you can ask git to ignore the file for changes with:

```
git update-index --assume-unchanged AlphaWallet/Settings/Types/Constants+Credentials.swift
```

Then you can just change that file locally, and forget about it. This has to be done for every repo. i.e. if you clone it, you'll have to do this again for the new repo.

PR #1602, opened 2019-12-03
######## 50 characters ############################

prediator:
Spacers issue fixed.
########72 characters##################################################
Problem
# beforeTokenScriptFileStatus Not found
Solution
# Removed beforeTokenScriptFileStatus and used UIView as spacers


hboon:
@prediator I might have missed some context. What does this PR fix? Did you get a build error before applying this PR?

PR #1614, opened 2019-12-11
Always support decimal comma as decimal separator in addition to what is specified by locale when entering amounts

### LINKED ISSUE #1599: Decimal point doesn't work in spanish
Maybe because it uses a comma instead

hboon:
Fixes #1599 

There are a few ways to change the locale or preferred language for the app:

1. Change the locale in Settings.app
2. Change the preferred language in Settings.app
3. Change the preferred language in Settings.app for AlphaWallet only (iOS 13 and newer)
4. Change the language in AlphaWallet Settings tab

When the user sets the in-app locale (method 4) to one that uses a decimal comma (E.g. Spanish), the `.decimalPad` keyboard still (only) shows decimal point. The `.decimalPad` keyboard only switches to the decimal comma correctly with methods 1, 2 and 3.

This PR changes it so that we always consider the decimal point as the one defined by the locale.

### DIFF
--- a/AlphaWallet/Foundation/EtherNumberFormatter.swift
+++ b/AlphaWallet/Foundation/EtherNumberFormatter.swift
@@ -1,9 +1,11 @@
 // Copyright SIX DAY LLC. All rights reserved.
-
 import BigInt
 import Foundation
 
 final class EtherNumberFormatter {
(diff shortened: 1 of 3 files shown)


hboon:
@James-Sangalli have to be a bit more careful when verifying this to be sure numbers are parsed correctly when entering eth or dollar amounts.

PR #1618, opened 2019-12-16
Restore dapp browsers bookmark functionality

### LINKED ISSUE #1463: Restore bookmark functionality to Dapp Browser
Was called "My Dapps". Watch out to not use the term "app", and just use "bookmark".

### LINKED ISSUE #1522: Can we have bookmark function back for DApp Browser
we can leave the home page empty and don’t provide the suggested websites list. But without bookmark, it is so difficult to use the DApp browser.

hboon:
Closes #1463, closes #1522

This PR restores bookmark functionality. Rename "dapp" references to "bookmark".


hboon:
Don’t think they’ll like it. But I was thinking we can produce a format. Maybe use RSS or OPML. And let the user add a “feed” by pasting in a URL that we get a list from? Then auto completion can come back :) talk about the URL outside of the app, eg Twitter and Blog. what do you think?

PR #1631, opened 2019-12-23
Reduce compiler and SwiftLint warnings

hboon:


hboon:
Expecting to see Travis complain for the main branch because I accidentally forced pushed to master and undo it.

PR #1641, opened 2020-01-05
remove redundant market queue class

James-Sangalli:


colourful-land:
> We can probably remove `MarketQueueHandlerTests` too.

Agree

PR #1671, opened 2020-01-21
gets the gas price from the node rather than guessing

### LINKED ISSUE #1667: Use gas prices from node in TokenScript

James-Sangalli:
Closes #1667

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -6,6 +6,7 @@ import Result
 import TrustKeystore
 import JSONRPCKit
 import APIKit
+import PromiseKit
 
 public struct PreviewTransaction {
     let value: BigInt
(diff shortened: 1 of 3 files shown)


AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -112,6 +112,23 @@ class TransactionConfigurator {
         }
     }
 
+    // Generic function to derive the typical acceptable gas price on each network
+    static public func estimateGasPrice(server: RPCServer, completion: @escaping (BigInt) -> Void) {
```
James-Sangalli: @hboon I know you don't like statics but this function is super generic and you shouldn't need to initialise a class just to use it

hboon: Nah. It's fine.

hboon:
@James-Sangalli related to this, should there be a safe guard around the gas price? Since we don't present the transaction confirmation dialog for TokenScript transactions like we do for other transactions, users might be unknowingly hit with an exorbitant gas price. A hardcoded gas price ceiling in the code might fix this.

Can you create an issue (or work it into this PR) if you think this is necessary?

James-Sangalli:
> Does xDai chain need to be handled specially (i.e. `TransactionConfigurator.estimateGasPrice(server:)` hardcoded to return `GasPriceConfiguration.xDaiGasPrice)` or is that handled already by this [1] ?
> 
> Slightly better if `TransactionConfigurator.estimateGasPrice(server:)` returns a `Promise` rather than accept a block. But no biggie.
> 
> [1] https://github.com/AlphaWallet/alpha-wallet-ios/blob/master/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift#L29

Haha interestingly, calling the xDAI node to ask for the gas price returns 240 GWEI despite it only needing 1 GWEI.... Thanks for pointing it out 

curl -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_gasPrice","params":[],"id":42}' https://dai.poa.network

{"jsonrpc":"2.0","result":"0x37e11d6000","id":42}

James-Sangalli:
@hboon I changed nothing with this but now this is an error

Screenshot 2020-01-21 15 53 28

PR #1675, opened 2020-01-24
rounds USD figures to 18 decimals

### LINKED ISSUE #1674: "Wrong Input" alert when sending ETH equivalent to 0.02 USD
Happens if the ETH amount converted from USD cannot be represented accurately with 18 decimals.

James-Sangalli:
Fixes #1674


hboon:
@James-Sangalli you can do it without adding the extension:

```
return String((amount / cryptoToDollarRate).rounded(to: 18))
```

Sorry took a while to respond, I thought the user might not get back the ETH they originally entered if they converted back and forth. Eg. 0.02 USD -> X ETH -> 0.0199999 USD, etc. But turns out I was overthinking it. Your solution is much simpler.

James-Sangalli:
@hboon actually your method doesn't quite work as that extension puts it into scientific notation which is then rejected by the input box.

PR #1678, opened 2020-01-28
Add and use API key for Etherscan

### LINKED ISSUE #1677: Use API key for accessing Etherscan
https://medium.com/etherscan-blog/psa-for-developers-implementation-of-api-key-requirements-starting-from-february-15th-2020-b616870f3746

hboon:
Closes #1677 

This PR 

* adds an Etherscan API key
* uses it for access
* use the same approach to manage the key described in #1583 

Since the deadline to use an API key is Feb 15, a build will have to go live before then for transaction history and auto detection of ERC20 tokens from transactions


hboon:
@James-Sangalli dropped you an email with the replacement file that should not be checked in.

PR #1686, opened 2020-02-03
Add named colors to asset catalog as part of UI consistency update

hboon:
No functional or visible change as the colors aren't used yet.

Also paving the way for Dark Mode in the future.


hboon:
@James-Sangalli if only Apple approves submissions 1/10 your speed.

PR #1690, opened 2020-02-04
remove blockchain name text

### LINKED ISSUE #1462: Keep one network label in each token card
Re-attempt to address this issue as boat #1414 has fallen (fallen un-executable = fallen dead).

There are two network labels only one needs to remain. Kill the 2nd (The one below the title). Keep the 1st (the one right to the title)

63941384-da791780-caae-11e9-9ee4-d29926d65ed6

James-Sangalli:
Closes #1462

@hboon before merging could you please quickly set the spacing back to how it was before? It has made the token cards smaller

Also, we no longer have a target to show the contract page anymore, I believe this is possible to do on a label indirectly but I am not sure it is valuable to the user as it is not obvious at all that clicking the label does anything and the tx page allows you to access etherscan in a more obvious fashion

### DIFF
--- a/AlphaWallet/Settings/Types/RPCServers.swift
+++ b/AlphaWallet/Settings/Types/RPCServers.swift
@@ -305,19 +305,6 @@ enum RPCServer: Hashable, CaseIterable {
         }
     }
 
-    var blockChainName: String {
-        switch self {
-        case .xDai:
-            return R.string.localizable.blockchainXDAI()
-        case .artis_sigma1:
-            return R.string.localizable.blockchainARTISSigma1()
-        case .artis_tau1:
-            return R.string.localizable.blockchainARTISTau1()
-        case .main, .rinkeby, .ropsten, .custom, .callisto, .classic, .kovan, .sokol, .poa, .goerli:
-            return R.string.localizable.blockchainEthereum()
-        }
-    }
-
     var blockChainNameColor: UIColor {
         switch self {
         case .main: return .init(red: 41, green: 134, blue: 175)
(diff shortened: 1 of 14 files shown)


hboon:
> @hboon before merging could you please quickly set the spacing back to how it was before? It has made the token cards smaller

@James-Sangalli Can you help me screenshot the before and after so I can see which screen and spacing it is? I took a look, if 1 line is removed, of course it's going to be shorter :P

> Also, we no longer have a target to show the contract page anymore, I believe this is possible to do on a label indirectly but I am not sure it is valuable to the user as it is not obvious at all that clicking the label does anything and the tx page allows you to access etherscan in a more obvious fashion

I think it's still useful to have:

* It's much faster to reach the contract page that way compared to going through a transaction
* We don't always see a transaction (even less common when we have event support, but say if the `balance()` smart contract functions returns a hardcoded non-zero balance, there may not be a transaction)
* We have the code already
* Power users find it useful. Victor managed to discover it himself and likes using it
* It doesn't negatively affect users who doesn't know about it

We can either change the remaining blockchain name pill label to a button (better) or use a UITapGestureRecognizer (not so good).

James-Sangalli:
Before:
Screen Shot 2020-02-04 at 4 36 48 pm


After
Screen Shot 2020-02-04 at 4 35 38 pm

hboon:
@James-Sangalli how would you prefer that screen to be instead? I noticed it's shorter because the line with the label has been removed.

James-Sangalli:
@hboon same size just without the line in it

hboon:
@James-Sangalli Ok. Just the Wallet tab right? I'll tweak it after you have pushed again.

hboon:
@James-Sangalli BTW, the issuer is gone in the Wallet tab?

Oh, if it's too much work, I can implement https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5dfaf2ce114ed41d029b3e2b soon. I think I have most of the common things for Fiji set up and implemented already anyway. We can ignore the icons first, if we are undecided on it.

James-Sangalli:
@hboon no the issuer is still there but only if it has one

James-Sangalli:
Closing due to Fiji design, maybe good for reference later.

PR #1691, opened 2020-02-04
prevents excessive gas limtis set by dapp browser

### LINKED ISSUE #1563: Gas limit exceeds block gas limit on The new MakerDAO vault
444FC058-3329-4C3E-80E2-749DFDBC7856
9369F68A-D9A6-4CFE-AA8F-40EF1A0F85CF

oasis.app

The new MakerDAO portal returns a super high gas limit when opening vault.

James-Sangalli:
Fixes #1563 

Successfully opened a CDP vault with this tx by using this PR: https://etherscan.io/tx/0xcfb06154d20e17025421ff844e8f438995f6565ca0ddbd0973aada2c8c22a834

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -63,10 +63,9 @@ class TransactionConfigurator {
         self.session = session
         self.account = account
         self.transaction = transaction
-
         self.configuration = TransactionConfiguration(
             gasPrice: min(max(transaction.gasPrice ?? GasPriceConfiguration.defaultPrice, GasPriceConfiguration.minPrice), GasPriceConfiguration.maxPrice),
-            gasLimit: transaction.gasLimit ?? GasLimitConfiguration.maxGasLimit,
+            gasLimit: min(transaction.gasLimit ?? GasLimitConfiguration.maxGasLimit, GasLimitConfiguration.maxGasLimit),
             data: transaction.data ?? Data()
         )
     }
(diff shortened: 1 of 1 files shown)


AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -63,10 +63,15 @@ class TransactionConfigurator {
...
-
+        let gasLimit: BigInt
+        if let txGasLimit = transaction.gasLimit, txGasLimit < GasLimitConfiguration.maxGasLimit {
+            gasLimit = txGasLimit
+        } else {
+            gasLimit = GasLimitConfiguration.maxGasLimit
+        }
         self.configuration = TransactionConfiguration(
             gasPrice: min(max(transaction.gasPrice ?? GasPriceConfiguration.defaultPrice, GasPriceConfiguration.minPrice), GasPriceConfiguration.maxPrice),
-            gasLimit: transaction.gasLimit ?? GasLimitConfiguration.maxGasLimit,
```
hboon: This is optional, but I think you can simplify it with:

```
let gasLimit = [transaction.gasLimit, GasLimitConfiguration.maxGasLimit].min() ?? GasLimitConfiguration.maxGasLimit
```

The annoying `??` is because `min()` returns an optional to cater for when the collection is empty.

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -101,7 +106,11 @@ class TransactionConfigurator {
                     if limit == BigInt(21000) {
                         return limit
                     }
-                    return limit + (limit * 20 / 100)
+                    if limit > GasLimitConfiguration.maxGasLimit {
```
hboon: We need to calculate the +20% before comparing. Because if `limit` is approximately > 80% of `GasLimitConfiguration.maxGasLimit`, it'll run `else` and compute a value bigger than GasLimitConfiguration.maxGasLimit.

hboon: Can use `min()` here too.

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
James-Sangalli: Yup, already aware of that but your way is more elegant

James-Sangalli:
@hboon force pushed

PR #1697, opened 2020-02-07
Fix: ETH amount in Send screen gets cleared after scanning QR code with only address and further editing gets cleared too

### LINKED ISSUE #1681: Text field (amount) periodically cleared whilst typing
- Click on currency/token (Tested with ETH mainnet and rinkeby)
- click Send action
- scan address, address populated
- amount is 0, enter new amount (eg .09)
- after a few seconds, amount goes back to 0. Expect: typed amount to remain.

hboon:
Fixes #1681


jzaki:
@hboon FYI same behaviour for the address field. Not sure if that is fixed with this too.

- scan address, field populated
- delete address field
- modify amount field, and wait a moment
- address re-populates with previously deleted value

PR #1708, opened 2020-02-11
Fix invalid signature for valid tokenscript

hboon:
Calls to signature verification API is sometimes not successful.

Before this PR, such cases are all treated as the TokenScript file having an invalid signature and displayed as such in the UI. This isn't always true.

This PR changes the verification result handling so that we only show "invalid signature" for TokenScript files when a request to the API fails with 400. Otherwise, it'll retry the request again.


James-Sangalli:
@hboon if the signature was valid the first time and the file has not changed, can we simply keep it as verified? Or was that your other PR? Seems redundant to have to give it an invalid status when it was valid in the past

hboon:
>@hboon if the signature was valid the first time and the file has not changed, can we simply keep it as verified? Or was that your other PR? Seems redundant to have to give it an invalid status when it was valid in the past 

@James-Sangalli I was going to work on another PR that provides some caching for that.

But wanted to fix this issue first. It will still bite us with caching enabled (and worse, since restarting wouldn’t trigger the signature verification again).

hboon:
@James-Sangalli ready to test again. It still does what it says in the PR. Retries more kinds of failures but with a backoff on each.

PR #1722, opened 2020-02-17
removes redundant and misleading import image and text

James-Sangalli:


AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -28,7 +28,6 @@ class TokensViewController: UIViewController {
     private let account: Wallet
     lazy private var tableViewFilterView = SegmentedControl(titles: TokensViewModel.segmentedControlTitles)
     lazy private var collectiblesCollectionViewFilterView = SegmentedControl(titles: TokensViewModel.segmentedControlTitles)
-    private var importWalletView: UIView?
     private var importWalletLayer = CAShapeLayer()
```
hboon: We can remove `importWalletLayer` too.

hboon: Sorry, was poorly abstracted :P

PR #1760, opened 2020-03-26
prevent zero input on address scan

### LINKED ISSUE #1751: remove the pre filled 0 in send page
after put in an ens or paste the address into the address field(don't use the paste button)
it will auto put a 0 in the amount field
IMG_F6370B3CCC96-1

James-Sangalli:
Closes #1751


hboon:
@James-Sangalli my bad for not being clear: 

> Does it set the amount to 0 if the user fills in a non-zero amount and then scans a EIP681 link that has value = 0?

PR #1764, opened 2020-04-02
URL in bookmarks and history list not visible because they are have the same color (white) as the background #1726

### LINKED ISSUE #1726: URL in bookmarks and history list not visible because they are have the same color (white) as the background
URL is visible when user taps on the cell:

Bookmarks | History | Output
-|-|-
||

nimitparekh2020:
This issue fixed #1726
- Change URL colour

Bookmarks | History | Output
-|-|-
||

### DIFF
--- a/AlphaWallet/Browser/ViewModel/BrowserHistoryCellViewModel.swift
+++ b/AlphaWallet/Browser/ViewModel/BrowserHistoryCellViewModel.swift
@@ -39,7 +39,7 @@ struct BrowserHistoryCellViewModel {
     }
 
     var urlColor: UIColor? {
-        return Colors.appBackground
+        return Colors.appTint
     }
 
     var imageViewShadowColor: UIColor {
--- a/AlphaWallet/Browser/ViewModel/MyDappCellViewModel.swift
+++ b/AlphaWallet/Browser/ViewModel/MyDappCellViewModel.swift
@@ -39,7 +39,7 @@ struct MyDappCellViewModel {
     }
 
     var domainNameColor: UIColor? {
-        return Colors.appBackground
+        return Colors.appTint
     }
 
     var addRemoveButtonFont: UIFont {


AlphaWallet/Browser/ViewModel/BrowserHistoryCellViewModel.swift
```diff
@@ -39,7 +39,7 @@ struct BrowserHistoryCellViewModel {
     }
 
     var urlColor: UIColor? {
-        return Colors.appBackground
+        return UIColor(red: 0, green: 127, blue: 237)
```
hboon, line 42: Great. Can you:

1. Do it for bookmarks too?
2. `return Colors.appTint` instead, which is the same color, which is basically R.color.azure(). Using `Colors.appTint` where it's the app tint color helps us switch the tint color. This is used in the other repo too and that app has a different tint color.

nimitparekh2020: I make changes into the Bookmark and History section. @hboon kindly review that.

nimitparekh2020:
@hboon I will keep in mind and doing squash commit.

hboon:
@nimitparekh2020 I mean the commit message :)

hboon:
@nimitparekh2020 good, but can you make it one line without the manual word wrap?

PS: This seems trivial, but I hope that ones we get past this and do it consistently, it'll all be good and helpful in the long run.

nimitparekh2020:
> @nimitparekh2020 good, but can you make it one line without the manual word wrap?
> 
> PS: This seems trivial, but I hope that once we get past this and do it consistently, it'll all be good and helpful in the long run.

@hboon I checked Github there is limit of characters for word wrap that's why trim the line.
I tried but it didn't align into single line.

hboon:
@nimitparekh2020 you are right, it's just in the GitHub UI. It's already a 1-liner in the commit itself. I'll merge this soon. Thanks!

PR #1766, opened 2020-04-07
enable scrolling in TS views

### LINKED ISSUE #1765: In TokenScript action page, cannot scroll up or down
IMG_9676DD2FA1D1-1

James-Sangalli:
Fixes #1765

### DIFF
--- a/AlphaWallet/Tokens/Views/TokenInstanceWebView.swift
+++ b/AlphaWallet/Tokens/Views/TokenInstanceWebView.swift
@@ -50,7 +50,7 @@ class TokenInstanceWebView: UIView {
         super.init(frame: .zero)
 
         webView.isUserInteractionEnabled = false
-        webView.scrollView.isScrollEnabled = false
+        webView.scrollView.isScrollEnabled = true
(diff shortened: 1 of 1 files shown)


James-Sangalli:
> @James-Sangalli originally, I thought:
> 
> A) item-view(s) are supposed to be short
> B) we always calculate (we try!) the height of the item-view and always show it without the webview scrolling, so only the entire list scrolls
> 
> So if that is no longer true, can you help to verify that we don't break shorter views and that their webviews don't scroll? i.e. you can't scroll with the rubber-banding effect when the webview is already showing the entire height of the content. (if it's not true, the height calculation might be wrong :( )
> 
> Also verify if scrolling is works properly if there are more than 1 very tall item-view or with many short item-views. Because there are more than 1 active vertical scrollview with this change.

@hboon if you try with the karma token, you will see that the view-iconified/item-view layout is not scrollable. Only inside the action views can you scroll. I have tried it now on numerous tokens and none of the item-view layouts are broken nor are the action views broken.

PR #1769, opened 2020-04-08
Fix: A few database migrations were skipped

hboon:


hboon:
Of course a re-run fixes Travis.

PR #1773, opened 2020-04-13
Add event support as part of TokenScript 2020/03

### LINKED ISSUE #1729: iOS: interpretation a filter (currently only key=value)
@colourful-land said this:

I envision the final product is a filter that can be used to obtain object in:

1) attestations, especially when one attestation depends on another with specific predicate (which is a filter)
2) tokens, during token negotiation
3) data objects, e.g. finding events.

But in this issue, we only need a filter that can obtain an event through a key=value format. That is, αWallet can parse the filter used in line 139

https://github.com/AlphaWallet/TokenScript-Examples/blob/27ffe6a83a7c41631ebfe22e262e21850f234327/examples/ENS/ENS.xml#L137-L141

### LINKED ISSUE #1732: iOS adding ASNX module parsing capacity (only used for events now)
It means iOS should be aware of this file's line 26 to line 34

https://github.com/AlphaWallet/TokenScript-Examples/blob/27ffe6a83a7c41631ebfe22e262e21850f234327/examples/ENS/ENS.xml#L26-L34

### LINKED ISSUE #1733: realm database schema, which has event-attribute relation, raw (filtered) event data and cached token attributes

### LINKED ISSUE #1740: Add ability to retrieve events inside TokenScript

hboon:
Closes #1729, closes #1732, closes #1733, closes #1740 

Doesn't include schema change: "selections" and "attribute-types" wrapper tags

### DIFF
--- a/AlphaWallet/AppDelegate.swift
+++ b/AlphaWallet/AppDelegate.swift
@@ -52,7 +52,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
     func applicationWillResignActive(_ application: UIApplication) {
         protectionCoordinator.applicationWillResignActive()
     }
-    
+
     func applicationDidBecomeActive(_ application: UIApplication) {
         //Lokalise.shared.checkForUpdates { _, _ in }
         protectionCoordinator.applicationDidBecomeActive()
@@ -90,7 +90,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
         //TODO: if we handle other types of URLs, check if handled==false, then we pass the url to another handlers
         return handled
     }
-    
+
     // Respond to amazon SNS registration
     func application(_ application: UIApplication,
                      didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
--- a/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
+++ b/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
@@ -1,0 +1,149 @@
+// Copyright © 2020 Stormbird PTE. LTD.
+
+import Foundation
+import BigInt
+import PromiseKit
+import web3swift
+
+class EventSourceCoordinator {
+    private var wallet: Wallet
+    private let config: Config
+    private let tokensStorages: ServerDictionary<TokensDataStore>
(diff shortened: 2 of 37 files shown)


AlphaWallet/AppDelegate.swift
```diff
@@ -52,7 +52,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDele
     func applicationWillResignActive(_ application: UIApplication) {
         protectionCoordinator.applicationWillResignActive()
     }
-
```
James-Sangalli, line 55: Opps!

hboon: Errorenous whitespace editing? I blame it on AppCode's auto saving :P

AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
```diff
@@ -0,0 +1,149 @@
...
+            //TODO support things like "$prefix-{tokenId}"
+            optionalFilter = nil
+        }
+        guard let (filterValue, textEquivalent) = optionalFilter else { return nil }
+        guard let filterValueTypedForEventFilters = filterValue.coerceToArgumentTypeForEventFilter(parameterType) else { return nil }
+        return (filter: [filterValueTypedForEventFilters], textEquivalent: textEquivalent)
+    }
+
+    static func convertToImplicitAttribute(string: String) -> AssetImplicitAttributes? {
+        let prefix = "${"
```
James-Sangalli, line 143: Hmm, I thought this was standard JS? Swift can't parse that more naturally?

James-Sangalli:
@hboon closes https://github.com/AlphaWallet/alpha-wallet-ios/issues/1740 ?

hboon:
@James-Sangalli yes, have updated 1st comment. Bad Boon.

AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
hboon: Not that I know of. II wanted to make it replace it verbatim, eg. "${tokenId}", but not "somePrefix#{tokenId}" for now. So it ended up looking like that without using Regex.

Any idea how to improve it?

hboon:
@James-Sangalli updated 1st comment. This PR actually closes several issues.

PR #1782, opened 2020-04-21
Support <selection> for filtering actions

### LINKED ISSUE #1776: Support <selection> for filtering actions
- [x] "normal" TokenScripts
- [x] action-only TokenScripts

In scope:

* Define selections:
    * `/<token>/<selection>`
    * `<action>/<selection>`
* Specifing a selection:
    * `/<action>/<exclude selection="">`
    * `<action exclude="">`

Out of scope:

* `<action available-to="">` (and any form of `available-to`)

hboon:
Closes #1776

@James-Sangalli would you refer to #1776 for what is in-scope/out-of-scope and buzz me if there's something unexpected in the scope?

- [x] Can't  close #1776 yet. Pending https://github.com/AlphaWallet/TokenScript/issues/340.

Have included test cases for the parser: `TokenScriptFilterParserTests`


hboon:
@James-Sangalli making this WIP since the grammer for the filter expressions aren't documented yet. Feel free to merge if you think what's available is enough for you. The current stage of this PR wouldn't work with action-only scripts for ERC20 for example since there's no attribute to filter with (and no `balance` implicit attribute).

hboon:
@James-Sangalli feel free to test, but I'll make amendments to the filter expression parsing after talking to Weiwu. `<selection>` etc will continue to work the same way. Filtering grammer is going to be a superset of what is in this PR now, so if you have tested anything, what has worked should continue to work.

hboon:
> The enable button still shows up

@James-Sangalli should fix that. As well as hiding actions button if denial message is not provided for a matching selection.

hboon:
@James-Sangalli This PR now also includes the updated parser to handle https://github.com/AlphaWallet/TokenScript/issues/340.

Should be complete now

PR #1793, opened 2020-05-11
Support local-ref

### LINKED ISSUE #1790: Implement ts:local-ref

hboon:
Closes  #1790

@James-Sangalli Left a couple of comments that might cause this PR to not work as expected:

- [x] https://github.com/AlphaWallet/TokenScript-Examples/pull/15#discussion_r422803989
- [x] https://github.com/AlphaWallet/TokenScript-Examples/pull/15#discussion_r422805104
- [x] https://github.com/AlphaWallet/TokenScript-Examples/pull/15#discussion_r422805285
- [x] https://github.com/AlphaWallet/TokenScript-Examples/pull/15#discussion_r422805890

### DIFF
--- a/AlphaWallet/Browser/Factory/WKWebViewConfiguration.swift
+++ b/AlphaWallet/Browser/Factory/WKWebViewConfiguration.swift
@@ -69,6 +69,8 @@ extension WKWebViewConfiguration {
         webViewConfig.userContentController.add(messageHandler, name: Method.signPersonalMessage.rawValue)
         webViewConfig.userContentController.add(messageHandler, name: Method.signMessage.rawValue)
         webViewConfig.userContentController.add(messageHandler, name: Method.signTypedMessage.rawValue)
+        //TODO extract like `Method.signTypedMessage.rawValue` when we have more than 1
+        webViewConfig.userContentController.add(messageHandler, name: TokenInstanceWebView.SetProperties.setActionProps)
         return webViewConfig
     }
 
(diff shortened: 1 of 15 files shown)


hboon:
@James-Sangalli can you paste me the wallet address somewhere?

James-Sangalli:
> @James-Sangalli can you paste me the wallet address somewhere?

0x42ea529282DDE0AA87B42d9E83316eb23FE62c3f

hboon:
> Email doesn't resolve (might be because of the comment, I will try that again later) 

@James-Sangalli forced pushed with fix

> but regardless of this it hits an exception of line 122 in AssetAttribute (which shouldn't happen).

How do I reproduce this?

James-Sangalli:
> > Email doesn't resolve (might be because of the comment, I will try that again later)
> 
> @James-Sangalli forced pushed with fix
> 
> > but regardless of this it hits an exception of line 122 in AssetAttribute (which shouldn't happen).
> 
> How do I reproduce this?

Happened by simply clicking through the ENS token card. I am testing your new update now, will get back to you.

James-Sangalli:
@hboon was able to successfully set a new email record: https://etherscan.io/tx/0xe76d1c55776ada393875765a1f41a60f70b9294812c579b6b2c3ae13a1abde39

However, the price for renewal does not load. To reproduce simply click on the renewal action and you will notice the price is set to zero. The console shows that it comes back as a "0" string with no errors.

hboon:
@James-Sangalli fixed. Action attributes that depend on event-based attributes weren't resolved.

This is still true: both token-level and action-level attributes that are dependent on function-based attributes are still not supported. It's complicated. At least for iOS, we might need to rewrite some parts of the resolution code.

hboon:
> The crash has disappeared.

Let me know if you figure it out. There's probably a bug somewhere.

James-Sangalli:
photo_2020-05-14_10-34-51

The price resolves but the value is not registered in the function call data, seems it is missing the ability to take value params as references. Possible to fix in this PR?

hboon:
> The price resolves but the value is not registered in the function call data, seems it is missing the ability to take value params as references. Possible to fix in this PR?

@James-Sangalli Done

PR #1736, opened 2020-02-24
schema 2020 03

James-Sangalli:
To support https://github.com/AlphaWallet/TokenScript/pull/332

Merge once all schema related tasks are done and we have TokenScripts served. 

TODO 
- [x] #1730
- [x] #1729 
- [x] #1728 
- [x] #1732 
- [x] #1733 
- [x] #1734 
- [x] #1735 
- [x] #1740 
- [x] #1775 
- [x] #1776 
- [x] #1777
- [x] #1787 
- [x] #1790
- [x] #1808

PR #1800, opened 2020-05-15
Upgrade schema 2020 06

### LINKED ISSUE #1797: Implement schema 2020/06
Do after #1736 

Source: https://github.com/AlphaWallet/TokenScript/pull/345

James-Sangalli:
merge after #1736 

Closes #1797


AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -829,38 +812,79 @@ extension XMLHandler {
...
 
+    static func getEventParameterName(fromEthereumEventElement ethereumEventElement: XMLElement) -> String? {
+        guard let eventParameterName = ethereumEventElement["select"] else { return nil }
+        return eventParameterName
+    }
+
+    static func getEventDefinition(fromContractElement contractElement: XMLElement, xmlContext: XmlContext) -> EventDefinition? {
+        let addressElements = XMLHandler.getAddressElements(fromContractElement: contractElement, xmlContext: xmlContext)
+        guard let address = addressElements.first?.text.flatMap({ AlphaWallet.Address(string: $0.trimmed)}) else { return nil }
+        guard let eventName = contractElement.at_xpath("asnx:module", namespaces: xmlContext.namespaces)?["name"] else { return nil }
```
hboon, line 831: Is the attribute ("name") renamed to "module" now?

```
    <xs:element name="event">
        <xs:complexType>
            <!-- TODO: this should be whatever ASN.X allows as element name -->
            <xs:attribute name="module" type="xs:NCName"/>
            <xs:attribute name="filter" type="xs:string"/>
            <xs:attribute name="select" type="xs:NCName"/>
        </xs:complexType>
    </xs:element>
```

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -800,7 +783,7 @@ extension XMLHandler {
     }
 
     fileprivate static func getServerForNativeCurrencyAction(fromRoot root: XMLDocument, xmlContext: XmlContext) -> RPCServer? {
-        return root.at_xpath("/action/input/token/ethereum".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)?["network"].flatMap { Int($0) }.flatMap { RPCServer(chainID: $0) }
+        return root.at_xpath("/card/input/token/ethereum:transfer".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)?["network"].flatMap { Int($0) }.flatMap { RPCServer(chainID: $0) }
```
hboon, line 786: This wouldn't work because `addToXPath()` breaks up the path components and adds the prefix `xmlContext.namespacePrefix` to each of them. `addToXPath()` is used because adding the prefix to each component in that string literal was getting untidy.

So probably have to do this instead (untested):

```
let path = "/card/input/token".addToXPath(namespacePrefix: xmlContext.namespacePrefix) + "/ethereum:transfer"
return root.at_xpath(path, namespaces: xmlContext.namespaces)?["network"].flatMap { Int($0) }.flatMap { RPCServer(chainID: $0) }
```

hboon: But I don't think this line should be changed. It's still "ethereum", not "ethereum:transfer" because this is used for action cards, like this:

```
<ts:input>
    <ts:token name="xdai">
        <ts:ethereum network="100"/>
    </ts:token>
</ts:input>
```

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -896,50 +920,91 @@ extension XMLHandler {
...
     }
 
     fileprivate static func getTokenScriptActionOnlyActionElements(fromRoot root: XMLDocument, xmlContext: XmlContext) -> XPathObject {
-        return root.xpath("/action".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
+        return root.xpath("/card".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
     }
 
     fileprivate static func getActionTransactionFunctionElement(fromActionElement actionElement: XMLElement, xmlContext: XmlContext) -> XMLElement? {
-        return actionElement.at_xpath("transaction/ethereum".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
+        return actionElement.at_xpath("transaction/ethereum:call".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
```
hboon, line 943: @colourful-land I remember reading that @James-Sangalli feedback on this too. But Ethereum docs and tutorials say:

call — invoke without gas + read-only
transaction - invoke function with gas

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
hboon: So the change for "ethereum:transfer" has to be made somewhere else

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
colourful-land: Jump to the discussion here. Happy to change the current behaviour after collected feedbacks

colourful-land: @James-Sangalli merge https://github.com/AlphaWallet/TokenScript-Examples/pull/78 and https://github.com/AlphaWallet/TokenScript/pull/353 so that you can address this change request

hboon:
@James-Sangalli on mobile so will try to check back. But the change to for the XPaths to extract cards might need to specify if they are for token or action. Maybe. Just in case you merge.

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
James-Sangalli: this has changed again to ethereum:value so I am removing the above change and keeping it as is (value is still called value but with the ethereum: prefix rather than ts: prefix, the xmlcontext should get this I believe)

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
James-Sangalli: <ts:contract name="ETHRegistrarController">
    <ts:address network="1">0x283af0b28c62c092c9727f1ee09c02ca627eb7f5</ts:address>
    <asnx:module name="NameRegistered">
      <sequence>
        <element name="name" ethereum:type="string"/>
        <element name="label" ethereum:type="bytes32" ethereum:indexed="true"/>
        <element name="owner" ethereum:type="address" ethereum:indexed="true"/>
        <element name="cost" ethereum:type="uint"/>
        <element name="expires" ethereum:type="uint"/>
      </sequence>
    </asnx:module>
  </ts:contract>

colourful-land: > Has the attribute ("name") renamed to "module" now?

No. `moulde="xxx"` has been there for months, way before our renaming.

If we allowed events to have name, it would mean the name of the event:

    <event name="Birthday">

This means the event is using a data object called "Birthday"

    <event module="Birthday">

In ASN.X terminology, ASN.1 Data Object's structure/schema is called a module.

e.g. you have 3 data objects about 3 people in the module "person".

hboon: For posterity: I mean "event" when I typed "name" above, and I was commenting on the wrong snippet. Sorry. James should know what I am referring to now :)

colourful-land:
> @James-Sangalli on mobile so will try to check back. But the changes to the XPaths to extract cards might need to specify if they are for token or action. Maybe. Just in case you merge.

@James-Sangalli the xpaths are `cards/card[@type="action"]` and `cards/card[@type="token"]`. Verified here:

    ENS$ xqilla -i ENS.canonicalized.xml <(echo 'declare default element namespace "http://tokenscript.org/2020/06/tokenscript";count(/token/cards/card[@type="action"])')

AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -932,11 +932,11 @@ extension XMLHandler {
     }
 
     fileprivate static func getTokenScriptTokenInstanceCardElements(fromRoot root: XMLDocument, xmlContext: XmlContext) -> XPathObject {
-        return root.xpath("/token/cards/card".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
+        return root.xpath("/card[@type=\"token\"]".addToXPath(namespacePrefix: xmlContext.namespacePrefix), namespaces: xmlContext.namespaces)
```
hboon, line 935: FYI. You can use single quotes to avoid escaping.

colourful-land: Does it actually work? "/card" means the root element is `/card` which is true only for X-Dai bridge.

AlphaWallet/TokenScriptClient/Models/FunctionOrigin.swift
```diff
@@ -257,60 +272,72 @@ struct FunctionOrigin {
...
         }
     }
 
-    fileprivate static func createInput(fromInputElement inputElement: XMLElement, withInputType inputType: SolidityType) -> AssetFunctionCall.Argument? {
+    fileprivate static func createInput(fromInputElement inputElement: XMLElement, root: XMLDocument, xmlContext: XmlContext, withInputType inputType: SolidityType) -> AssetFunctionCall.Argument? {
         if let inputName = inputElement["ref"].nilIfEmpty {
             return .ref(ref: inputName, type: inputType)
+        } else if let inputName = inputElement["local-ref"].nilIfEmpty {
+            let attributes = XMLHandler.getCardAttributeTypeElements(fromRoot: root, xmlContext: xmlContext)
+            let attributeIds = attributes.compactMap { $0["id"] }
```
hboon, line 335: @James-Sangalli The "id" introduced from the recent commit in 2020/03 might need to be changed to "name" too

James-Sangalli: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1811

James-Sangalli:
@hboon ready for review

PR #1807, opened 2020-05-20
Make local-ref work with views too, so they work with both normal views and action views, but not item-views, and tweaking local-ref resolution so that it covers both props and card attributes

### LINKED ISSUE #1790: Implement ts:local-ref

hboon:
Closes #1790 

This PR:

* Makes local-ref work with views too (but not item-view)
* local-ref resolution now includes card attributes in addition to props

@James-Sangalli Better to assume all the attribute ids/names are still in a single namespace for now as an implementation incompatibility. Just have to not give attributes the same names. I'll fix it after we get 2020/06 is merged. It takes a bigger refactor before being able to do it properly

### DIFF
--- a/AlphaWallet/UI/Views/TokenCardRowView.swift
+++ b/AlphaWallet/UI/Views/TokenCardRowView.swift
@@ -28,6 +28,7 @@ class TokenCardRowView: UIView, TokenCardRowViewProtocol {
 	private let spaceAboveBottomRowStack = UIView.spacer(height: 10)
 	private var checkboxRelatedConstraintsWhenShown = [NSLayoutConstraint]()
 	private var checkboxRelatedConstraintsWhenHidden = [NSLayoutConstraint]()
+	private var lastTokenHolder: TokenHolder?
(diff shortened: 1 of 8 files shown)


AlphaWallet/UI/Views/TokenCardRowView.swift
```diff
@@ -359,6 +362,12 @@ extension TokenCardRowView: TokenInstanceWebViewDelegate {
 	}
 
 	func reinject(tokenInstanceWebView: TokenInstanceWebView) {
-		//no-op
+		//Refresh if view, but not item-view
+        if isStandalone {
```
James-Sangalli, line 366: opps :)

hboon: You mean it was a bug? It's by design :P I thought it was supposed to only work for action views. We agreed on: "web3.action.setProps".

hboon: Oh the tab+space. Damn relic.

PR #1813, opened 2020-05-20
modifies readme

hellolucas:
- adds main image
- changes android to iOS mentions
- adds maintenance, contributor and license badges


James-Sangalli:
Excellent, but please point the MIT license badge to our licence rather than a sample one

PR #1799, opened 2020-05-14
Display suggestions when importing seed phrase #1568

nimitparekh2020:
Fixes #1568

PR #1821, opened 2020-05-22
Display suggestions from wordlist when typing seed phrase to import

### LINKED ISSUE #1568: Display suggestions when importing seed phrase
Please use Moonlight app as a reference. 
Please improve the look of segmented controls. 

Import_with_Seed_iOS

hboon:
Closes #1568 

Supersedes #1799 

After PR:


James-Sangalli:
@hboon looks good but when you get to the next work the last word hangs there, is that supposed to happen?

hboon:
@James-Sangalli I guess we can remove the suggestions when there's only 1 word and it matches the last word?

I just noticed a strange side-effect, if you type: "tea" and it suggests "teach team" and you tap one of the words, the suggests "teach team" still appear. It wasn't intended but looks like it's nice to keep it around because you can tap again to correct it. What do you think?

James-Sangalli:
> @James-Sangalli I guess we can remove the suggestions when there's only 1 word and it matches the last word?
> 
> I just noticed a strange side-effect, if you type: "tea" and it suggests "teach team" and you tap one of the words, the suggests "teach team" still appear. It wasn't intended but looks like it's nice to keep it around because you can tap again to correct it. What do you think?

well it should only suggest one word, cannot suggest two words at once... 

Why not simply clear the last suggestion and start fresh when you go onto the new word? The past word has zero influence on the next.

hboon:
It's autocompletion.

tea -> Possible words are "teach" and "team". So 2 words suggested, with the prefix "tea".

It still displays the suggestions "teach" and "team" after tapping on one of them, but before typing the next word. So while it's a bit odd, I'm thinking that the user can tap it again to choose a different word if they have tapped wrongly. eg.

1\. Start with ""
2\. Type "tea". Suggests "teach" and "team"
3\. Tap "teach". Now shows "teach" in the textbox. Still suggests "teach" and "team"
4\. Whoops, tapped wrongly. Tap "team". Now shows "team in the textbox.

hboon:
> Why not simply clear the last suggestion and start fresh when you go onto the new word? The past word has zero influence on the next.

Yes, that or keep the strange behavior, which seems useful https://github.com/AlphaWallet/alpha-wallet-ios/pull/1821#issuecomment-632573266

PR #1823, opened 2020-05-22
Refactor all access to QRCodeReaderViewController

### LINKED ISSUE #1822: Refactor all access to QRCodeReaderViewController
Refactor all access to QRCodeReaderViewController so it goes through ScanQRCodeCoordinator

vladyslav-iosdev:
Closes #1822


hboon:
FYI: "Closes #xxx" has to be together, i.e not "Closes something something #xxx"

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
```diff
@@ -3,6 +3,7 @@
 import Foundation
 import UIKit
 import PromiseKit
+import AVFoundation
```
hboon, line 6: Can you wrap this into a static function in `ScanQRCodeCoordinator` (or into the `CanScanQRCode` protocol's default extension, look for where `promptUserOpenSettingsToChangeCameraPermission()`  is defined, but maybe that's just for view controllers) so that `AVFoundation` doesn't need to be imported in the few classes that do now?

```
AVCaptureDevice.authorizationStatus(for: .video) != .denied
```

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -8,36 +8,52 @@ protocol ScanQRCodeCoordinatorDelegate: class {
...
+            bordersColor: Colors.qrCodeRectBorders,
+            messageText: R.string.localizable.qrCodeTitle(),
+            torchTitle: R.string.localizable.light(),
+            torchImage: R.image.light(),
+            chooseFromPhotoLibraryButtonImage: R.image.browse()
+        )
         controller.delegate = self
-        controller.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(dismiss))
+        controller.title = R.string.localizable.browserScanQRCodeTitle()
+        controller.navigationItem.leftBarButtonItem = UIBarButtonItem(image: R.image.backWhite(), style: .plain, target: self, action: #selector(dismiss))
```
hboon, line 33: - [ ] Let's make this the standard `Cancel` button for now. (it's odd to present modally and have back button, and this does feel like this should be a modal operation).

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -8,36 +8,52 @@ protocol ScanQRCodeCoordinatorDelegate: class {
...
+        navigationController.dismiss(animated: true, completion: nil)
+        delegate?.didCancel(in: self)
     }
 }
 
 extension ScanQRCodeCoordinator: QRCodeReaderDelegate {
+    
     func readerDidCancel(_ reader: QRCodeReaderViewController!) {
         reader.stopScanning()
         reader.dismiss(animated: true)
```
hboon, line 59: - [ ] This has to be changed to call `navigationController.dismiss(animated: true, completion: nil)` instead.

- [ ] Also the same line in `reader(_:didScanResult:)`

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -8,36 +8,52 @@ protocol ScanQRCodeCoordinatorDelegate: class {
...
+        controller.navigationItem.leftBarButtonItem = UIBarButtonItem(image: R.image.backWhite(), style: .plain, target: self, action: #selector(dismiss))
         controller.delegate = self
         return controller
     }()
-
-    init(
-        navigationController: NavigationController = NavigationController()
-        ) {
+    
+    let navigationController: NavigationController
```
hboon, line 38: - [ ] Can we make it typed `UINavigationController` instead?

- [ ] Can you make this private (if you need to keep it as a property, I think it can just be a local variable)? It can be created and used internally within `ScanQRCodeCoordinator` and not created and passed it from the call site

- [ ] Create another UINavigationController property, this is the one that is passed in and let `ScanQRCodeCoordinator` handle the presenting and dismissing of the private UINavigationController which is holding `QRCodeReaderViewController`.

eg. in `DappBrowserCoordinator`, change it from:

```
let coordinator = ScanQRCodeCoordinator(navigationController: NavigationController())
coordinator.delegate = self
addCoordinator(coordinator)
coordinator.start()
navigationController.present(coordinator.navigationController, animated: true, completion: nil)
```

to:

```
let coordinator = ScanQRCodeCoordinator(navigationController: navigationController) //Pass in `navigationController`
coordinator.delegate = self
addCoordinator(coordinator)
coordinator.start() //move the next line into `start()`
//navigationController.present(coordinator.navigationController, animated: true, completion: nil)
```

I know this isn't consistent across the app, but refer to `SendCoordinator`'s start() and its delegate for an example:

```
navigationController.pushViewController(sendViewController, animated: true)
```

Generally, we should move the codebase towards doing this. The child coordinator is the one responding for showing and hiding the view controller it manages

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
vladyslav-iosdev: what if i will expand `CanScanQRCode` with next function?
```
func requestDeviceAuthorization() -> Bool {
        guard AVCaptureDevice.authorizationStatus(for: .video) != .denied else {
            promptUserOpenSettingsToChangeCameraPermission()
            return false
        }
        return true
    }
```
and all we need to do its just 
`guard navigationController.requestDeviceAuthorization() else { return }`

hboon: Looks good, but let's call it `ensureHasDeviceAuthorization()` instead? "request" sounds like it will request, and if the request failed, it will return false, but the request hasn't been processed yet.

vladyslav-iosdev: @hboon yes, shure it much better name. i will update it, and also update 
`"photos" = "Browse";`
do i need to make squash fore these commits?

hboon: Yes, squash them since it looks like they all belong as 1 commit.

hboon:
@vladyslav-iosdev is this ready for review?

vladyslav-iosdev:
> @vladyslav-iosdev is this ready for review?

yes

hboon:
>> @vladyslav-iosdev is this ready for review? ...

@vladyslav-iosdev would you help add me as reviewer (or click the refresh button beside my name in reviewer list) in the PRs in the future? Easier for me to track. Just for the future ones. The few you have just replied to is fine.

PR #1824, opened 2020-05-22
Add copy button to Transaction Details + rename "More Details" to "View in Etherscan"

### LINKED ISSUE #1792: Add copy button to Transaction Details + rename "More Details" to "View in Etherscan"
Visual (by Tomek, if in question, ask him)
https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/dashboard?seid=5e170e6ef7b0f25419ad8e00

Android: https://github.com/AlphaWallet/alpha-wallet-android/issues/1377

vladyslav-iosdev:
Closes #1792

### DIFF
--- a/AlphaWallet/Settings/Types/ConfigExplorer.swift
+++ b/AlphaWallet/Settings/Types/ConfigExplorer.swift
@@ -11,8 +11,9 @@ struct ConfigExplorer {
         self.server = server
     }
 
-    func transactionURL(for ID: String) -> URL? {
-        guard let endpoint = explorer(for: server) else { return .none }
+    func transactionURL(for ID: String) -> (url: URL, name: String?)? {
+        let result = explorer(for: server)
+        guard let endpoint = result.url else { return .none }
         let urlString: String? = {
             switch server {
             case .poa:
@@ -23,36 +24,57 @@ struct ConfigExplorer {
                 return endpoint + "/tx/" + ID
             }
         }()
-        guard let string = urlString else { return .none }
-        return URL(string: string)!
+        guard let string = urlString, let url = URL(string: string) else { return .none }
+
+        return (url: url, name: result.name)
     }
 
-    private func explorer(for server: RPCServer) -> String? {
+    func explorerName(for server: RPCServer) -> String? {
+        switch server {
+        case .main, .kovan, .ropsten, .rinkeby, .goerli:
+            return "Etherscan"
+        case .classic:
+            return "Gastracker"
+        case .poa:
+            return "POA Explorer"
+        case .custom, .callisto:
+            return nil
+        case .sokol:
+            return "Sokol Explorer"
+        case .xDai:
+            return "Blockscout"
+        case .artis_sigma1, .artis_tau1:
+            return "ARTIS" 
+        }
+    }
+
+    private func explorer(for server: RPCServer) -> (url: String?, name: String?) {
+        let nameForServer = explorerName(for: server)
         switch server {
         case .main:
-            return "https://etherscan.io"
+            return ("https://etherscan.io", nameForServer)
         case .classic:
-            return "https://gastracker.io"
+            return ("https://gastracker.io", nameForServer)
         case .kovan:
-            return "https://kovan.etherscan.io"
+            return ("https://kovan.etherscan.io", nameForServer)
         case .ropsten:
-            return "https://ropsten.etherscan.io"
+            return ("https://ropsten.etherscan.io", nameForServer)
         case .rinkeby:
-            return "https://rinkeby.etherscan.io"
+            return ("https://rinkeby.etherscan.io", nameForServer)
         case .poa:
-            return "https://poaexplorer.com"
+            return ("https://poaexplorer.com", nameForServer)
         case .sokol:
-            return "https://sokol-explorer.poa.network"
+            return ("https://sokol-explorer.poa.network", nameForServer)
         case .xDai:
-            return "https://blockscout.com/poa/dai/"
+            return ("https://blockscout.com/poa/dai/", nameForServer)
         case .goerli:
-            return "https://goerli.etherscan.io"
+            return ("https://goerli.etherscan.io", nameForServer)
         case .artis_sigma1:
-            return "https://explorer.sigma1.artis.network"
+            return ("https://explorer.sigma1.artis.network", nameForServer)
         case .artis_tau1:
-            return "https://explorer.tau1.artis.network"
+            return ("https://explorer.tau1.artis.network", nameForServer)
         case .custom, .callisto:
-            return .none
+            return (.none, nameForServer)
         }
     }
 }
--- a/AlphaWallet/Transactions/ViewControllers/TransactionViewController.swift
+++ b/AlphaWallet/Transactions/ViewControllers/TransactionViewController.swift
@@ -3,6 +3,7 @@
 import UIKit
 import Result
 import SafariServices
+import MBProgressHUD
 
 protocol TransactionViewControllerDelegate: class, CanOpenURL {
 }
@@ -49,12 +50,12 @@ class TransactionViewController: UIViewController {
             .spacer(),
             header,
             .spacer(),
-            item(title: viewModel.fromLabelTitle, value: viewModel.from),
-            item(title: viewModel.toLabelTitle, value: viewModel.to),
+            item(title: viewModel.fromLabelTitle, value: viewModel.from, icon: R.image.copy()),
+            item(title: viewModel.toLabelTitle, value: viewModel.to, icon: R.image.copy()),
             item(title: viewModel.gasFeeLabelTitle, value: viewModel.gasFee),
             item(title: viewModel.confirmationLabelTitle, value: viewModel.confirmation),
             .spacer(),
-            item(title: viewModel.transactionIDLabelTitle, value: viewModel.transactionID),
+            item(title: viewModel.transactionIDLabelTitle, value: viewModel.transactionID, icon: R.image.copy()),
             item(title: viewModel.createdAtLabelTitle, value: viewModel.createdAt),
             item(title: viewModel.blockNumberLabelTitle, value: viewModel.blockNumber),
         ]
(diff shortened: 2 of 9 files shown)


hboon:
Have forced pushed and squashed into a single commit

AlphaWallet/Transactions/ViewControllers/TransactionViewController.swift
```diff
@@ -119,36 +120,38 @@ class TransactionViewController: UIViewController {
...
-        let copyAction = UIAlertAction(title: R.string.localizable.copy(), style: .default) { _ in
-            UIPasteboard.general.string = value
-        }
-        let cancelAction = UIAlertAction(title: R.string.localizable.cancel(), style: .cancel) { _ in }
-        alertController.addAction(copyAction)
-        alertController.addAction(cancelAction)
-        present(alertController, animated: true, completion: nil)
+        let hud = MBProgressHUD.showAdded(to: view, animated: true)
+        hud.mode = .text
+        hud.label.text = viewModel.addressCopiedText
```
hboon, line 144: - [ ] The bubble that appears shouldn't say "Address copied" when it's the "transaction #" or "Gas Fee", etc

AlphaWallet/Transactions/ViewModels/TransactionDetailsViewModel.swift
```diff
@@ -54,9 +54,22 @@ struct TransactionDetailsViewModel {
...
+    }
 
     var detailsURL: URL? {
-        return ConfigExplorer(server: server).transactionURL(for: transaction.id)
+        return ConfigExplorer(server: server).transactionURL(for: transaction.id)?.url
+    }
+    
+    var detailsButtonText: String {
+        if let name = ConfigExplorer(server: server).transactionURL(for: transaction.id)?.name {
+            let value = R.string.localizable.viewIn()
```
hboon, line 68: - [ ] You can set the localisation string as this instead:

```
"view in" = "View in %";
```

And `R` will generate a function with an argument instead, so you can do:

```
return R.string.localizable.viewIn(name)
```

AlphaWallet/Settings/Types/ConfigExplorer.swift
```diff
@@ -23,36 +24,59 @@ struct ConfigExplorer {
...
+        case .poa:
+            return "POA Explorer"
+        case .custom, .callisto:
+            return nil
+        case .sokol:
+            return "Sokol Explorer"
+        case .xDai:
+            return "BlockScout"
+        case .artis_sigma1:
+            return "Artis"
```
hboon, line 47: - [ ] return "ARTIS" instead.

hboon:
@vladyslav-iosdev can you help to Squash and rebase master? Shouldn’t have the merge commit

hboon:
@vladyslav-iosdev test suite seems to be failing. Can you make them pass? You need to run in the iOS 12.4 simulator (13.x will fail because there's a bug with keychain access with 13.x)

followed by this: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1824#issuecomment-634652932

vladyslav-iosdev:
> @vladyslav-iosdev
> 
> * [ ]  test suite seems to be failing. Can you make them pass? You need to run in the iOS 12.4 simulator (13.x will fail because there's a bug with keychain access with 13.x). Test suite might be failing because we broke it recently. If you `git rebase master`, it should bring in the latest master which has the fix.
> * [ ]  followed by this: [#1824 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/1824#issuecomment-634652932)

seems like done.

hboon:
@vladyslav-iosdev squash them into a single commit?

vladyslav-iosdev:
> @vladyslav-iosdev squash them into a single commit?

@hboon i tried to perform  squash, but i am now shure if its done.

hboon:
Maybe there's a bit of confusion because there are 2 operations that are performed using very similar commands (blame git UX):

> > @vladyslav-iosdev squash them into a single commit?
> 
> @hboon i tried to perform squash, but i am not shure if its done.


* rebase master: `git rebase master`
* squash: `git rebase -i <commit>`

Screenshot 2020-05-29 at 12 02 45 PM

Nope. But I don't think it's a problem with squashing yet. If you look at the Commits tab in this page, you'll see there are a few commits titled: "Merge branch of '#1792' ...". Each one is likely caused because the latest changes from master (at that point) is merged into your feature branch ("#1792") rather than rebased. Could you have pull master when you are checked out in the feature branch? The command to run then should have been:

```
(while feature-branch is checked out)
git fetch origin master:master #replace origin with remote that points to upstream AlphaWallet repo. This updates your local `master` branch
git rebase master #This rewrites your feature-branch so that it looks like feature-branch is based on the updated master rather than the version of `master` when you created feature-branch
```

eg. master was (oldest commits first):

```
c1
c2
c3 <- master
```

you create feature-branch and add commits f1 and f2:

```
c1
c2
c3 <- master
f1
f2 <- feature-branch
```

master is now updated to (because we merge other PRs, etc):

```
c1
c2
c3
c4 <- new master (and you will have this in your local master after you do `git fetch <upstream-remote) master:master`)
```

After `git rebase master` while you have feature-branch checked out, it becomes:

```
c1
c2
c3
c4 <- master
f1'
f2' <- feature-branch
```

Notice where c4 is included in feature-branch, and note that f1' and f2' replaces f1 and f2. Sometimes there's merge conflicts while running git rebase master because git is trying to figure out how to form f1' and f2'. [1]

And then you can squash:

Let's say, after rebase master from earlier, you have this:

```
c1
c2
c3
c4 <- master
f1'
f2' <- feature-branch
```

We want to squash f1' and f2' into a single commit, so we do:

```
git rebase -i c4 #c4 is the commit just older than the one you want to change
```

And git should open a text editor like this:

```
pick f1'
pick f2'
```

Change it to this and save:

```
pick f1'
squash f2' (if there's f3, f4, etc and you want them to all be squashed into f1', you'd change them to squash too)
```

Then we should end up with [2]:

```
c1
c2
c3
c4 <- master
f1'' <- feature-branch
```

---

At point [1]:

If we had done a merge instead of a rebase master, we would have gotten:

```
c1
c2
c3
f1
f2
c4
m1 <- feature-branch #This is a merge commit for c4 + f2 
```

We don't want to do this ^. Let's call this [3]. Compare [2] and [3]. [2] has several advantages over [3]:

A. Much easier and correct to review, because it's applied over the current master
B. Because it's applied over the current master, the Git UI can merge it cleanly. Any merge conflict you had to handle during the rebase master operation would have had to be handled here instead anyway, so there's actually no extra work.
C. The resulting history after merge is clearer, so we can look up where changes were made and why they were made (which goes back to clean commit messages which states the reason for the change and summary of the change)

vladyslav-iosdev:
> Maybe there's a bit of confusion because there are 2 operations that are performed using very similar commands (blame git UX):
> 
> > > @vladyslav-iosdev squash them into a single commit?
> > 
> > 
> > @hboon i tried to perform squash, but i am not shure if its done.
> 
> * rebase master: `git rebase master`
> * squash: `git rebase -i <commit>`
> 
> Screenshot 2020-05-29 at 12 02 45 PM
> 
> Nope. But I don't think it's a problem with squashing yet. If you look at the Commits tab in this page, you'll see there are a few commits titled: "Merge branch of '#1792' ...". Each one is likely caused because the latest changes from master (at that point) is merged into your feature branch ("#1792") rather than rebased. Could you have pull master when you are checked out in the feature branch? The command to run then should have been:
> 
> ```
> (while feature-branch is checked out)
> git fetch origin master:master #replace origin with remote that points to upstream AlphaWallet repo. This updates your local `master` branch
> git rebase master #This rewrites your feature-branch so that it looks like feature-branch is based on the updated master rather than the version of `master` when you created feature-branch
> ```
> 
> eg. master was (oldest commits first):
> 
> ```
> c1
> c2
> c3 <- master
> ```
> 
> you create feature-branch and add commits f1 and f2:
> 
> ```
> c1
> c2
> c3 <- master
> f1
> f2 <- feature-branch
> ```
> 
> master is now updated to (because we merge other PRs, etc):
> 
> ```
> c1
> c2
> c3
> c4 <- new master (and you will have this in your local master after you do `git fetch <upstream-remote) master:master`)
> ```
> 
> After `git rebase master` while you have feature-branch checked out, it becomes:
> 
> ```
> c1
> c2
> c3
> c4 <- master
> f1'
> f2' <- feature-branch
> ```
> 
> Notice where c4 is included in feature-branch, and note that f1' and f2' replaces f1 and f2. Sometimes there's merge conflicts while running git rebase master because git is trying to figure out how to form f1' and f2'. [1]
> 
> And then you can squash:
> 
> Let's say, after rebase master from earlier, you have this:
> 
> ```
> c1
> c2
> c3
> c4 <- master
> f1'
> f2' <- feature-branch
> ```
> 
> We want to squash f1' and f2' into a single commit, so we do:
> 
> ```
> git rebase -i c4 #c4 is the commit just older than the one you want to change
> ```
> 
> And git should open a text editor like this:
> 
> ```
> pick f1'
> pick f2'
> ```
> 
> Change it to this and save:
> 
> ```
> pick f1'
> squash f2' (if there's f3, f4, etc and you want them to all be squashed into f1', you'd change them to squash too)
> ```
> 
> Then we should end up with [2]:
> 
> ```
> c1
> c2
> c3
> c4 <- master
> f1'' <- feature-branch
> ```
> 
> At point [1]:
> 
> If we had done a merge instead of a rebase master, we would have gotten:
> 
> ```
> c1
> c2
> c3
> f1
> f2
> c4
> m1 <- feature-branch #This is a merge commit for c4 + f2 
> ```
> 
> We don't want to do this ^. Let's call this [3]. Compare [2] and [3]. [2] has several advantages over [3]:
> 
> A. Much easier and correct to review, because it's applied over the current master
> B. Because it's applied over the current master, the Git UI can merge it cleanly. Any merge conflict you had to handle during the rebase master operation would have had to be handled here instead anyway, so there's actually no extra work.
> C. The resulting history after merge is clearer, so we can look up where changes were made and why they were made (which goes back to clean commit messages which states the reason for the change and summary of the change)

now seems like i did rebase and now its ok. no merge.
Screenshot 2020-05-29 at 11 00 06

hboon:
@vladi8556 definitely better :) It looks like your local master isn't updated yet. Notice there are 5 commits in the Commits tab (in this page). You can see that too if you last screenshot. Notice there are 5 commits and then the 6th commit is master. If you update your master locally and then rebase again, you should see only 1 commit both in that screenshot as well as Commits tab (in this page, after you force push).

(so the way you just rebased is correct, just need to make sure master is updated).

hboon:
@vladyslav-iosdev  Remember to buzz me for review if done

vladyslav-iosdev:
> @vladyslav-iosdev Remember to buzz me for review if done

@hboon You can check it

PR #1825, opened 2020-05-22
Make Wallet tab table background white when there are not enough tokens to fill the screen #1818

### LINKED ISSUE #1818: Make Wallet tab table background white when there are not enough tokens to fill the screen
Screenshot 2020-05-21 at 4 39 24 PM

vladyslav-iosdev:
Closes #1818

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
@@ -221,6 +221,7 @@ class TokensViewController: UIViewController {
 
     private func reloadTableData() {
         tableView.reloadData()
+        updateTableViewBackground()
     }
 
     private func reload() {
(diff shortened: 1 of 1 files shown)


hboon:
@vladyslav-iosdev is this ready for review?

vladyslav-iosdev:
> @vladyslav-iosdev is this ready for review?

yes

hboon:
@vladyslav-iosdev I've forced push after rebasing master.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -368,6 +368,7 @@ extension TokensViewController: UITableViewDataSource {
     }
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        updateTableViewBackground()
```
hboon, line 371: - [ ] It works, but seems to be unsafe because modifying the `UITableView` instance (although it's just the view) during one of the data source callbacks might become dangerous in the future.

There's now a `TokensViewController.reloadTableData()`. Maybe we can move the call into it?

```
private func reloadTableData() {
    tableView.reloadData()
    updateTableViewBackground()
}
```

It's possible we forget to call `reloadTableData()` and call `tableView.reloadData()` directly, but at least it's just a cosmetic bug.

vladyslav-iosdev: This done.

PR #1840, opened 2020-05-28
add an extra gwei to the gas estimate to prevent hanging transactions

### LINKED ISSUE #1833: Gas price +1
Can we be a bit more aggressive with the gas price? If the node returns with a gas price n, then we should use n+1 as the gas price.

James-Sangalli:
Closes #1833

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -124,11 +124,12 @@ class TransactionConfigurator {
                     switch result {
                     case .success(let balance):
                         if let gasPrice = BigInt(balance.drop0x, radix: 16) {
-                            if gasPrice > GasPriceConfiguration.maxPrice {
+                            if (gasPrice + GasPriceConfiguration.oneGwei) > GasPriceConfiguration.maxPrice {
                                 // Guard against really high prices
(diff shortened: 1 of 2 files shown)


AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -128,7 +128,8 @@ class TransactionConfigurator {
                                 // Guard against really high prices
                                 seal.fulfill(GasPriceConfiguration.maxPrice)
                             } else {
-                                seal.fulfill(gasPrice)
+                                //Add an extra gwei because the estimate is sometimes too low
+                                seal.fulfill(gasPrice + BigInt(EthereumUnit.gwei.rawValue))
```
hboon, line 132: Good to make a constant in `GasPriceConfiguration.oneGwei = EthereumUnit.gwei.rawValue`

Also need to check that it isn't bigger than the max after + 1

James-Sangalli: Hmm? I thought that this is implicit given that gwei is like eth. You would assume one by default no?

hboon: Took a bit more effort to parse and verify is correct because I read `.rawValue` as `1_000_000_000` rather than 1 gwei.

James-Sangalli:
@hboon ready

PR #1843, opened 2020-05-28
use label not name in implicit props

James-Sangalli:
@hboon forgot about this.


AlphaWallet/TokenScriptClient/Models/AssetImplicitAttributes.swift
```diff
@@ -3,16 +3,16 @@
 import Foundation
 
 enum AssetImplicitAttributes: String, CaseIterable {
-    case name
```
hboon, line 6: This shouldn't be changed. It's the "name" of a token.

James-Sangalli: @hboon sussed

hboon:
@James-Sangalli I forgot to check if the test suite needs to be updated :P

PR #1851, opened 2020-05-29
Delete the redundant comment: "//We should refresh balance of the ETH every 10 seconds." #1847

### LINKED ISSUE #1847: Delete the redundant comment: "//We should refresh balance of the ETH every 10 seconds."

vladyslav-iosdev:
Closes #1847


hboon:
@vladyslav-iosdev Thanks, I did a simple amendment and shortened the commit message. It's redundant since it's reflected in the code.

PR #1864, opened 2020-06-01
Analytics

### LINKED ISSUE #1865: Implement analytics

hboon:
Closes #1865

* Enable once when we build with an updated `Constants+Credentials.swift` with the analytics key
* It doesn't log any events at the moment, though the functions to log events are there
* It logs user sessions data, so this is still useful.

In a future PR, once we decide which events to log, we can add them


hboon:
- [ ] Need to fix tests

PR #1832, opened 2020-05-26
Update shared ButtonsBar look #1819

vladyslav-iosdev:
Closes #1819

PR #1875, opened 2020-06-03
Update shared ButtonsBar look #1819

### LINKED ISSUE #1819: Update shared ButtonsBar look
The `ButtonsBar` class is used across the app. The buttons are currently all green.

- [ ] Let's make them appear in 3 configurations:

A\. 1 green, 1 white, and a "...":

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/dashboard?seid=5e4130b758245792d0d24970

B\. All green, no "..." (there can be more than 1 button):

C\. All white, no "..." (there can be more than 1 button):

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5dfc4ba8e901f31623b867d4

5a and 5b below are the only cases where we use configuration A (correction hboon 20200610: `TokenViewController` uses configurationA too). There is one use of C in `CreateInitialWalletViewController` (look for ButtonsBar.configureSecondary(), this is shown for new users without wallets), every other instance of `ButtonsBar` use B.

- [ ] Assume that each button's style (green or white) can be configured separately (in the future, we might mix green and white button in A and not always 1st green, 2nd white, or B might not always be green)

- [ ] In the 2 view controllers mentioned below (TokensCardViewController and TokenInstanceViewController), noticed that there is an additional separator above the buttons (so only with configuration A). Either add the separator within the view controllers or `ButtonsBar`

How to go to screens which uses TokenScript
---
Don't worry about what is TokenScript for now.

1\. Enable the Kovan network

2\. Add the contract 0x2B58A9403396463404c2e397DBF37c5EcCAb43e5:

3\. Tap the token:

Screenshot 2020-05-22 at 11 05 39 AM

4a\. (We are now in TokensCardViewController) Tap one of the rows with the picture:

5a\. The buttons at the bottom (in TokenInstanceViewController):

4b\. (after 3) Tap and hold one of the rows with the picture

5b\. The buttons at the bottom (in TokensCardViewController):

supersedes #1702

vladyslav-iosdev:
Closes #1819

Supersedes #1832


vladyslav-iosdev:
@hboon You can check.

AlphaWallet/Tokens/ViewControllers/TokenInstanceViewController.swift
```diff
@@ -121,21 +121,22 @@ class TokenInstanceViewController: UIViewController, TokenVerifiableStatusViewCo
...
+            let action = viewModel.actions[index]
             button.setTitle(action.name, for: .normal)
             button.addTarget(self, action: #selector(actionButtonTapped), for: .touchUpInside)
             switch account.type {
             case .real:
                 if let selection = action.activeExcludingSelection(selectedTokenHolders: [tokenHolder], forWalletAddress: account.address) {
                     if selection.denial == nil {
                         button.isHidden = true
-                    } else {
+                    } else if button.isHidden != true {
```
hboon, line 136: This else-if seems redundant:

```
                    } else if button.isHidden != true {
                         button.isHidden = false
```

Is it needed?

hboon:
@vladyslav-iosdev can you also rebase master and verify that the change introduced in #1881 is still incorporated?

vladyslav-iosdev:
> can you also rebase master and verify that the change introduced in #1881 is still incorporated?

Done

AlphaWallet/Tokens/ViewControllers/TokenInstanceViewController.swift
hboon: @vladyslav-iosdev this doesn't seem to be fix yet?

hboon:
@vladyslav-iosdev I have forced pushed after `git rebase master` (could you have not pull the lasted into your local master before running `git rebase master`?)

AlphaWallet/Tokens/ViewControllers/TokenInstanceViewController.swift
vladyslav-iosdev: i removed it, but seems like during rebase it returned back. i will fix it

vladyslav-iosdev:
> @vladyslav-iosdev I have forced pushed after `git rebase master` (could you have not pull the lasted into your local master before running `git rebase master`?)

I did rebase, removed unnecessary lines, checked ERC20. Should be ok.

hboon:
@vladyslav-iosdev I see this:



Maybe I messed it up with the force push (to commit 6ce266f4c2bddb4eda246f4b0550ee5b3ab054ae). Can you verify if this is the commit you have?

vladyslav-iosdev:
> @vladyslav-iosdev I see this:
> 
> 
> 
> Maybe I messed it up with the force push (to commit [6ce266f](https://github.com/AlphaWallet/alpha-wallet-ios/commit/6ce266f4c2bddb4eda246f4b0550ee5b3ab054ae)). Can you verify if this is the commit you have?

@hboon i don't have this commit, my last commit `59f01496`

hboon:
@vladyslav-iosdev that’s the one I forced pushed.

vladyslav-iosdev:
> @vladyslav-iosdev that’s the one I forced pushed.

@hboon  now should be ok, i pushed in wrong branch

vladyslav-iosdev:
> I've forced push after `git rebase master`, but took this screenshot before the rebase. It's showing 3 green buttons together for xDai:
> 
> 

@hboon  Does it wrong behaviour? How it should be?

hboon:
@vladyslav-iosdev something like in https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-641139358, with the 3rd (and additional buttons) hidden in the "more" button.

vladyslav-iosdev:
> @vladyslav-iosdev something like in [#1875 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-641139358), with the 3rd (and additional buttons) hidden in the "more" button.

seems like everything is ok, 

TokenViewController displays `N` green buttons, like before (configuration C)
CreateInitialWalletViewController displays configuration B
TokensCardViewController and TokenInstanceViewController display configuration A
Or TokenViewController also should use A configuration

vladyslav-iosdev:
> I'm sorry. I missed out `TokenViewController` when I wrote the issue. Can you make it use configuration A too? You should be able to make it show 3 actions when you tap the xDai card. Also Sai (on mainnet), which should have ~5 actions.

@hboon there are cases when we have only 2 options, what should be i this case, do i need to show more button?

hboon:
> @hboon there are cases when we have only 2 options, what should be i this case, do i need to show more button?

If there are only 2 buttons, for all configuration C, we should just show those 2 buttons, and hide the "More" button.

vladyslav-iosdev:
> > @hboon there are cases when we have only 2 options, what should be i this case, do i need to show more button?
> 
> If there are only 2 buttons, for all configuration C, we should just show those 2 buttons, and hide the "More" button.

@hboon done

hboon:
> > @hboon there are cases when we have only 2 options, what should be i this case, do i need to show more button?
> 
> If there are only 2 buttons, for all configuration C, we should just show those 2 buttons, and hide the "More" button.

@vladyslav-iosdev I'm sorry. When I said "configuration C", I  really meant "configuration A".

> A. 1 green, 1 white, and a "...":

Here's how configuration A should be for various button counts:

1: green
2: green, white (that means send ETH should be 1 green Send button and 1 white Receive button)
3 or more: green, white, "..."

But yes, you are right that `TokenViewController` should use configuration A.

hboon:
I think it's handling the actions being changed dynamically. You can see this in SAI/DAI:

When I tap DAI in Wallets tab, I see this briefly (good):



And then about a second or two later, the buttons change (still good) because some processing was done which took some time so "Enable Migration" is no longer available anymore (still good), but noticed that it is 1 green button and 1 "More" button, it should be 1 green, 1 white and 1 more button:

hboon:
I also noticed that when `TokenCardsViewController` and `TokenInstanceViewController` show 1 action button, it is white. It should be 1 green button (because this is configuration A like described in https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-642466589).




You can reproduce this with Cryptokitty (0x06012c8cf97bead5deae237070f9587f8e7a266d) on mainnet.

vladyslav-iosdev:
> I also noticed that when `TokenCardsViewController` and `TokenInstanceViewController` show 1 action button, it is white. It should be 1 green button (because this is configuration A like described in [#1875 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-642466589)).
> 
> 
> 
> 
> 
> You can reproduce this with Cryptokitty (0x06012c8cf97bead5deae237070f9587f8e7a266d) on mainnet.

@hboon see it. thx. I find where issue was

James-Sangalli:
photo_2020-06-12_11-41-04

@hboon @vladyslav-iosdev another thing I noticed is that if the action included in the two buttons is later filtered out by a selection it will only show the one button.

Maybe it would be more wise to emulate android and simply not show the buttons until the selection is resolved

hboon:
> Maybe it would be more wise to emulate android and simply not show the buttons until the selection is resolved

Yes, a separate issue though. Because the TokenScript file might still change while this screen is active so the actions can change even if we don't show the buttons until the selection is resolved.

@vladyslav-iosdev James is referring to this problem: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-642535949. Can you fix it too?

vladyslav-iosdev:
> > Maybe it would be more wise to emulate android and simply not show the buttons until the selection is resolved
> 
> Yes, a separate issue though. Because the TokenScript file might still change while this screen is active so the actions can change even if we don't show the buttons until the selection is resolved.
> 
> @vladyslav-iosdev James is referring to this problem: [#1875 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-642535949). Can you fix it too?

@hboon i updated `ButtonsBar` with ability to dynamically handle displaying state. Now  seems like it works. tested with `cDai`.

vladyslav-iosdev:
> I think it's handling the actions being changed dynamically. You can see this in SAI/DAI:
> 
> When I tap DAI in Wallets tab, I see this briefly (good):
> 
> 
> 
> And then about a second or two later, the buttons change (still good) because some processing was done which took some time so "Enable Migration" is no longer available anymore (still good), but noticed that it is 1 green button and 1 "More" button, it should be 1 green, 1 white and 1 more button:
> 
> 

@hboon i cant reproduce this with latest  commit

vladyslav-iosdev:
> I also noticed that when `TokenCardsViewController` and `TokenInstanceViewController` show 1 action button, it is white. It should be 1 green button (because this is configuration A like described in [#1875 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/1875#issuecomment-642466589)).
> 
> 
> 
> 
> 
> You can reproduce this with Cryptokitty (0x06012c8cf97bead5deae237070f9587f8e7a266d) on mainnet.

Also cant reproduce this 
Screenshot 2020-06-12 at 13 23 12

hboon:
@vladyslav-iosdev your current commit works already, I think. Halfway through checking it.

hboon:
@James-Sangalli it's in!

PR #1876, opened 2020-06-03
Display an error message when QR code for EIP681 link has invalid contract #1874

### LINKED ISSUE #1874: Display an error message when QR code for EIP681 link has invalid contract
After #1862 is merged

This is in `SendViewController`.

Show "Invalid token" with `UIAlertController`.

Generate a link at https://brunobar79.github.io/eip681-link-generator/# and use the QR code

---

This should populate the fields and not show an error message:

contract: 0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359 (this should show up as the SAI token)
receiver: any Ethereum address for wallet
amount: 1000000000000000000

---

This should show the error message and not populate the textfields:

contract: any Ethereum address that is not a token (eg. a wallet address)
receiver: any Ethereum address for wallet
amount: 1000000000000000000

vladyslav-iosdev:
Closes #1874

### DIFF
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -47,7 +47,9 @@ class SendViewController: UIViewController, CanScanQRCode {
     private let recipientViewModel = SendViewSectionHeaderViewModel(
         text: R.string.localizable.sendRecipient().uppercased()
     )
-//    private let initialTransferType: TransferType
+    //We use weak link to make sure that token alert will be deallocated by close button tapping.
+    //We storing link to make shure that only one alert is displaying on the screen.
+    private weak var invalidTokenAlert: UIViewController?
     let targetAddressTextField = AddressTextField()
     lazy var amountTextField = AmountTextField(server: session.server)
     weak var delegate: SendViewControllerDelegate?
@@ -71,7 +73,6 @@ class SendViewController: UIViewController, CanScanQRCode {
         self.storage = storage
         self.ethPrice = cryptoPrice
         self.assetDefinitionStore = assetDefinitionStore
-//        self.initialTransferType = transferType
         self.viewModel = .init(transferType: transferType, session: session, storage: storage)
 
         super.init(nibName: nil, bundle: nil)
@@ -350,7 +351,7 @@ class SendViewController: UIViewController, CanScanQRCode {
         case .nativeCryptocurrency(_, let recipient, let amount):
             currentSubscribableKeyForNativeCryptoCurrencyBalance = session.balanceViewModel.subscribe { [weak self] viewModel in
                 guard let celf = self else { return }
-                guard let tokenObject = celf.storage.token(forContract: celf.viewModel.transferType.contract) else { return }
+                guard celf.storage.token(forContract: celf.viewModel.transferType.contract) != nil else { return }
                 celf.configureFor(contract: celf.viewModel.transferType.contract, recipient: recipient, amount: amount, shouldConfigureBalance: false)
             }
             session.refresh(.ethBalance)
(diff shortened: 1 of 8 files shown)


AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -357,16 +357,24 @@ class SendViewController: UIViewController, CanScanQRCode {
...
-
-        guard let result = QRCodeValueParser.from(string: result) else { return }
-        switch result {
-        case .address(let recipient):
-            guard let tokenObject = storage.token(forContract: viewModel.transferType.contract) else { return }
-            let amountAsIntWithDecimals = EtherNumberFormatter.full.number(from: amountTextField.ethCost, decimals: tokenObject.decimals)
-            configureFor(contract: transferType.contract, recipient: .address(recipient), amount: amountAsIntWithDecimals)
-        case .eip681(let protocolName, let address, let functionName, let params):
-            checkAndFillEIP681Details(protocolName: protocolName, address: address, functionName: functionName, params: params)
+        if let result = QRCodeValueParser.from(string: result) {
```
hboon, line 360: I don't think we should check the error here. So we should still guard and return like before the PR.

It's actually when fetching the token's details (symbol, decimals etc) where it fails if the contract is not a token. The error here could for example be if the EIP681 is not well-formed or maybe not even a URL.

It's the body of:

> self.delegate?.lookup(contract: contract, in: self) { data in

Around https://github.com/AlphaWallet/alpha-wallet-ios/pull/1876/files#diff-6aa3b98b96c470b61f3e704b8b9be6c8R396

Can you see if you can reproduce it?

vladyslav-iosdev: > I don't think we should check the error here. So we should still guard and return like before the PR.
> 
> It's actually when fetching the token's details (symbol, decimals etc) where it fails if the contract is not a token. The error here could for example be if the EIP681 is not well-formed or maybe not even a URL.
> 
> It's the body of:
> 
> > self.delegate?.lookup(contract: contract, in: self) { data in
> 
> Around https://github.com/AlphaWallet/alpha-wallet-ios/pull/1876/files#diff-6aa3b98b96c470b61f3e704b8b9be6c8R396
> 
> Can you see if you can reproduce it?

@hboon so you mean to show alert for all these cases
.name, .symbol, .balance, .decimals, .nonFungibleTokenComplete, .delegateTokenComplete, .failed:
in checkAndFillEIP681Details?

vladyslav-iosdev: @hboon can you confirm this?

hboon: > so you mean to show alert for all these cases .name, .symbol, .balance, .decimals, .nonFungibleTokenComplete, .delegateTokenComplete, .failed: in checkAndFillEIP681Details?

Not for all of them.

Those like `.name`, etc are partial results. Ignore those

I think just:

* `.nonFungibleTokenComplete`
* `.delegateTokenComplete`
* but not `.failed`.

This should help to test:

1. Comment out this line in `SendViewController` (around line 395) and replace with `if false` to force it go to `else`:

```
if self.storage.token(forContract: contract) != nil {
```

Then:

A. Generate a QR code for a EIP681 link which should show the alert we want to show with this PR:

contract: 0x007bEe82BDd9e866b2bd114780a47f2261C684E3 (this is a wallet, so we expect it to fail)
receiver: 0x007bEe82BDd9e866b2bd114780a47f2261C684E3
amount: 1

B. Generate a QR code for a EIP681 link which should work and not show errors:

contract: 0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359 (this is the SAI token, on mainnet)
receiver: 0x007bEe82BDd9e866b2bd114780a47f2261C684E3
amount: 3000000000000000000

C. Generate a QR code for EIP681 transfer (click "ETH payment" in the page instead of ERC20) and make sure it doesn't trigger this error

vladyslav-iosdev: @hboon 
i walked through these cases, A and C works as you mentioned, but B doesnt work, (maybe i did something wrong)  all the time it enters to delegateTokenComplete case

hboon: > i walked through these cases, A and C works as you mentioned, but B doesnt work, (maybe i did something wrong) all the time it enters to delegateTokenComplete case

@vladyslav-iosdev B works for me. Did you try on mainnet?

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -47,6 +47,7 @@ class SendViewController: UIViewController, CanScanQRCode {
     private let recipientViewModel = SendViewSectionHeaderViewModel(
         text: R.string.localizable.sendRecipient().uppercased()
     )
+    private weak var invalidTokenAlert: UIViewController?
```
hboon, line 50: Can you add a comment above this line to explain that it is relying on the `weak var` being nil-ed when the alert closes, to make sure that at most 1 instance of the alert is shown at any time?

vladyslav-iosdev: Done

hboon:
@James-Sangalli just noticed you were the one who raised this. See if you want to take a look. Otherwise, we can merge.

James-Sangalli:
@vladyslav-iosdev please use the repo directly from now on instead of the fork, cheers.

PR #1891, opened 2020-06-05
Show "Available: <eth/token balance>" in send screen #1886

### LINKED ISSUE #1886: Show "Available: <eth/token balance>" in send screen
https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5dfc4ba7aa48e01aa20e886e

Also highlights this "Available: xxx" as well as the amount entered in red when the amount entered is more than the balance like in:

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5e2ac09836436754bc975a1f

This is probably going to make `SendInputErrors.wrongInput` redundant.

- [ ] If `SendInputErrors.wrongInput` is still used after this PR, point that out so we can improve the error message

edit hboon: `SendInputErrors.wrongInput` is probably still around, but only if the `input = 0`

vladyslav-iosdev:
Closes #1886

### DIFF
--- a/AlphaWallet/Transfer/ViewModels/SendViewModel.swift
+++ b/AlphaWallet/Transfer/ViewModels/SendViewModel.swift
@@ -2,6 +2,7 @@
 
 import Foundation
 import UIKit
+import BigInt
 
 struct SendViewModel {
     private let session: WalletSession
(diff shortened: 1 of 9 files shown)


AlphaWallet/Transfer/ViewModels/SendViewModel.swift
```diff
@@ -67,4 +67,35 @@ struct SendViewModel {
     var recipientsAddress: String {
         return R.string.localizable.sendRecipientsAddress()
     }
+
+    var availableLabelText: String? {
+        switch transferType {
+        case .nativeCryptocurrency:
+            if let balance = session.balance {
+                return R.string.localizable.sendAvailable("\(balance.amountFull) \(transferType.symbol)")
```
hboon, line 75: There should be a colon like this: "Available: xxx"

AlphaWallet/Transfer/ViewModels/SendViewModel.swift
```diff
@@ -67,4 +67,35 @@ struct SendViewModel {
...
+            }
+        case .ERC20Token(let token, _, _):
+            let tokenBalance = storage.token(forContract: token.contractAddress)?.valueBigInt
+            if let tokenBalance = tokenBalance {
+                return R.string.localizable.sendAvailable("\(tokenBalance) \(transferType.symbol)")
```
hboon, line 80: There should be a colon like this: "Available: xxx"

vladyslav-iosdev:
> Looks great. Can you help to make these changes:
> 
> * [x]  Always hide the equivalent `alternativeAmountLabel` (just `isHidden = true` where it's created and leave a comment "//Intentionally not sure the equivalent amount for now"). We might need it for forks
> * [x]  Round the amount in "Available: 1.2345 ETH" to 4 decimal places. And apply `.roundingMode = .down` (Create a new formatter if you have to, the existing ones in `EtherNumberFormatter.swift`, if we are using it, will be updated across the app later)

Done, if i understand correctly.

vladyslav-iosdev:
> * [x]  Can you make the value display like "0.3000" instead of ".3000" when the whole number part is 0?:

this should be ok

vladyslav-iosdev:
> * I noticed that in some ERC20s, it shows a zero balance ("Available: .0000") even though it should be > 0. Maybe it's a printing/formatting because it only changes to red when I key in an amount greater than the known balance.

yep right, its because we applying NumberFormatter that parforms round down, and take only 4 after `.`. What do i need to do with this formatter?

vladyslav-iosdev:
> > yep right, its because we applying NumberFormatter that parforms round down, and take only 4 after `.`. What do i need to do with this formatter?
> 
> Seems to be ok now in the most recent commit. But I think it's not taking into account `decimals` of a token correctly. For example, for KICK, if I have a balance of `888,888`, it's showing "Available: 8.8888e-05" (I can't figure out quickly why). But DAI/SAi with a balance of `1.7` gets displayed correctly as `Available: 1.7`

@hboon how can i test it on my side? i can only watch thise `0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362` wallet, and can't get to `SendViewController`

hboon:
> i can only watch thise 0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362 wallet, and can't get to SendViewController

@vladyslav-iosdev change the code :) There's probably a check somewhere for `.real` and `.watch`. Force it to go to `.real`.

vladyslav-iosdev:
> > i can only watch thise 0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362 wallet, and can't get to SendViewController
> 
> @vladyslav-iosdev change the code :) There's probably a check somewhere for `.real` and `.watch`. Force it to go to `.real`.

@hboon now should be ok

PR #1892, opened 2020-06-05
Display ENS name (if available) and Ethereum address for current wallet under "Change / Add Wallet" in Settings screen #1887

### LINKED ISSUE #1887: Display ENS name (if available) and Ethereum address for current wallet under "Change / Add Wallet" in Settings screen
As in https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5dfc1f53e5325a98132d1d74

We can see how it's implemented in the screen shown after tapping on the cell. `AccountsViewController`

vladyslav-iosdev:
Closes #1887

### DIFF
--- a/AlphaWallet/Settings/ViewControllers/SettingsViewController.swift
+++ b/AlphaWallet/Settings/ViewControllers/SettingsViewController.swift
@@ -107,6 +107,28 @@ class SettingsViewController: UIViewController {
         }
     }
 
+    private func configureChangeWalletCellWithResolvedESN(_ row: SettingsWalletRow, cell: SettingTableViewCell) {
+        cell.configure(viewModel: .init(
+            titleText: row.title,
+            subTitleText: self.viewModel.addressReplacedWithESN(),
+            icon: row.icon)
+        )
+
+        let serverToResolveEns = RPCServer.main
+        let address = account.address
+
+        ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { [weak self] result in
+            guard let strongSelf = self else { return }
+
+            let viewModel: SettingTableViewCellViewModel = .init(
+                titleText: row.title,
+                subTitleText: strongSelf.viewModel.addressReplacedWithESN(result.value),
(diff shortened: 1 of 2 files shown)


AlphaWallet/Settings/ViewControllers/SettingsViewController.swift
```diff
@@ -107,6 +107,30 @@ class SettingsViewController: UIViewController {
...
+
+            let viewModel: SettingTableViewCellViewModel = .init(
+                titleText: row.title,
+                subTitleText: ensName,
+                icon: row.icon
+            )
+            cell.configure(viewModel: viewModel)
+        }
+
+        cell.configure(viewModel: viewModel)
```
hboon, line 131: The second `cell.configure(viewModel: viewModel)` has to be moved up, to before calling `getENSNameFromResolver()` because the latter can return a cached result immediately. You can see the problem by doing this:

1. Switch to Settings tab
2. Wait for ENS reverse resolution to complete and ENS name shown
3. Tap "Change / Add Wallet"
4. Tap "Done".
5. Observe: it shows the address and wouldn't show ENS name even after waiting for a while. Expected: Show ENS name (and right away because of caching).

AlphaWallet/Settings/ViewControllers/SettingsViewController.swift
```diff
@@ -107,6 +107,30 @@ class SettingsViewController: UIViewController {
...
+
+        let serverToResolveEns = RPCServer.main
+        let address = account.address
+
+        ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { result in
+            guard let ensName = result.value else { return }
+
+            let viewModel: SettingTableViewCellViewModel = .init(
+                titleText: row.title,
+                subTitleText: ensName,
```
hboon, line 125: The `subTitleText` should be include the Ethereum address. Refer to `AccountViewModel.addresses` to see how it should be displayed.

hboon:
Remember to rebase master since it has changed since.

PR #1893, opened 2020-06-05
Tweak Next and Done buttons for textboxes in Send eth/ERC20 screen #1890

### LINKED ISSUE #1890: Tweak Next and Done buttons for textboxes in Send eth/ERC20 screen
`AmountTextFieldDelegate` should:

- [x] have a function like `AddressTextFieldDelegate.shouldReturn(in:)`, maybe called `didTapInputAccessoryButton`
- [x] add property similar to `AddressTextField.returnKeyType` (called `inputAccessoryButtonTitle`. Maybe create a new enum for it with cases: `.done` and `.next`)

and SendViewController should use them so that:

- [x] the amount field's `inputAccessoryView` (look for `inputAccessoryView` in `AmountTextField.swift`) shows Next instead of Done (careful where `AmountTextField` is used elsewhere. They should still have the `.done` and hide keyboard behaviour
- [x] tap on Next for amount should go to the address textfield

Also in SendViewController:

- [x] the address field's keyboard shows Done instead of Next

vladyslav-iosdev:
Closes #1890


hboon:
Forced pushed after `git rebase master`

PR #1898, opened 2020-06-08
Tweak how result of resolving ENS name (-> address) in Send screen is displayed

hboon:
Before PR|After PR
-|-
|

### DIFF
--- a/AlphaWallet/Style/AppStyle.swift
+++ b/AlphaWallet/Style/AppStyle.swift
@@ -164,7 +164,7 @@ enum DataEntry {
 
     enum Font {
         static let text = Fonts.regular(size: ScreenChecker().isNarrowScreen ? 14: 18)
-        static let label = Fonts.regular(size: 10)
+        static let label = Fonts.bold(size: 13)
         static let textFieldTitle = Fonts.regular(size: 13)
         static let textFieldStatus = Fonts.bold(size: 13)
         static let textField = Fonts.regular(size: ScreenChecker().isNarrowScreen ? 14: 17)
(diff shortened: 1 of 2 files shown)


tomekalphawallet:
> Personally don't think this is an improvement, what do you think @tomekalphawallet?

Hmmm... why?

In the future, we will have to find a space for "Address Book"
When validating, you usually look at the last 4 characters. 

Screenshot 2020-06-23 10 59 18

James-Sangalli:
> > Personally don't think this is an improvement, what do you think @tomekalphawallet?
> 
> Hmmm... why?
> 
> In the future, we will have to find a space for "Address Book"
> When validating, you usually look at the last 4 characters.
> 
> Screenshot 2020-06-23 10 59 18

@tomekalphawallet from my perspective I find it unsettling not to see the whole address... I may have a bias there however

hboon:
@James-Sangalli I'll rebase and push again

hboon:
> rebase

@James-Sangalli done. Would you do a quick check again because `master` had changed quite a bit.

tomekalphawallet:
> > > Personally don't think this is an improvement, what do you think @tomekalphawallet?
> > 
> > 
> > Hmmm... why?
> > In the future, we will have to find a space for "Address Book"
> > When validating, you usually look at the last 4 characters.
> > Screenshot 2020-06-23 10 59 18
> 
> @tomekalphawallet from my perspective I find it unsettling not to see the whole address... I may have a bias there however

If that would be a case, maybe on tap you can view the whole address. 

Screenshot 2020-06-24 10 33 19

PR #1902, opened 2020-06-09
remove issuer label

James-Sangalli:
https://github.com/AlphaWallet/alpha-wallet-android/issues/1453

### DIFF
--- a/AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
+++ b/AlphaWallet/TokenScriptClient/Models/XMLHandler.swift
@@ -235,15 +235,6 @@ private class PrivateXMLHandler {
         return "en"
     }
 
-    //TODO may need to support action-only. Is it the same? Do along with signature verification
-    var issuer: String {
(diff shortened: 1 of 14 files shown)


hboon:
@James-Sangalli but the problem on Android is that the issuer is always shown as "AlphaWallet", right? Do we want to remove it entirely?

James-Sangalli:
> @James-Sangalli but the problem on Android is that the issuer is always shown as "AlphaWallet", right? Do we want to remove it entirely?

Nope, it works correctly. It displays AlphaWallet because we signed it with that certificate and therefore it says that. Same in iOS

hboon:
@James-Sangalli got it.

James-Sangalli:
@hboon we should probably also remove the "ethereum blockchain" text since it's pretty redundant (but keep the spacing), keen?

hboon:
> @hboon we should probably also remove the "ethereum blockchain" text since it's pretty redundant (but keep the spacing), keen?

@James-Sangalli going in here: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1902#issuecomment-641034194

PR #1907, opened 2020-06-09
Update Wallet tab cell's layout #1903

### LINKED ISSUE #1903: Update Wallet tab cell's layout
Part of #1665 

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/5dfaf2ce114ed41d029b3e2b

* The native crypto cells like Eth and xDai still has current fiat value, the appreciation/depreciation and fiat value appreciation
* Don't show Issuer because it's removed in #1902
* Also the Add / Hide screen

Out of scope
---
Don't add the token icon

vladyslav-iosdev:
Closes #1903

### DIFF
--- a/AlphaWallet/Core/Formatters/CurrencyFormatter.swift
+++ b/AlphaWallet/Core/Formatters/CurrencyFormatter.swift
@@ -11,6 +11,20 @@ class CurrencyFormatter {
         //TODO support multiple currency values
         formatter.currencyCode = Currency.USD.rawValue
         formatter.numberStyle = .currency
+
+        return formatter
+    }
+
+    static var usdFormatter: NumberFormatter {
+        let formatter = NumberFormatter()
+        formatter.minimumFractionDigits = 2
+        formatter.maximumFractionDigits = 2
+        formatter.roundingMode = .down
+        //TODO support multiple currency values
+        formatter.numberStyle = .currency
+        formatter.positiveFormat = "0.00" + " " + Constants.Currency.usd
+        formatter.negativeFormat = "-0.00" + " " + Constants.Currency.usd
+
         return formatter
     }
 
--- a/AlphaWallet/Settings/Types/Constants.swift
+++ b/AlphaWallet/Settings/Types/Constants.swift
@@ -25,6 +25,9 @@ public struct Constants {
     public static let artisTau1MagicLinkHost = "artis_tau1.aw.app"
     public static let customMagicLinkHost = "custom.aw.app"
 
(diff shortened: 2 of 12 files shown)


vladyslav-iosdev:
> `EthTokenViewCell` should be like this instead:
> 
> A. .mainnet and .xdai show "xxx USD" on the left, "yyy USD (1.23 %) on the right
> B. everything else: show "xxx Blockchain"
> 
> I think the current state of the PR looks like:
> 
> * A is good
> * but B is showing "" on the left and "- -" on the right.

@hboon as i undastand correctly i need to check server, right?
what if i will appy next
```
case .nativeCryptocurrency where token.server == .main || token.server == .xDai:
            let cell = tableView.dequeueReusableCell(withIdentifier: EthTokenViewCell.identifier, for: indexPath) as! EthTokenViewCell
```
and  for rest of `nativeCryptocurrency` will dequeue `FungibleTokenViewCell`

hboon:
I think it's better to keep using `EthTokenViewCell` for all `.nativeCryptoCurrency` (ETH/xDai as well as the other networks/servers which do not have a fiat value). It's semantically more correct since FungibleTokenViewCell are meant for ERC20s ("fungible" is not the best name of course) vs EthTokenViewCell which is for native crypto. This makes it possible in the future if we want to decorate native crypto (eth/xDai etc) differently from other tokens.

We should be able to tell if fiat value is available by looking at `TokensViewModel.ticker(for:)`. i.e. if there is no ticker (price), then we can't calculate the fiat value, appreciation, etc. And we can display "xxx Blockchain" and leave the RHS empty. The generation for "- -" probably also gives a clue how to check for it?

So we don't have to check for the server explicitly.

vladyslav-iosdev:
> I think it's better to keep using `EthTokenViewCell` for all `.nativeCryptoCurrency` (ETH/xDai as well as the other networks/servers which do not have a fiat value). It's semantically more correct since FungibleTokenViewCell are meant for ERC20s ("fungible" is not the best name of course) vs EthTokenViewCell which is for native crypto. This makes it possible in the future if we want to decorate native crypto (eth/xDai etc) differently from other tokens.
> 
> We should be able to tell if fiat value is available by looking at `TokensViewModel.ticker(for:)`. i.e. if there is no ticker (price), then we can't calculate the fiat value, appreciation, etc. And we can display "xxx Blockchain" and leave the RHS empty. The generation for "- -" probably also gives a clue how to check for it?
> 
> So we don't have to check for the server explicitly.

@hboon Changed!

PR #1916, opened 2020-06-11
Get rid of or replace unnecessary `default: ` in switches across the code base #1909

### LINKED ISSUE #1909: Get rid of or replace unnecessary `default: ` in switches across the code base
There should be some that are necessary, but we should be able to get rid of quite a number of them.

Similar to #1905

vladyslav-iosdev:
Closes #1909


AlphaWallet/Market/Coordinators/UniversalLinkCoordinator.swift
```diff
@@ -416,7 +416,7 @@ class UniversalLinkCoordinator: Coordinator {
         switch server {
         case .xDai:
             errorMessage = R.string.localizable.aClaimTokenFailedNotEnoughXDAITitle()
-        default:
```
hboon, line 419: Line 69 too

PR #1919, opened 2020-06-11
Typing/pasting an Ethereum address into the address field in send ERC20/ETH screen should also do reverse ENS lookup #1914

### LINKED ISSUE #1914: Typing/pasting an Ethereum address into the address field in send ERC20/ETH screen should also do reverse ENS lookup
Display it in the label below the address textbox (the label that shows the address when an ENS name is entered)

Relevant:

* SendViewController
* ENSReverseLookupCoordinator

Should be careful not to have unnecessary look ups:

* Don't try when the address length is not complete

vladyslav-iosdev:
Closes #1914

### DIFF
--- a/AlphaWallet/UI/AddressTextField.swift
+++ b/AlphaWallet/UI/AddressTextField.swift
@@ -15,6 +15,33 @@ class AddressTextField: UIControl {
     private let textField = UITextField()
     //Always resolve on mainnet
     private let serverToResolveEns = RPCServer.main
+    private var addressString: String? {
+        switch addressOrEnsName {
+        case .address(let value):
+            return value.eip55String
+        case .ensName, .none:
+            return nil
+        }
+    }
+
+    private var addressOrEnsName: AddressOrEnsName? {
+        didSet {
+            ensAddressLabel.text = addressOrEnsName?.stringValue
+            ensAddressLabel.isHidden = ensAddressLabel.text == nil
+        }
+    }
+
+    private var textFieldText: String {
+        get {
+            return textField.text ?? ""
+        }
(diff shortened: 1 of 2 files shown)


AlphaWallet/UI/AddressTextField.swift
```diff
@@ -293,12 +343,27 @@ class AddressTextField: UIControl {
...
                 guard oldTextValue == strongSelf.textField.text?.trimmed else { return }
-                strongSelf.ensAddressLabel.isHidden = false
-                strongSelf.ensAddressLabel.text = address.address
+
+                strongSelf.resolvedState = .address(value: address.address)
             }
         }
     }
 
+    private func queueEnsResolution(ofValue addressString: String) {
```
hboon, line 352: Can't build without making this non-private.

But is it better if we just keep it private and call it internally, in `AddressTextField.value`'s setter? Otherwise we have to call `queueEnsResolution(ofValue:)` in other places where `AddressTextField` is used, not just `SendViewController`

vladyslav-iosdev: > There's a problem when I perform this sequence in the Send ETH screen:
> 
> 1. Type ENS name
> 2. Address resolves and gets displayed
> 3. Tap clear
> 4. Tap paste (a valid address, the same address as the one that appeared in step 2, if it matters)
> 5. The address doesn't appear in the textbox (it's still empty), but the ENS name appears below in green

@hboon fixed

vladyslav-iosdev:
> I mean, `AddressOrEnsName.swift` already exists :)

@hboon  Done

PR #1922, opened 2020-06-12
Background color revealed in Wallet and Transactions tab when pulling down to refresh is wrong #1912

### LINKED ISSUE #1912: Background color revealed in Wallet and Transactions tab when pulling down to refresh is wrong
Reproducible this on iPhone X 13.3 device. But not on iPhone X 12.4 simulator and 13.3 simulator

The annotated section should be white

vladyslav-iosdev:
Closes #1912


vladyslav-iosdev:
> Looks good. Can you do the same thing for the Transactions tab too?

@hboon  Done.

vladyslav-iosdev:
> Looks good, but can you split out the changes to the Settings tabs to a different PR? It's for a different purpose. In that PR, can you also check if "Support" and other similar screens opened from the Settings tab works the same way?

ok, np.

PR #1926, opened 2020-06-15
use node gas price on dapp transactions

### LINKED ISSUE #1908: Use the gas price provided by wallet instead of DApp
For example, 
if I make an ETH transfer now, the suggested gas price in the transaction page is 30 Gwei
If I use app.ens.domains or uniswap.exchange, the suggested gas price in the transaction page is 9 Gwei, which is a fixed number not changing with the market gas price

This will cause a big UX issue for non-technical users.

James-Sangalli:
Closes #1908

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -41,7 +41,7 @@ class TransactionConfigurator {
         }
     }
 
-    private var requestEstimateGas: Bool {
+    private var gasLimitNotSet: Bool {
         return transaction.gasLimit == .none
     }
 
@@ -112,6 +112,17 @@ class TransactionConfigurator {
         }
     }
 
+    func estimateGasPrice() {
+        TransactionConfigurator.estimateGasPrice(server: self.session.server).done { [weak self] gasPrice in
+            guard let strongSelf = self else { return }
+            strongSelf.configuration = TransactionConfiguration(
+                    gasPrice: gasPrice,
+                    gasLimit: strongSelf.transaction.gasLimit ?? GasLimitConfiguration.maxGasLimit,
+                    data: strongSelf.configuration.data
+            )
+        }
+    }
+
     // Generic function to derive the typical acceptable gas price on each network
     static public func estimateGasPrice(server: RPCServer) -> Promise<BigInt> {
         return Promise { seal in
@@ -144,15 +155,22 @@ class TransactionConfigurator {
 
 // swiftlint:disable function_body_length
     func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
-        switch transaction.transferType {
-        case .nativeCryptocurrency, .dapp:
-            guard requestEstimateGas else {
-                return completion(.success(()))
-            }
+        if gasLimitNotSet {
             estimateGasLimit()
+        }
+        switch transaction.transferType {
+        case .dapp:
+            estimateGasPrice()
             configuration = TransactionConfiguration(
                     gasPrice: calculatedGasPrice,
                     gasLimit: GasLimitConfiguration.maxGasLimit,
+                    data: transaction.data ?? configuration.data
+            )
+            completion(.success(()))
+        case .nativeCryptocurrency:
+            configuration = TransactionConfiguration(
+                    gasPrice: calculatedGasPrice,
+                    gasLimit: GasLimitConfiguration.minGasLimit,
                     data: transaction.data ?? configuration.data
             )
             completion(.success(()))


AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -146,6 +157,7 @@ class TransactionConfigurator {
     func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
         switch transaction.transferType {
         case .nativeCryptocurrency, .dapp:
+            estimateGasPrice()
```
hboon, line 160: Should `estimateGasPrice()` be called just within the `.nativeCryptoCurrency` and `.dapp` or before the `switch`?

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -112,6 +112,17 @@ class TransactionConfigurator {
         }
     }
 
+    func estimateGasPrice() {
```
hboon, line 115: Is `transaction.gasPrice` ever calculated earlier? Because `estimateGasPrice()` will override it.

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -112,6 +112,17 @@ class TransactionConfigurator {
...
+        let strongSelf = self
```
hboon, line 116: Better to move the strongSelf into the block in the next line, more idiomatic.

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
James-Sangalli: Yes, the other methods calculate it beforehand. With dapp transactions we don't get the chance

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
James-Sangalli: only for non dapp transactions

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
hboon: Then maybe we should split the 2 cases? So that we don't override the the gas price that came along with `.nativeCryptocurrency`?

James-Sangalli: @hboon done, don't even need the gas limit thing on natives

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -145,16 +156,20 @@ class TransactionConfigurator {
...
         switch transaction.transferType {
-        case .nativeCryptocurrency, .dapp:
-            guard requestEstimateGas else {
-                return completion(.success(()))
-            }
-            estimateGasLimit()
+        case .dapp:
+            estimateGasPrice()
+            completion(.success(()))
             configuration = TransactionConfiguration(
```
hboon, line 162: This should be before the `completion(.success(()))`?

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -145,16 +156,20 @@ class TransactionConfigurator {
 // swiftlint:disable function_body_length
     func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
         switch transaction.transferType {
-        case .nativeCryptocurrency, .dapp:
-            guard requestEstimateGas else {
-                return completion(.success(()))
-            }
-            estimateGasLimit()
```
hboon, line 152: Do we still need to call `estimateGasLimit()` for `.dapp`?

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -112,6 +112,17 @@ class TransactionConfigurator {
         }
     }
 
+    func estimateGasPrice() {
+        TransactionConfigurator.estimateGasPrice(server: self.session.server).done { gasPrice in
+            let strongSelf = self
```
hboon, line 117:
```
TransactionConfigurator.estimateGasPrice(server: self.session.server).done { [weak self] gasPrice in
```

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -155,6 +155,9 @@ class TransactionConfigurator {
 
 // swiftlint:disable function_body_length
     func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
+        if requestEstimateGas {
```
hboon, line 158: ! requestEstimateGas

:)

James-Sangalli: Haha no

hboon: Damn. Who named that variable!

Issue #1665, opened 2020-01-18
Update token icons in Wallet tab [Fiji]


tomekalphawallet:
We will have rounded icons 40x40. It's easy to generate for coins, more difficult for tokens without graphics. For those ones, use colors (Emerald, Indigo, Radical, Pumpkin, Cerulean, Azure, Mine) and up to 4 characters inside (ENS, SFPF, DEFI, DAO). Use the Tokenary app for inspiration. 

If AlphaWallet wants to be the best wallet to hold tokens, we need to make it attractive. In the future, I will provide few more designs how you can manage your tokens quickly (by holding and swiping left to reveal hidden menu) 

Sketch Cloud:
https://sketch.cloud/s/e4OJW/a/oE3MPQ
https://sketch.cloud/s/e4OJW/a/8MQgg8

Zeplin:
https://zpl.io/V0j8yZK
https://zpl.io/VxXB74J

- We are using the same card as we used in "Settings", so we have an icon 40x40, then the same Headline, secondary line (in terms of font)
https://zpl.io/2GdzLDj

tomekalphawallet:
![Tokens_with_icons](https://user-images.githubusercontent.com/51817359/72662253-3c4a0780-39e5-11ea-9ef0-6f960235d86f.png)

hboon:
@tomekalphawallet are the 24hrs/appreciation/value for the native crypto (eth and xDai) removed? If they are, can you help to make a version in the BC project with them? They'll want to keep it.

hboon:
- [ ] Update layout #1903 - token icon
- [ ] Add token icon

tomekalphawallet:
![Zrzut ekranu 2020-06-09 o 08 42 58-kopia](https://user-images.githubusercontent.com/51817359/84119718-0f9a9900-aa35-11ea-9655-5afb257b387c.png)
![Zrzut ekranu 2020-06-09 o 08 42 58](https://user-images.githubusercontent.com/51817359/84119727-11645c80-aa35-11ea-9b4d-629c1ff097c2.png)

hboon:
@tomekalphawallet can you also help provide the icons for these networks like you did for Ethereum mainnet and xDai?

- [ ] ARTIS — https://artis.eco
- [ ] POA — https://www.poa.network
- [ ] Ethereum classic — https://iconscout.com/icon/ethereum-classic-logo
- [ ] callisto — https://callisto.network

tomekalphawallet:
[...]

PR #1933, opened 2020-06-17
Add 3rd party sources for token icons — OpenSea (for ERC721) and a GitHub repo of images

hboon:
The token icons that are displayed in the Wallet tab and Send ERC20/native crypto screens.

Not included in this PR:

* Updated hardcoded native crypto icons in #1665 
* There could be a bug where the symbol in the programmatically generated icon is sometimes wrong. Will try to reproduce it separately.


James-Sangalli:
@hboon the build is breaking because of this #1934 but nothing to do with this PR

PR #1934, opened 2020-06-18
fix test case that incorrectly sets the gas limit to 90k when it should be 21k

James-Sangalli:


AlphaWalletTests/Transfer/Controllers/TransactionConfiguratorTests.swift
```diff
@@ -32,11 +32,11 @@ class TransactionConfiguratorTests: XCTestCase {
...
     
     func testLoadEtherConfiguration() {
-        let configurator = TransactionConfigurator(session: .make(), account: .make(), transaction: .make(gasLimit: BigInt(90000), gasPrice: .none))
+        let configurator = TransactionConfigurator(session: .make(), account: .make(), transaction: .make(gasLimit: BigInt(21000), gasPrice: .none))
 
         configurator.load { _ in }
         
         XCTAssertEqual(BigInt(GasPriceConfiguration.defaultPrice), configurator.configuration.gasPrice)
-        XCTAssertEqual(BigInt(90000), configurator.configuration.gasLimit)
+        XCTAssertEqual(BigInt(21000), configurator.configuration.gasLimit)
```
hboon, line 40: Looks like the problem is here instead, in `TransactionConfigurator.load(completion:)`:

```
if gasLimitNotSet {
	estimateGasLimit()
}
```

The old version only continues with the rest of the function if the gas limit has not been set:

```
guard gasLimitNotSet else {
	return completion(.success(()))
}
estimateGasLimit()
```

hboon: While we are at it, do you want to make the gas limit value in this test case a number that is not `90000` and `21000`? Easier to detect when it's failing, because those 2 numbers are used in `GasLimitConfiguration`.

James-Sangalli:
@hboon good for review again

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -155,11 +155,11 @@ class TransactionConfigurator {
...
 // swiftlint:disable function_body_length
     func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
-        if gasLimitNotSet {
-            estimateGasLimit()
-        }
         switch transaction.transferType {
         case .dapp:
+            if gasLimitNotSet {
+                estimateGasLimit()
+            }
```
hboon, line 162: Should it still do an early return like before #1926 is applied or is removing the early return intentional?

```
if gasLimitNotSet {
    estimateGasLimit()
} else {
    completion(.success(()))
}
```

James-Sangalli: Nope. In the case of dapp, it usually has the gas limit but it never sets the gas price properly (defaults to 9 gwei). The other cases can get the gas price properly.

PR #1938, opened 2020-06-19
Make add custom token more reliable

hboon:
Still with Infura, but changing or adding any other compatible RPC node will get the same benefit

This PR does it by:

* decreasing timeout from default 60s to 4s. Too short, and slower connections might not work with the app; too long a timeout, and it's too long a wait (especially since we introduced retries with this PR) when the server doesn't respond.
* retrying when individual smart contract call fails — we have to make many multiple smart contract calls to detect a token. If one of the calls fails, we retry instead of failing the whole process

The retry function is achieved by introducing a new `withRetry(times:task:)` function.


hboon:
- [ ] Ask Victor to try it out again

PR #1939, opened 2020-06-20
Allow user to set nonce for transaction

### LINKED ISSUE #1928: Allow users to set nonce in transaction advanced page
image

Since we won’t implement “speed up” and “cancel” transaction features any time soon, we should at least provide a “manual” way to do it. Otherwise, user will stuck with the pending transaction and there is no easy way to solve it.

hboon:
Closes #1928

Also displays nonce in

* transaction confirmation screen
* transaction (history) detail

|Before PR|After PR|
-|-
| 
|
|


James-Sangalli:
@hboon people only need this feature for two reasons:

1. Cancelling a transaction (in which case they send a blank tx with the same nonce but a higher gas price)
2. Speeding up a transaction (in which case they send the same tx but with a higher gas price and same nonce)

Since there are only two good reasons for this, I think it is better not to do it this way and to simply have a button on each transaction record to cancel or speed up if not already confirmed. 

Being able to arbitrarily set the nonce doesn't give the user what they are looking for which can only be 1 or 2. They should not have to know what a nonce is either.

hboon:
> Since there are only two good reasons for this, I think it is better not to do it this way and to simply have a button on each transaction record to cancel or speed up if not already confirmed.

I think a button to cancel or speed it up from the transaction is better too. But I presume it will take a while to design and properly implement it, otherwise, it'll just be more confusing; whereas tucking it in this advanced screen shouldn't confuse users, but still make it available to them if they know where to look.

@James-Sangalli But feel free to close this PR if the functionality isn't needed urgently.

hboon:
Closing since it's not desirable

zhangzhongnan928:
@James-Sangalli we should have this merged and in the next release.
Even I have experienced pending transaction issues several times. Without this "feature", the only way is waiting or exporting private key to a wallet with have such function.

For example, yesterday gf.network's GFT transaction has been pending for hours, and we can't do anything

zhangzhongnan928:
Tested the build, works very well.
First, all the functions work as expected 
Second, for normal user they hardly see the option, so won’t have the situations where users mistakenly change the nonce.

PR #1962, opened 2020-06-24
Fix: setter should use newValue, not the attribute name (which is the old value)

hboon:
Just happened to not make a difference in functionality, and it's (currently) only used in test code. Code was wrong, nevertheless.


hboon:
@James-Sangalli newbie mistake :P

PR #1963, opened 2020-06-24
Showing the send screen the first time for a token which has no exchange rate/price should not show the down arrow #1949

### LINKED ISSUE #1949: Showing the send screen the first time for a token which has no exchange rate/price should not show the down arrow
The behavior is correct (arrow not shown) when we scan an EIP681 link to switch from Ether (mainnet) to something that has no rate (eg. DAI/SAI)

vladyslav-iosdev:
Closes #1949

### DIFF
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -221,9 +221,9 @@ class SendViewController: UIViewController, CanScanQRCode {
         amountLabel.font = viewModel.textFieldsLabelFont
         amountLabel.textColor = viewModel.textFieldsLabelTextColor
         amountTextField.currentPair = viewModel.amountTextFieldPair
-        amountTextField.isAlternativeAmountEnabled = viewModel.isAlternativeAmountEnabled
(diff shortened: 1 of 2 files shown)


AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -221,9 +221,9 @@ class SendViewController: UIViewController, CanScanQRCode {
         amountLabel.font = viewModel.textFieldsLabelFont
         amountLabel.textColor = viewModel.textFieldsLabelTextColor
         amountTextField.currentPair = viewModel.amountTextFieldPair
-        amountTextField.isAlternativeAmountEnabled = viewModel.isAlternativeAmountEnabled
-        amountTextField.selectCurrencyButton.isHidden = viewModel.selectCurrencyButtonHidden
-        amountTextField.selectCurrencyButton.expandIconHidden = !viewModel.isAlternativeAmountEnabled
+        amountTextField.isAlternativeAmountEnabled = false//viewModel.isAlternativeAmountEnabled
```
hboon, line 224: Can you delete the commented out line if it's supposed to be removed?

PR #1945, opened 2020-06-23
Fix: App icon in launch (aka splash) screen should not move (up) abruptly #1897

hadwanihardik:
Fixes #1897

PR #1964, opened 2020-06-24
Fix: App icon in launch (aka splash) screen should not move (up) abruptly #1897

### LINKED ISSUE #1897: App icon in launch (aka splash) screen should not move (up) abruptly
During a cold launch, this happens:

1. the launch screen (which includes the app icon) appears
2. SplashView(Controller) appears (probably due to `ProtectionCoordinator`) which includes the app icon too.
3. SplashViewController is hidden

The app icon in (1) and (2)'s y position is slightly different so it looks like the logo moved up abruptly. The icon should not move at all.

Need to verify the fix works on 5s, iPhone X and a bigger than 5s phone with no notch.

hadwanihardik:
Fixes #1897


hadwanihardik:
Please let me know if something is missing.

hboon:
@hardikhadwani https://github.com/AlphaWallet/alpha-wallet-ios/pull/1945#pullrequestreview-435535028 and the entire Podfile.lock file is broken.

PR #1969, opened 2020-06-25
Price per token should be ETH not the token name #1961

### LINKED ISSUE #1961: Price per token should be ETH not the token name
IMG_EB6B295654F4-1

vladyslav-iosdev:
Closes #1961

### DIFF
--- a/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
+++ b/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
@@ -49,7 +49,7 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         return viewModel.token.server
     }
     let assetDefinitionStore: AssetDefinitionStore
-    lazy var pricePerTokenField = AmountTextField(tokenObject: viewModel.token)
+    lazy var pricePerTokenField = AmountTextField(tokenObject: viewModel.ethToken)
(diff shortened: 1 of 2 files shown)


AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
```diff
@@ -49,7 +49,7 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
         return viewModel.token.server
     }
     let assetDefinitionStore: AssetDefinitionStore
-    lazy var pricePerTokenField = AmountTextField(tokenObject: viewModel.token)
```
hboon, line 52: I think the fix might be simpler and more correct by replacing:

```
AmountTextField(tokenObject: viewModel.token)
```

with:

```
AmountTextField(tokenObject: ethToken)
```

And you can get ethToken via `TokensDataStore(forServer:)`.

Can you see if that makes sense and is right? The problem being in the send screen, we want the amount textfield for entering the amount of token the send screen is for. But in this screen, no matter which token it is, we want the amount textfield to be for entering Ether (or whichever native crypto token like xDai, etc).

James-Sangalli:
@vladyslav-iosdev please use the main repo instead of a fork from now, cheers.

hboon:
@vladyslav-iosdev is this good for review?

vladyslav-iosdev:
> @vladyslav-iosdev is this good for review?

yes

PR #1970, opened 2020-06-25
Reduce number of NumberFormatter and EtherNumberFormatter defined, if possible #1954

### LINKED ISSUE #1954: Reduce number of NumberFormatter and EtherNumberFormatter defined, if possible
Do a sweep, especially in `CurrencyFormatter.swift` and `EtherNumberFormatter.swift`.

Generally, we want to round down to 4 decimal places for crypto balances (but for transaction amounts, we display the full value) and round down to 2 decimal places for fiat.

Can you extract the `4` decimal places to a constant? (the other project fork will want 5 instead)

Take care with refactoring because:

* amounts are sensitive
* the other project fork is affected by this too. We'll probably copy these changes over to the other repo.

Add test cases with something like this:

```
        let values: [(BigInt, String)] = [
            (BigInt("1123450000000000000"), "1.1234"),
            (BigInt("1123440000000000000"), "1.1234"),
            (BigInt("1123490000000000000"), "1.1234"),
            (BigInt("1123499000000000000"), "1.1234"),
            (BigInt("1123990000000000000"), "1.1239"),

            (BigInt("1123850000000000000"), "1.1238"),
            (BigInt("1123840000000000000"), "1.1238"),
            (BigInt("1123890000000000000"), "1.1238"),
            (BigInt("1123899000000000000"), "1.1238"),
        ]
        for each in values {
            //EtherNumberFormatter.short.string(from: each.0, decimals: 18)
        }
```

vladyslav-iosdev:
Closes #1954

### DIFF
--- a/AlphaWallet/Core/Formatters/CurrencyFormatter.swift
+++ b/AlphaWallet/Core/Formatters/CurrencyFormatter.swift
@@ -2,47 +2,45 @@
 
 import Foundation
 
-class CurrencyFormatter {
-    static var formatter: NumberFormatter {
-        let formatter = NumberFormatter()
-        formatter.minimumFractionDigits = 2
-        formatter.maximumFractionDigits = 2
-        formatter.roundingMode = .down
-        //TODO support multiple currency values
-        formatter.currencyCode = Currency.USD.rawValue
-        formatter.numberStyle = .currency
+extension NumberFormatter {
 
-        return formatter
+    static let currency = Formatter(.currency)
+    static let usd = Formatter(.usd)
+
+    class Formatter {
+        private let formatter: NumberFormatter
+
+        fileprivate init(_ configuration: NumberFormatterConfiguration) {
+            formatter = configuration.formatter
+        }
+
+        func string(from number: Double) -> String? {
+            return formatter.string(from: number as NSNumber)
+        } 
     }
+}
 
-    static var usdFormatter: NumberFormatter {
+private enum NumberFormatterConfiguration {
+    case usd
+    case currency
+
+    var formatter: NumberFormatter {
         let formatter = NumberFormatter()
-        formatter.minimumFractionDigits = 2
-        formatter.maximumFractionDigits = 2
         formatter.roundingMode = .down
-        //TODO support multiple currency values
         formatter.numberStyle = .currency
-        formatter.positiveFormat = "0.00" + " " + Constants.Currency.usd
-        formatter.negativeFormat = "-0.00" + " " + Constants.Currency.usd
+        formatter.minimumFractionDigits = Constants.formatterFractionDigits
+        formatter.maximumFractionDigits = Constants.formatterFractionDigits
 
-        return formatter
-    }
+        switch self {
+        case .currency:
+            //TODO support multiple currency values
+            formatter.currencyCode = Currency.USD.rawValue
+        case .usd:
+            formatter.positiveFormat = "0.00" + " " + Constants.Currency.usd
+            formatter.negativeFormat = "-0.00" + " " + Constants.Currency.usd
+            formatter.currencyCode = String()
+        }
 
-    static var plainFormatter: EtherNumberFormatter {
-        let formatter = EtherNumberFormatter.full
-        formatter.groupingSeparator = ""
-        return formatter
-    }
-
-    static var shortFormatter: NumberFormatter {
-        let formatter = NumberFormatter()
-        formatter.numberStyle = .currency
-        formatter.minimumFractionDigits = 4
-        formatter.maximumFractionDigits = 4
-        formatter.roundingMode = .down
-        formatter.positiveFormat = "0.0000"
-        formatter.negativeFormat = "-0.0000"
-        
         return formatter
     }
 }
--- a/AlphaWallet/Export/ViewModels/PromptBackupWalletAfterExceedingThresholdViewViewModel.swift
+++ b/AlphaWallet/Export/ViewModels/PromptBackupWalletAfterExceedingThresholdViewViewModel.swift
@@ -16,7 +16,7 @@ struct PromptBackupWalletAfterExceedingThresholdViewViewModel: PromptBackupWalle
     }
 
     var description: String {
-        let prettyAmount = CurrencyFormatter.formatter.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
+        let prettyAmount = NumberFormatter.currency.string(from: dollarValueInUsd) ?? "-"
         return R.string.localizable.backupPromptAfterHittingThresholdDescription(prettyAmount)
     }
 
--- a/AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
+++ b/AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
@@ -50,12 +50,7 @@ class EthCurrencyHelper {
 
     public func valueChanged24h(currencyAmountWithoutSymbol: Double?) -> String? {
         if let percentChange = percentageChange24h, let value = currencyAmountWithoutSymbol {
-            let change = value * percentChange / 100
-            if let string = CurrencyFormatter.usdFormatter.string(from: NSNumber(value: change)) {
-                return string
-            } else {
-                return nil
-            }
+            return NumberFormatter.usd.string(from: value * percentChange / 100)
         } else {
             return nil
         }
--- a/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
@@ -11,14 +11,6 @@ class TokensViewModel {
     private let filterTokensCoordinator: FilterTokensCoordinator
     var tokens: [TokenObject]
     let tickers: [RPCServer: [AlphaWallet.Address: CoinTicker]]
-    private var amount: String? {
-        var totalAmount: Double = 0
-        filteredTokens.forEach { token in
-            totalAmount += amount(for: token)
-        }
-        guard totalAmount != 0 else { return "--" }
-        return CurrencyFormatter.formatter.string(from: NSNumber(value: totalAmount))
-    }
 
     var filter: WalletFilter = .all {
         didSet {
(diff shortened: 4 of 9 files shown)


AlphaWallet/Export/ViewModels/PromptBackupWalletAfterExceedingThresholdViewViewModel.swift
```diff
@@ -16,7 +16,7 @@ struct PromptBackupWalletAfterExceedingThresholdViewViewModel: PromptBackupWalle
     }
 
     var description: String {
-        let prettyAmount = CurrencyFormatter.formatter.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
+        let prettyAmount = NumberFormatter().currency.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
```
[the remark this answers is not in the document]
vladyslav-iosdev: > I'll need to check it somemore, but what if instead of:
> 
> ```
> let prettyAmount = NumberFormatter().currency.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
> ```
> 
> we make it like this:
> 
> ```
> let prettyAmount = NumberFormatter.currency.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
> ```
> 
> So we don't need to create an instance of `NumberFormatter` every time, and also we can cache them as static vars? formatters have poor initialization performance and we since we really only have a handful unique ones, it feels like we can make do with static versions.
> 
> The new names and API are definitely more Swifty though. They are good.

@hboon it seems like done, you can check it.

vladyslav-iosdev:
> I suppose the API is so that we are meant to use it like this:
> 
> ```
> if let string = NumberFormatter.currency.usd.string(from: NSNumber(value: change)) {
> ```
> 
> i.e. with a `.usd` or `.short`
> 
> There are these lines which omits them:
> 
> ```
> let prettyAmount = NumberFormatter.currency.string(from: NSNumber(value: dollarValueInUsd)) ?? "-"
> 
> return NumberFormatter.currency.string(from: NSNumber(value: totalAmount))
> 
> let formattedFee = NumberFormatter.currency.string(from: NSNumber(value: price.price * feeInDouble))
> ```
> 
> This introduces a bug right? Since the formatter might have been changed in a previous call, we don't actually know which format is being used.
> 
> If this is unintended, how about we make `NumberFormatter.currency` return an instance of another (maybe inner) class which is not `NumberFormatter` nor a subclass of it to make the error impossible?

damn, you right. i missed it. i will fix this.

vladyslav-iosdev:
>  If this is unintended, how about we make `NumberFormatter.currency` return an instance of another (maybe inner) class which is not `NumberFormatter` nor a subclass of it to make the error impossible?
> 

i am not shure if i undarstand your last lines, but check chenges and let me know.

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
```diff
@@ -2,47 +2,57 @@
...
-        let formatter = NumberFormatter()
-        formatter.minimumFractionDigits = 2
-        formatter.maximumFractionDigits = 2
-        formatter.roundingMode = .down
-        //TODO support multiple currency values
-        formatter.currencyCode = Currency.USD.rawValue
-        formatter.numberStyle = .currency
+extension NumberFormatter {
+
+    //NOTE: Not shure about name
```
hboon, line 7: Typo: "sure"

vladyslav-iosdev: ;)

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
```diff
@@ -2,47 +2,62 @@
...
+
+        fileprivate init(_ configuration: NumberFormatterConfiguration) {
+            formatter = configuration.formatter
+        }
+
+        func string(from number: NSNumber) -> String? {
+            return formatter.string(from: number)
+        }
+
+        func string(for: Any?) -> String? {
```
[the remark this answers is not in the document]
hboon: But I realise we have both:

```
func string(from number: NSNumber) -> String? {
    return formatter.string(from: number)
}

func string(for: Any?) -> String? {
    return formatter.string(for: number)
}
```

It seems like `NumberFormatter.string(from:)` and `Formatter.string(for:)` do the same thing. Do you know if they are different? Otherwise let's just use `NumberFormatter.string(from:)` since it's in a subclass.

It also looks like we always call with a `Double`, so we can remove this:

```
func string(from number: NSNumber) -> String? {
```

and replace the second version with this:

```
func string(from number: Double) -> String? {
    return formatter.string(from: number as NSNumber)
}
```

What do you think?

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
```diff
@@ -2,47 +2,62 @@
...
+            formatter.currencyCode = Currency.USD.rawValue
+        case .short:
+            formatter.minimumFractionDigits = Constants.currencyFormatterFractionDigits
+            formatter.maximumFractionDigits = Constants.currencyFormatterFractionDigits
+
+            formatter.positiveFormat = "0.0000"
+            formatter.negativeFormat = "-0.0000"
+            formatter.currencyCode = String()
+        case .usd:
+            formatter.minimumFractionDigits = Constants.currencyFormatterFractionDigits
```
hboon, line 53: This is a bit strange because `Constants.currencyFormatterFractionDigits = 4`. But it seems to work (because the result is 2 decimal places). Do you know why? And can you make it clearer?

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
```diff
@@ -2,47 +2,62 @@
...
 
-    static var plainFormatter: EtherNumberFormatter {
-        let formatter = EtherNumberFormatter.full
-        formatter.groupingSeparator = ""
-        return formatter
-    }
+            //TODO support multiple currency values
+            formatter.currencyCode = Currency.USD.rawValue
+        case .short:
+            formatter.minimumFractionDigits = Constants.currencyFormatterFractionDigits
```
hboon, line 46: I think if the number of decimals is set to 4 in `CurrencyFormatter`, it has to be moved to `EtherNumberFormatter` and the call site will have to use `EtherNumberFormatter` instead. Example in `SendViewModel`:

```
let value = NumberFormatter.short.string(for: balanceValue)!
```

Similarly EtherNumberFormatter should only have decimals=4 (if there's code that set it to 2, it should be moved to `CurrencyFormatter`.

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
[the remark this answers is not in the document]
vladyslav-iosdev: yep it will be great improvement, i will update it.

AlphaWallet/Transfer/ViewModels/SendViewModel.swift
```diff
@@ -91,13 +91,13 @@ struct SendViewModel {
         switch transferType {
         case .nativeCryptocurrency:
             if let balance = session.balance {
-                let balanceValue = CurrencyFormatter.plainFormatter.string(from: balance.value).doubleValue
-                let value = CurrencyFormatter.shortFormatter.string(for: balanceValue)!
+                let balanceValue = EtherNumberFormatter.plain.string(from: balance.value).doubleValue
+                let value = NumberFormatter.short.string(for: balanceValue)!
```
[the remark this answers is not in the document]
vladyslav-iosdev: what if i will return just a string, non optional string?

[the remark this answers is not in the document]
hboon: Just ""? Should be ok too.

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
vladyslav-iosdev: actually i dont know, i guess it because we set `positiveFormat/negativeFormat`, and it makes calculations depend on them.

hboon: Very strange. Can you try to do this: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1970#discussion_r448122914 and maybe the strangeness will go away. Basically keep 4 decimal places "mentioned" only in 1 formatter file and 2 decimal places in the other formatter file.

AlphaWallet/Core/Formatters/CurrencyFormatter.swift
vladyslav-iosdev: @hboon seems like done, check it.

AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
```diff
@@ -51,7 +51,7 @@ class EthCurrencyHelper {
     public func valueChanged24h(currencyAmountWithoutSymbol: Double?) -> String? {
         if let percentChange = percentageChange24h, let value = currencyAmountWithoutSymbol {
             let change = value * percentChange / 100
-            if let string = CurrencyFormatter.usdFormatter.string(from: NSNumber(value: change)) {
+            if let string = NumberFormatter.usd.string(from: change) {
```
hboon, line 54: Sorry missed this one. The if-let-else isn't necessary. Just return `NumberFormatter.usd.string(from: change)`.

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -17,7 +17,7 @@ class TokensViewModel {
             totalAmount += amount(for: token)
         }
         guard totalAmount != 0 else { return "--" }
-        return CurrencyFormatter.formatter.string(from: NSNumber(value: totalAmount))
+        return NumberFormatter.currency.string(from: totalAmount)
```
hboon, line 20: It looks like the `amount` property is not used anymore. Shall we remove it?

AlphaWallet/Transfer/ViewModels/SendViewModel.swift
```diff
@@ -91,13 +91,11 @@ struct SendViewModel {
...
             if let balance = session.balance {
-                let balanceValue = CurrencyFormatter.plainFormatter.string(from: balance.value).doubleValue
-                let value = CurrencyFormatter.shortFormatter.string(for: balanceValue)!
+                let value = EtherNumberFormatter.plain.string(from: balance.value)
                 return R.string.localizable.sendAvailable("\(value) \(transferType.symbol)")
             }
         case .ERC20Token(let token, _, _):
-            let balanceValue = CurrencyFormatter.plainFormatter.string(from: token.valueBigInt, decimals: token.decimals).doubleValue
-            let value = CurrencyFormatter.shortFormatter.string(for: balanceValue)!
+            let value = EtherNumberFormatter.plain.string(from: token.valueBigInt, decimals: token.decimals)
```
hboon, line 98: Oh. This is much clearer. Thanks!

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
vladyslav-iosdev: seems like we can to remove it.

vladyslav-iosdev:
@hboon seems like this one is ready to check.

PR #1971, opened 2020-06-25
add type to asnx:module syntax

James-Sangalli:
https://github.com/AlphaWallet/TokenScript-Examples/pull/93

### DIFF
--- a/AlphaWallet/TokenScriptClient/Models/XMLHandler+XPaths.swift
+++ b/AlphaWallet/TokenScriptClient/Models/XMLHandler+XPaths.swift
@@ -106,7 +106,7 @@ extension XMLHandler {
         let addressElements = XMLHandler.getAddressElements(fromContractElement: contractElement, xmlContext: xmlContext)
         guard let address = addressElements.first?.text.flatMap({ AlphaWallet.Address(string: $0.trimmed) }) else { return nil }
         guard let eventName = asnModuleElement.at_xpath("namedType", namespaces: xmlContext.namespaces)?["name"] else { return nil }
-        let parameters = asnModuleElement.xpath("namedType/type/sequence/element", namespaces: xmlContext.namespaces).compactMap { each -> EventParameter? in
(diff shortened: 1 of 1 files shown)


AlphaWallet/TokenScriptClient/Models/XMLHandler+XPaths.swift
```diff
@@ -106,7 +106,7 @@ extension XMLHandler {
         let addressElements = XMLHandler.getAddressElements(fromContractElement: contractElement, xmlContext: xmlContext)
         guard let address = addressElements.first?.text.flatMap({ AlphaWallet.Address(string: $0.trimmed) }) else { return nil }
         guard let eventName = asnModuleElement.at_xpath("namedType", namespaces: xmlContext.namespaces)?["name"] else { return nil }
-        let parameters = asnModuleElement.xpath("namedType/sequence/element", namespaces: xmlContext.namespaces).compactMap { each -> EventParameter? in
+        let parameters = asnModuleElement.xpath("namedType/type/sequence/element", namespaces: xmlContext.namespaces).compactMap { each -> EventParameter? in
```
hboon, line 109: @James-Sangalli might need to support both with and without the `type` tag. Depends on the conversation starting from https://github.com/AlphaWallet/TokenScript/pull/365#issuecomment-649322308

James-Sangalli: Only ENS is broken. Let's not keep trash around. We can also make it into another schema... @colourful-land

colourful-land: We need to support both with and without because you can't ensure all αW users upgrade over night.

1. Roll out a patch release for  αW andoird + iOS
2. Wait a few weeks
3. Update repo server (expecting not many to still use the old αW which will crash with `<typp>`

colourful-land: @James-Sangalli can you help with testing the modification I made which should allow both cases? If it works on your iOS, we should release asap so we can do the Repo update almost at the same time as when Activity is done.

PR #1972, opened 2020-06-25
Extract a UILabel wrapper or subclass which displays Ethereum address and/or ENS name #1950

### LINKED ISSUE #1213: ENS loading issues
1. If the network is slow, users may submit a transition before ENS load up. It will end up with Invalid Address error message.

- [ ] need an Indication to tell users, the address is loading up.

2. Once triggered Invalid Address error message, the wallet will stop loading the ENS. Means once users click the send before ENS load up, the ENS will never load up again, unless you kill the app and restart it.

- [ ] need to continue loading.

### LINKED ISSUE #1950: Extract a UILabel wrapper or subclass which displays Ethereum address and/or ENS name
There at at least a few possibilities:

* Just address (sometimes the full address, sometimes shortened to 0xabcd...abcd)
* Just ENS
* "<ENS> | <address>" (in Settings screen)
* Maybe more?

Can you replace all of them with a single class with a way to configure them?

If it makes sense, let it be able to resolve, reverse resolve ENS names too, but might not be.

### LINKED ISSUE #1977: Pasting an address in Import Wallet should not wait until ENS resolution is complete before enabling the watch button
Also see if this affects other places where `AddressTextField` is used. Eg send screen.

vladyslav-iosdev:
Closes #1950, closes #1213, closes #1977

### DIFF
--- a/AlphaWallet/Core/Views/AddressOrEnsNameLabel.swift
+++ b/AlphaWallet/Core/Views/AddressOrEnsNameLabel.swift
@@ -1,0 +1,136 @@
+//
+//  AddressOrEnsNameLabel.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 23.06.2020.
+//
+
+import UIKit
+
+class AddressOrEnsNameLabel: UILabel {
+
+    enum AddressOrEnsResolution {
+        case invalidInput
+        case resolved(AddressOrEnsName?)
+    }
+
+    enum AddressFormat {
+        case full
+        case truncateMiddle
+
+        func formattedAddress(_ address: AlphaWallet.Address) -> String {
+            switch self {
+            case .full:
+                return address.eip55String
+            case .truncateMiddle:
+                return address.truncateMiddle
+            }
+        }
+    }
+
+    private let serverToResolveEns = RPCServer.main
+    private var inResolvingState: Bool = false {
+        didSet {
+            if inResolvingState && shouldShowLoadingIndicator {
+                loadingIndicator.startAnimating()
+            } else {
+                loadingIndicator.stopAnimating()
+            }
+        }
+    }
+
+    let loadingIndicator: UIActivityIndicatorView = {
+        let indicator = UIActivityIndicatorView(style: .gray)
+        indicator.hidesWhenStopped = true
+        indicator.translatesAutoresizingMaskIntoConstraints = false
+        indicator.heightAnchor.constraint(equalToConstant: 15).isActive = true
+        indicator.widthAnchor.constraint(equalToConstant: 15).isActive = true
+
+        return indicator
+    }()
+
+    var addressString: String? {
+        switch addressOrEnsName {
+        case .address(let value):
+            return value.eip55String
+        case .ensName, .none:
+            return nil
+        }
+    }
+
+    var addressOrEnsName: AddressOrEnsName? {
+        didSet {
+            switch addressOrEnsName {
+            case .some(.address(let address)):
+                text = addressFormat.formattedAddress(address)
+            case .some(.ensName(let string)):
+                text = string
+            case .none:
+                text = nil
+            }
+
+            isHidden = text == nil
+        }
+    }
+
+    var stringValue: String? {
+        return addressOrEnsName?.stringValue
+    }
+
+    var addressFormat: AddressFormat = .truncateMiddle
+    var shouldShowLoadingIndicator: Bool = false
+
+    init() {
+        super.init(frame: .zero)
+        translatesAutoresizingMaskIntoConstraints = false
+        numberOfLines = 0
+        setContentHuggingPriority(.required, for: .horizontal)
+        setContentCompressionResistancePriority(.required, for: .horizontal)
+
+        textColor = DataEntry.Color.ensText
+        font = DataEntry.Font.label
+        textAlignment = .center
+    }
+
+    required init?(coder: NSCoder) {
+        return nil
+    }
+
+    func clear() {
+        addressOrEnsName = nil
+        inResolvingState = false
+    }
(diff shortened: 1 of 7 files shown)


AlphaWallet/UI/AddressTextField.swift
```diff
@@ -15,27 +15,16 @@ class AddressTextField: UIControl {
...
-        switch addressOrEnsName {
-        case .address(let value):
-            return value.eip55String
-        case .ensName, .none:
-            return nil
-        }
-    }
+    private let ensAddressLabel: AddressOrEnsNameLabel = {
+        let label = AddressOrEnsNameLabel()
+        label.addressFormat = .full
```
hboon, line 20: This should be `.truncateMiddle`

hboon:
Does this address both points in #1213 ? (if so you can mention "Closes #1213" too in the 1st comment in this PR).

vladyslav-iosdev:
> Closes #1213"

yes i does

hboon:
> > Closes #1213"
> 
> yes i does

You'll have to repeat the closes (or fixes keywords) like this: "Closes #xyz, closes #abc"

vladyslav-iosdev:
> 1. hweeboon.eth

@hboon as i undarstand the problem is in that we can't cancel sent request. What if i wrap body of getENSAddressFromResolver into DispatchWorkItem that allow us to cancel previous sent getENSAddressFromResolver operation?

hboon:
What if when we get the resolved result, we compare the "source" we have just resolved with the textfield contents? e.g (this is code from before the PR, I just added a line):

```
ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { [weak self] result in
    guard let resolvedESNname = result.value else {
        **COMPARE HERE `address` vs current value of textField **
        //Don't show an error when pasting what seems like a wrong ENS name for better usability
        strongSelf.addressOrEnsName = nil
        strongSelf.delegate?.didPaste(in: strongSelf)
        return
    }

    strongSelf.addressOrEnsName = .ensName(resolvedESNname)

    strongSelf.delegate?.didPaste(in: strongSelf)
}
```

Is this simpler?

vladyslav-iosdev:
> What if when we get the resolved result, we compare the "source" we have just resolved with the textfield contents? e.g (this is code from before the PR, I just added a line):
> 
> ```
> ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { [weak self] result in
>     guard let strongSelf = self else { return }
>     guard let resolvedESNname = result.value else {
>         **COMPARE HERE `address` vs current value of textField **
>         //Don't show an error when pasting what seems like a wrong ENS name for better usability
>         strongSelf.addressOrEnsName = nil
>         strongSelf.delegate?.didPaste(in: strongSelf)
>         return
>     }
> 
>     strongSelf.addressOrEnsName = .ensName(resolvedESNname)
> 
>     strongSelf.delegate?.didPaste(in: strongSelf)
> }
> ```
> 
> Is this simpler?

yep, you right:) Done.

AlphaWallet/UI/AddressTextField.swift
```diff
@@ -264,116 +228,43 @@ class AddressTextField: UIControl {
...
-                return
-            }
-
-            ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { [weak self] result in
-                guard let strongSelf = self else { return }
-                guard let resolvedESNname = result.value else {
-                    //Don't show an error when pasting what seems like a wrong ENS name for better usability
-                    strongSelf.addressOrEnsName = nil
-                    strongSelf.delegate?.didPaste(in: strongSelf)
+            ensAddressLabel.resolve(value) { [weak self] resolvation in
```
hboon, line 239: "resolvation" -> "resolution"

AlphaWallet/Core/Views/AddressOrEnsNameLabel.swift
```diff
@@ -0,0 +1,136 @@
...
+    func resolve(_ value: String, completion: @escaping ((AddressOrEnsResolution) -> Void)) {
+        clear()
+        
+        if let address = AlphaWallet.Address(string: value) {
+            inResolvingState = true
+            ENSReverseLookupCoordinator(server: serverToResolveEns).getENSNameFromResolver(forAddress: address) { [weak self] result in
+                guard let strongSelf = self else { return }
+                strongSelf.inResolvingState = false
+
+                if let resolvedESNname = result.value {
```
hboon, line 113: "resolvedESNname" -> "resolvedEnsName"

AlphaWallet/UI/AddressTextField.swift
```diff
@@ -405,27 +296,39 @@ extension AddressTextField: UITextFieldDelegate {
...
+            self.ensAddressResolve(value: newValue)
         }
 
-        informDelegateDidChange(to: newValue ?? "")
+        informDelegateDidChange(to: newValue)
         return true
     }
 
+    private func ensAddressResolve(value: String) {
+        ensAddressLabel.resolve(value) { [weak self] resolvation in
```
hboon, line 313: "resolvation -> "resolution" (but result might be more common?)

AlphaWallet/UI/AddressTextField.swift
```diff
@@ -405,27 +296,39 @@ extension AddressTextField: UITextFieldDelegate {
...
+            guard let strongSelf = self, value == strongSelf.textField.text?.trimmed else {
+                return
+            }
+
+            switch resolvation {
```
hboon, line 318: Seems like this switch can be extracted into a function. It's quite similar in `pasteAction()`, except for an error being displayed?

PR #1973, opened 2020-06-25
Make tap token in Wallet tab push instead of present modal #1956

### LINKED ISSUE #1956: Make tap token in Wallet tab push instead of present modal
There are many of these across the app that we will want to replace ultimately or change the modal presentation style to the card-style that comes with iOS 13. We have to be careful that it still displays correctly (especially the navigation bar) and that it gets popped (vs dismissed) correctly, and that clean up is done correctly. The coordinators might have been written to assume that it is presented modally and not pushed onto a navigation controller.

There should be a call to `makePresentationFullScreenForiOS13Migration()` for each of this change.  They are mainly to track these as TODOs. The call should be removed (and replaced with the explicit `modalPresentationStyle = xxx` line if we still want it to be modal).

This issue is for 1 of them:

You can trace it from `TokensViewController.tableView(_:didSelectRowAt:)`.

vladyslav-iosdev:
Closes #1956

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
@@ -318,6 +318,7 @@ class SingleChainTokenCoordinator: Coordinator {
 
         let tokensCardCoordinator = TokensCardCoordinator(
                 session: session,
+                navigationController: navigationController,
                 keystore: keystore,
                 tokensStorage: storage,
                 ethPrice: cryptoPrice,
@@ -328,17 +329,12 @@ class SingleChainTokenCoordinator: Coordinator {
         addCoordinator(tokensCardCoordinator)
         tokensCardCoordinator.delegate = self
         tokensCardCoordinator.start()
+
         switch (type, session.account.type) {
         case (.send, .real), (.request, _):
             makeCoordinatorReadOnlyIfNotSupportedByOpenSeaERC721(coordinator: tokensCardCoordinator, token: token)
-            tokensCardCoordinator.navigationController.makePresentationFullScreenForiOS13Migration()
-            navigationController.present(tokensCardCoordinator.navigationController, animated: true, completion: nil)
         case (.send, .watch):
             tokensCardCoordinator.isReadOnly = true
-            tokensCardCoordinator.navigationController.makePresentationFullScreenForiOS13Migration()
-            navigationController.present(tokensCardCoordinator.navigationController, animated: true, completion: nil)
-        case (_, _):
-            navigationController.displayError(error: InCoordinatorError.onlyWatchAccount)
         }
     }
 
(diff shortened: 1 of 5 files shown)


hboon:
@vladyslav-iosdev adding the "WIP" tag here instead of in the issue. I presume you mean for this PR.

AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
```diff
@@ -31,7 +31,7 @@ class SingleChainTokenCoordinator: Coordinator {
     private let cryptoPrice: Subscribable<Double>
     private let assetDefinitionStore: AssetDefinitionStore
     private let eventsDataStore: EventsDataStoreProtocol
-    private let navigationController: UINavigationController
+    private let navigationController: NavigationController
```
hboon, line 34: Seems like we can keep this as `UINavigationController` and to change `TokensCardCoordinator.swift` to use a `UINavigationController` instead.

The `NavigationController` class doesn't seems to be unnecessary anymore. We can probably remove it in a separate PR.

AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
```diff
@@ -327,19 +328,22 @@ class SingleChainTokenCoordinator: Coordinator {
         )
         addCoordinator(tokensCardCoordinator)
         tokensCardCoordinator.delegate = self
-        tokensCardCoordinator.start()
+
         switch (type, session.account.type) {
         case (.send, .real), (.request, _):
```
hboon, line 333: I think we need the `tokensCardCoordinator.start()` line needs to be moved into the switch so it gets called for the first 2 cases, but not the 3rd/last?

AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
```diff
@@ -327,19 +328,22 @@ class SingleChainTokenCoordinator: Coordinator {
...
             makeCoordinatorReadOnlyIfNotSupportedByOpenSeaERC721(coordinator: tokensCardCoordinator, token: token)
-            tokensCardCoordinator.navigationController.makePresentationFullScreenForiOS13Migration()
-            navigationController.present(tokensCardCoordinator.navigationController, animated: true, completion: nil)
+//            tokensCardCoordinator.navigationController.makePresentationFullScreenForiOS13Migration()
+//            navigationController.present(tokensCardCoordinator.navigationController, animated: true, completion: nil)
         case (.send, .watch):
             tokensCardCoordinator.isReadOnly = true
```
hboon, line 338: isReadyOnly doesn't seem to work probably anymore for watched wallets

AlphaWallet/Transactions/Coordinators/TokensCardCoordinator.swift
```diff
@@ -64,7 +64,8 @@ class TokensCardCoordinator: NSObject, Coordinator {
 
     func start() {
         rootViewController.configure()
-        navigationController.viewControllers = [rootViewController]
+//        navigationController.viewControllers = [rootViewController]
+        navigationController.pushViewController(rootViewController, animated: true)
```
hboon, line 68: Can you hide the tab bar when pushed?

hboon:
@vladyslav-iosdev good for review?

vladyslav-iosdev:
> @vladyslav-iosdev good for review?

@hboon Yes.

AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
```diff
@@ -515,7 +508,7 @@ class SingleChainTokenCoordinator: Coordinator {
 
 extension SingleChainTokenCoordinator: TokensCardCoordinatorDelegate {
     func didCancel(in coordinator: TokensCardCoordinator) {
-        navigationController.dismiss(animated: true)
+        navigationController.popToRootViewController(animated: true)//.dismiss(animated: true)
```
hboon, line 511: Can you remove the commented code?

vladyslav-iosdev: yep, done.

vladyslav-iosdev:
> Test suite is broken. I think it's easier if you change one or more of the `NavigationController` to `UINavigationController` :)

Done.

PR #1975, opened 2020-06-26
Fix add a custom token in Add / Hide Tokens screen does not show the token in the first list #1968

### LINKED ISSUE #1968: Fix add a custom token in Add / Hide Tokens screen does not show the token in the first list
Verify if true first.

This might happen because we disabled refreshing:

```
    private func handleTokenCollectionUpdates() {
        //no-op because we do not want to refresh this screen as it's a bad user experience if the tokens reload while the user is toggling their visibility
    }
```

This was disabled because if the list of tokens refreshes while the user changes the position of the token or switch between show/hiding them, the changes will be lost.

vladyslav-iosdev:
Closes #1968

### DIFF
--- a/AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
@@ -19,7 +19,8 @@ private enum AddHideTokenSections: Int {
     }
 }
 
-struct AddHideTokensViewModel {
+//NOTE: Changed to class to prevent update all ViewModel copies and apply updates only in one place.
+class AddHideTokensViewModel {
     private let sections: [AddHideTokenSections] = [.displayedTokens, .hiddenTokens]
     private let filterTokensCoordinator: FilterTokensCoordinator
     private var tokens: [TokenObject]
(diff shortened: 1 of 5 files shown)


AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
```diff
@@ -19,10 +19,10 @@ private enum AddHideTokenSections: Int {
     }
 }
 
-struct AddHideTokensViewModel {
+class AddHideTokensViewModel {
     private let sections: [AddHideTokenSections] = [.displayedTokens, .hiddenTokens]
     private let filterTokensCoordinator: FilterTokensCoordinator
-    private var tokens: [TokenObject]
+    var tokens: [TokenObject]
```
hboon, line 25: Looks like `tokens` can still be private

AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
```diff
@@ -19,10 +19,10 @@ private enum AddHideTokenSections: Int {
     }
 }
 
-struct AddHideTokensViewModel {
+class AddHideTokensViewModel {
```
hboon, line 22: Why do you change it from a struct to a class?

AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
vladyslav-iosdev: ooh, yep you right.

AlphaWallet/Tokens/ViewModels/AddHideTokensViewModel.swift
hboon: If there's a special reason, can you leave a comment like in `TokensViewModel`? Otherwise, it's easy in the future to change it back and the code only breaks silently.

vladyslav-iosdev: i changed it to class because we need to refresh view model in apply updates in only one place, then to not to update separately coordinator and view controller.

vladyslav-iosdev: ok, np.

PR #1983, opened 2020-07-01
Show current market price instead of price diff in wallet tab #1981

### LINKED ISSUE #1981: Show current market price instead of price diff in wallet tab
Amount + Ticker ( 3.4525 ETH)
Value in USD (674 USD)----------------------- Market Price ( 226 USD) Price change in % (+2.34%)

vladyslav-iosdev:
Closes #1981

### DIFF
--- a/AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
+++ b/AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
@@ -24,6 +24,10 @@ class EthCurrencyHelper {
         }
     }
 
+    var marketPrice: Double? {
+        return ticker?.price_usd
+    }
+
     private var percentageChange24h: Double? {
         if let percent_change_24h = ticker?.percent_change_24h {
             return percent_change_24h.rounded(to: 2)
--- a/AlphaWallet/Tokens/ViewModels/EthTokenViewCellViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/EthTokenViewCellViewModel.swift
@@ -98,8 +98,12 @@ struct EthTokenViewCellViewModel {
         }
     }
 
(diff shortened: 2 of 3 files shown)


vladyslav-iosdev:
@hboon i am not sure if i correctly get market price value for tiket.

hboon:
> @hboon i am not sure if i correctly get market price value for tiket.

Looks right

AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
```diff
@@ -24,6 +24,14 @@ class EthCurrencyHelper {
         }
     }
 
+    var marketPrice: Double? {
+        if let price = ticker?.price_usd {
+            return price.rounded(to: 2)
```
hboon, line 29: Let's wait until #1970 is merged and then we can use CurrencyFormatter here. and a few lines below without the hardcoded '2'.

hboon: @vladyslav-iosdev would you help to update and remove the '2' here and below since #1970 has been merged.

AlphaWallet/Tokens/ViewModels/EthCurrencyHelper.swift
```diff
@@ -24,6 +24,14 @@ class EthCurrencyHelper {
         }
     }
 
+    var marketPrice: Double? {
+        if let price = ticker?.price_usd {
```
hboon, line 28: Just this will do, right?

`return ticker?.price_usd`

vladyslav-iosdev: Done.

PR #1985, opened 2020-07-01
Show testnet label in cells if testnet in Wallet tab #1984

### LINKED ISSUE #1984: Show testnet label in cells if testnet in Wallet tab
Like in https://github.com/AlphaWallet/alpha-wallet-ios/issues/1981#issuecomment-652254180

Show that "ROPSTEN" label in the bottom row. We used to have something like that (and we still show it in other screens which haven't been updated yet), so remember to use the same colors because we use it for the token icons also.

Ignore this comment of mine:

> Do you want to move the testnet name upwards? There should be just 1 row, since "340.57 USD" shouldn't be shown; the tokens have no value.

vladyslav-iosdev:
Closes #1984

### DIFF
--- a/AlphaWallet/Tokens/Views/EthTokenViewCell.swift
+++ b/AlphaWallet/Tokens/Views/EthTokenViewCell.swift
@@ -17,12 +17,16 @@ class EthTokenViewCell: UITableViewCell {
     private var viewsWithContent: [UIView] {
         [titleLabel, valuePercentageChangeValueLabel, valuePercentageChangePeriodLabel, marketPriceLabel]
     }
+    private lazy var changeValueContainer: UIView = [marketPriceLabel, valuePercentageChangeValueLabel].asStackView(spacing: 5)
+
     private var tokenIconImageView: TokenImageView = {
         let imageView = TokenImageView()
         imageView.translatesAutoresizingMaskIntoConstraints = false
         return imageView
     }()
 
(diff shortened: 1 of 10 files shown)


AlphaWallet/Tokens/Views/EthTokenViewCell.swift
```diff
@@ -17,12 +17,21 @@ class EthTokenViewCell: UITableViewCell {
...
     private var tokenIconImageView: TokenImageView = {
         let imageView = TokenImageView()
         imageView.translatesAutoresizingMaskIntoConstraints = false
         return imageView
     }()
 
+    private var blockChainTagLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.backgroundColor = .red
```
hboon, line 31: Can you remove this? :)

vladyslav-iosdev: > Can you remove this? :)

yep,)

vladyslav-iosdev: @hboon  Done, you can check it.

vladyslav-iosdev:
> Looks good, except for one thing: the 1st row with balance and token name might be clipped but the blockchain label. Can you verify? Especially the descenders. Maybe just increase the margin by a tiny bit.

@hboon  now should be ok

PR #1997, opened 2020-07-03
Introduce a protocol with default implementation so reusable views like UITableViewCells don't need to hardcode a reusable identifier #1994

### LINKED ISSUE #1994: Introduce a protocol with default implementation so reusable views like UITableViewCells don't need to hardcode a reusable identifier
Add a new file `WithReusableIdentifier.swift`:

```
protocol WithReusableIdentifier {
	static var reusableIdentifier: String { get }
}

extension WithReusableIdentifier {
	static var reusableIdentifier: String {
		String(describing: self)
	}
}

extension UITableViewCell: WithReusableIdentifier {
}
```

and replace something like `AddHideTokensCell.identifier` with `AddHideTokensCell.reusableIdentifier` where `UITableView`s and `UICollectionView`s are used so we end up with:

```
tableView.register(AddHideTokensCell.self, forCellReuseIdentifier: AddHideTokensCell.reusableIdentifier)
```

And we can remove lines like:

```
static let identifier = "AddHideTokensCell"
```

There might be cases where we use more than 1 identifier for a cell class within a table/collection view. I can't remember if there's any of those. If so, just keep those as it is.

We could have abstracted it more to become:

```
tableView.register(AddHideTokensCell.self)
```

But it seems to be too much magic..

vladyslav-iosdev:
Closes #1994

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
@@ -146,10 +146,11 @@ class TokensViewController: UIViewController {
 
         consoleButton.addTarget(self, action: #selector(openConsole), for: .touchUpInside)
 
-        tableView.register(AddHideTokensCell.self, forCellReuseIdentifier: AddHideTokensCell.identifier)
-        tableView.register(FungibleTokenViewCell.self, forCellReuseIdentifier: FungibleTokenViewCell.identifier)
-        tableView.register(EthTokenViewCell.self, forCellReuseIdentifier: EthTokenViewCell.identifier)
-        tableView.register(NonFungibleTokenViewCell.self, forCellReuseIdentifier: NonFungibleTokenViewCell.identifier)
+        tableView.register(AddHideTokensCell.self)
+        tableView.register(FungibleTokenViewCell.self)
+        tableView.register(EthTokenViewCell.self)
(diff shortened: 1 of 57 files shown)


AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -315,7 +315,7 @@ extension TokensViewController: UITableViewDelegate {
 
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
         guard section == 0 else { return nil }
-        let header = tableView.dequeueReusableHeaderFooterView(withIdentifier: TableViewSectionHeader.reuseIdentifier) as? TableViewSectionHeader ?? TableViewSectionHeader(reuseIdentifier: TableViewSectionHeader.reuseIdentifier)
+        let header = tableView.dequeueReusableHeaderFooterView(withIdentifier: TableViewSectionHeader.reusableIdentifier) as? TableViewSectionHeader ?? TableViewSectionHeader(reuseIdentifier: TableViewSectionHeader.reusableIdentifier)
```
hboon, line 318: Can probably abstract this too in a similar manner?

vladyslav-iosdev: yep, np.

PR #1999, opened 2020-07-03
Make Import Wallet screen's buttons and text at the bottom move up and down when keyboard is show/hidden #1993

### LINKED ISSUE #1993: Make Import Wallet screen's buttons and text at the bottom move up and down when keyboard is show/hidden
4 tabs (Seed, Keystore JSON, etc)

Remember to test shorter devices. It might not fit for 5s. If it doesn't, just do a check with `ScreenChecker()` and retain the current behavior. Hopefully works for 6s?

vladyslav-iosdev:
Closes #1993


AlphaWallet/Core/Helpers/KeyboardChecker.swift
```diff
@@ -0,0 +1,61 @@
...
+//  Created by Vladyslav Shepitko on 03.07.2020.
+//
+
+import UIKit
+
+extension UIApplication {
+    var bottomSafeAreaHeight: CGFloat {
+        var bottomSafeAreaHeight: CGFloat = 0
+
+        if #available(iOS 11.0, *) {
```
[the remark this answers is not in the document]
vladyslav-iosdev: ok, my fail, copied from existed app.

AlphaWallet/Core/Helpers/KeyboardChecker.swift
```diff
@@ -0,0 +1,63 @@
...
+    init(viewController: UIViewController, resetHeightDefaultValue: CGFloat = -UIApplication.shared.bottomSafeAreaHeight) {
+        self.viewController = viewController
+        self.resetHeightDefaultValue = resetHeightDefaultValue
+        super.init()
+
+        notificationCenter.addObserver(self, selector: #selector(keyboardWillOpen), name: UIResponder.keyboardWillShowNotification, object: nil)
+        notificationCenter.addObserver(self, selector: #selector(keyboardWillHide), name: UIResponder.keyboardWillHideNotification, object: nil)
+    }
+
+    private var additionalBottonInset: CGFloat {
```
hboon, line 37: `Bottom` :)

vladyslav-iosdev: > So strange. But I thought I commented on this and posted a screenshot and you replied, but I can't find it anymore. Anyway, it is still happening if we toggle the software keyboard on/off (without hitting the Done button). But this could be due to a problem with the original layout? Because I can see it before this PR.
> 
> 

@hboon i found how it happens. Maybe you can help me to resolve this issue? The reason is because in method `keyboardWillHide`, we set keyboard height to `0`. when actual keyboard height  is `114`,  the problem is that i cant detect right keybard height. when keyboard hides, it tells that it height equals `114` even it hides at all from screen.

hboon: I don't know :)

vladyslav-iosdev:
> Notice that when we tap on the textbox with the virtual keyboard off, the bottom margin seems much more than expected:
> 
> 
> 
> Maybe it's retaining the height of the safe margin?

Now should be ok, also tested on small screen devices.

hboon:
@James-Sangalli careful that this sets up the framework for how we do it for a few other screens, e.g #2001

vladyslav-iosdev:
> Ok. Looks good. Let's just get it in first :)
> 
> Just 1 thing to fix. In the "Keystore" tab, can you make the "Import from iCloud/Dropbox/Google Drive" button scroll with the text box? It doesn't fit in iPhone 5s.

it seems like placing `footer Bar` on scroll view will be a bit complicated, i guess easier way is just adjust some heights for narrow screen size

hboon:
> it seems like placing footer Bar on scroll view will be a bit complicated, i guess easier way is just adjust some heights for narrow screen size

But the footer bar can stay there, just move the "Import from iCloud/Dropbox/Google Drive" into the scrollview?

vladyslav-iosdev:
`importKeystoreJsonFromCloudButton` as wel as `importSeedDescriptionLabel` they are snapped to `footerBar` top contraint and move with it when keyboard shows.

PR #2005, opened 2020-07-03
Hide seed phrase word count in Import Wallet screen when the selected tab is not "Seed" #2000

### LINKED ISSUE #2000: Hide seed phrase word count in Import Wallet screen when the selected tab is not "Seed"
eg. notice the "12":

vladyslav-iosdev:
Closes #2000

### DIFF
--- a/AlphaWallet/Wallet/ViewControllers/ImportWalletViewController.swift
+++ b/AlphaWallet/Wallet/ViewControllers/ImportWalletViewController.swift
@@ -25,7 +25,12 @@ class ImportWalletViewController: UIViewController, CanScanQRCode {
     private let roundedBackground = RoundedBackground()
     private let scrollView = UIScrollView()
     private let tabBar = SegmentedControl(titles: ImportWalletViewModel.segmentedControlTitles)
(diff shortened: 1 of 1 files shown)


AlphaWallet/Wallet/ViewControllers/ImportWalletViewController.swift
```diff
@@ -128,13 +130,15 @@ class ImportWalletViewController: UIViewController, CanScanQRCode {
         watchAddressTextField.delegate = self
         watchAddressTextField.returnKeyType = .done
 
+        let col1 = [mnemonicTextView.statusLabel, mnemonicCountLabel].asStackView()
```
hboon, line 133: `row2` instead of `col1` maybe? So we have:

```
[
  mnemonicTextView.label,
  .spacer(height: 4),
  mnemonicTextView,
  .spacer(height: 4),
  row2,
]
```

vladyslav-iosdev: yep you right.

PR #2014, opened 2020-07-07
Update header for native crypto / ERC 20 token screens #2007

### LINKED ISSUE #2007: Update header for native crypto / ERC 20 token screens
This is related to #1995. The part that should change is circled in red:

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/styleguide/components?cseid=5efeeb8d3e258fb78e39e09b

Refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/1995#issuecomment-653405327 to see how the line "Market Price: xxx" changes when tapped. There are 3 states. Start with "Value: xxx", then "Market Price: xxx", then "Available: ", then circle back again. Ignore "Earned so far: xxx" in Zeplin.

Notice that like the cells in Wallet tab, there are at least a few variations:

* native crypto where fiat value is available
* native crypto where fiat value is not available
* tokens, no fiat value
* native crypto or tokens for testnets

Can you see if there are other screens that uses this display or a very similar display? We might want to update those too.

vladyslav-iosdev:
Closes #2007

### DIFF
--- a/AlphaWallet/Transfer/Views/SendHeaderView.swift
+++ b/AlphaWallet/Transfer/Views/SendHeaderView.swift
@@ -3,162 +3,82 @@ import UIKit
 
 protocol SendHeaderViewDelegate: class {
     func didPressViewContractWebPage(inHeaderView: SendHeaderView)
+    func showHideMarketPriceSelected(inHeaderView: SendHeaderView)
 }
 
 class SendHeaderView: UIView {
-    private let background = UIView()
-    private let titleLabel = UILabel()
-    //TODO rename? Button now
-    private let blockchainLabel = UIButton(type: .system)
-    private let blockChainTagLabel = UILabel()
-    private let middleBorder = UIView()
-    private var footerStackView: UIStackView?
-    lazy private var viewsToShowOnlyIfAlternativeAmountsAreAvailable = {
-        return [middleBorder, footerStackView!]
-    }()
-    private let valuePercentageChangeValueLabel = UILabel()
-    private let valuePercentageChangePeriodLabel = UILabel()
-    private let valueChangeLabel = UILabel()
-    private let valueChangeNameLabel = UILabel()
-    private let valueLabel = UILabel()
-    private let valueNameLabel = UILabel()
 
     weak var delegate: SendHeaderViewDelegate?
 
-    override init(frame: CGRect) {
-        super.init(frame: frame)
+    private var tokenIconImageView: TokenImageView = {
+        let imageView = TokenImageView()
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        imageView.isUserInteractionEnabled = true
+        return imageView
+    }()
 
-        background.translatesAutoresizingMaskIntoConstraints = false
-        addSubview(background)
+    private let titleLabel: UILabel = {
+        let label = UILabel()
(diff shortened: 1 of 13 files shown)


AlphaWallet/Transfer/Views/SendHeaderView.swift
```diff
@@ -116,49 +220,30 @@ class SendHeaderView: UIView {
...
 
-        if viewModel.showAlternativeAmount {
-            viewsToShowOnlyIfAlternativeAmountsAreAvailable.showAll()
-        } else {
-            viewsToShowOnlyIfAlternativeAmountsAreAvailable.hideAll()
-        }
-    }
-
-    @objc private func showContractWebPage() {
-        delegate?.didPressViewContractWebPage(inHeaderView: self)
```
hboon, line 162: Sorry, I forgot to mention this. Can you make tapping the token icon open browser at the Etherscan (or blockscout, etc) page showing this token? I think this is the line of code that does that. Before the PR you can tap on the words "Ethereum Blockchain" to trigger this.

vladyslav-iosdev:
> Looks good. I understand it's a WIP. There shouldn't be "Market Price" for ERC20s (which don't have a price/ticker):
> 
> 
> 
> Also tap to cycle between "Market Price" etc when it's available.

yep, its wip.

hboon:
@vladyslav-iosdev is this good for review?

vladyslav-iosdev:
> @vladyslav-iosdev is this good for review?

yep

vladyslav-iosdev:
> * [ ]  Discover token (0x8C0Edb69eBF038BA0c7a4873E40fC09725064c2e on mainnet) shouldn't show `Value: xxx`
> *

this one detected as ERC20Token
here as i undarstand right, old map for displaying states
native crypto where fiat value is available         -   marketPriceLabel
native crypto where fiat value is not available   -   earnedSoFarLabel
tokens, no fiat value                                             -   valueLabel
native crypto or tokens for testnets                   -   blockChainTagLabel

and now it should be like this, right?

fiat value is available                                            -   marketPriceLabel
tokens, no fiat value                                             -   valueLabel
native crypto or tokens for testnets                    -   blockChainTagLabel

vladyslav-iosdev:
> Oh, I get where we misunderstood each other now :)
> 
> * [ ]  Here's how it should be:
> 
> if native crypto where fiat value (i.e price/ticker) is available (so just ETH for mainnet and xDai):
> 
> 1. show value
> 2. tap on value and switch to show Market Price
> 3. tap on the market price to show value again, i.e. looping back to (1)
> 
> (Ignore "Available", "Earned so Far". I said to include "Available" previously, but just realised we don't have a value for that yet)
> 
> If fiat value is not available:
> 
> A. If it's a testnet (`RPCServer.isTestnet`), show the blockchain label
> B. If it's not a testnet, hide the row completely. So this will be basically be all ERC20 tokens on non-testnets
> 
> * [ ]  Showing the token name and symbol at the navigation bar is good. Can you remove the name below the token icon? It seems to be showing up for ERC20 (but not native cryto). Hide it for both cases, only show the symbol. So instead of "1 AlphaWallet Discover (ALP)", only show "1 ALP", and for native crypto too: instead of "0.0997 (ETH)", just show "0.90997 ETH"

now should be ok.

hboon:

PR #2019, opened 2020-07-08
Moved "Verified by xxx" into navigation bar #2018

### LINKED ISSUE #2018: Moved "Verified by xxx" into navigation bar
After #2014 is merged

The current version:

Simulator Screen Shot - iPhone X - 2020-07-08 at 11 40 56

It should look like in #1991 (it can be green or red depending on whether it's verified). There's a few helper functions that help to show it.

vladyslav-iosdev:
Closes #2018

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/TokenViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/TokenViewController.swift
@@ -3,6 +3,7 @@
 import Foundation
 import UIKit
 import BigInt
+import PromiseKit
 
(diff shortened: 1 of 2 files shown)


vladyslav-iosdev:
> The "Verified by xxx" button seems to be showing up incorrectly. For example, it appears for both the xDai token and Ether (mainnet) token. It should appear only for the xDai token, but not for the Ether token

yep, sorry its my fail, i copied from `@objc protocol VerifiableStatusViewController: CanOpenURL2 {` but forgot to update it

hboon:

PR #2032, opened 2020-07-15
add warning message to prevent users sharing their seed phrase with scammers

James-Sangalli:
Similar to https://github.com/AlphaWallet/alpha-wallet-android/pull/1510


James-Sangalli:
@hboon actually neither the original or the amended fit on the iphone 5s screen

James-Sangalli:
https://github.com/AlphaWallet/alpha-wallet-ios/issues/2033

Issue #1826, opened 2020-05-23
Add "My QR" button to show wallet QR code in QR code scanner screen


hboon:
Study #1814

vladyslav-iosdev:
@hboon My QR button always should present in QR code scanner screen, or not? (for example if we do scan qr code from send screen, do we need show this button)?

hboon:
@vladyslav-iosdev yes, always show it.

hboon:
Closed with #2037

PR #2037, opened 2020-07-15
All-encompassing QR code scanner #1723

### LINKED ISSUE #1723: All-encompassing QR code scanner
So that QR code scanner can be available in various screens and "jump" to the right screen depending on what is scanned instead of the ones we have now which only allows users to scan what is supported in the current screen.

Let’s make the scanner functionality configurable with options - for handling EIP681, WalletConnect etc -, mirroring how UIImagePickerController allows us to choose from photo, library, all, etc.

There’s a Zeplin screen which has the button, but let’s assume it’s a button available in the Wallets tab navigation bar (because: see Out of Scope below). 

Out of scope 
—-
When this is available, we should probably also make it available in the quick action list accessibly by tapping and holding the app icon in the iOS home screen.

vladyslav-iosdev:
Closes  #1723, #1826


vladyslav-iosdev:
@hboon also can how can i determin if detected address is not Ethereum wallet address?

hboon:
>@hboon also can how can i determin if detected address is not Ethereum wallet address? ...

@vladyslav-iosdev you mean if it’s a contract address and not wallet? We can’t. We can only do a simple length and string check to make sure it’s an Ethereum address. While we might do something smarter like querying the address to figure out if it’s a token/contract, sometimes the user might really want to watch it like a “wallet”, so can’t really be sure. Also a contract might be a smart contract wallet :p

So have to present all options that applies to Ethereum addresses. I haven’t checked the code yet, but remember the comment I made about providing configurable “source” options like UIImagePicker. We probably will use this same code but with specific options in existing screens like Send screen where is address to send to or EIP681 but never prompt the user to add a custom contract. Similarly in the import wallet screen.

hboon:
@vladyslav-iosdev are you waiting for me to review this?

vladyslav-iosdev:
> @vladyslav-iosdev are you waiting for me to review this?

actually i dont't know, i worked on it to long ago, and i don't remember what, not finished here yet. but anyway u can check it.

hboon:
@vladyslav-iosdev sorry,  my bad. I'll check it out.

AlphaWallet/AlphaWallet.entitlements
```diff
@@ -2,24 +2,24 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
```
hboon, line 4: Can you see if you can restore this file and not generate a diff?

hboon (changes requested):
So the universal scanner (for this PR at least) is only activated in the Wallet tab, right? The scanner functionality in other screens remain unchanged?

- [ ] When the QR code is for an EIP681 payment link like this:

```
"ethereum:0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359/transfer?address=0x007bEe82BDd9e866b2bd114780a47f2261C684E3&uint256=1.3e18"
```

Rather than prompting the user to open in browser, it should open TokenViewController for that token (fetch the contract data first, like user is manually adding a token, ala `NewTokenViewController` if it's not already in `TokenObject`) and then bring up the send screen.

- [ ] The number of lines in the localization files are not the same, can you see if it's because of this PR or is it a problem in `master`?

- [ ] When handling `.address`, we can check if the address is already in the database. If it is, we can omit `.addCustomToken`.

- [ ] When handling `.address`, can you add 1 more action: "Open in Etherscan"? Hardcoded to be for mainnet (or add it in a future PR, up to you)

- [ ] In the "Select Asset" screen for `.sendToAddress`, I think we'll have to only include ERC20 tokens for now. Can you create an issue to follow up to support other token types? Their transfer flow/screen is different (partly because they use different smart contract functions)

vladyslav-iosdev:
> So the universal scanner (for this PR at least) is only activated in the Wallet tab, right? The scanner functionality in other screens remain unchanged?

yes, right.

hboon:
> > So the universal scanner (for this PR at least) is only activated in the Wallet tab, right? The scanner functionality in other screens remain unchanged?
> 
> yes, right.

That's great. More self-contained :) Thanks.

AlphaWallet/AlphaWallet.entitlements
vladyslav-iosdev: > Can you see if you can restore this file and not generate a diff?

it seems like i fixed it, but i am not sure.

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -8,12 +8,59 @@ protocol ScanQRCodeCoordinatorDelegate: class {
...
+typealias WCURL = String
+
+private enum ScanQRCodeResolution {
+    case address(AlphaWallet.Address)
+    case walletConnect(WCURL)
+    case other(String)
+    case url(URL)
+
+    init(rawValue: String) {
+        if let address = AlphaWallet.Address(string: rawValue) {
```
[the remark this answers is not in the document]
vladyslav-iosdev: > Should have something like this:
> 
> ```
> let rawValue = rawValue.trimmed
> ```
> 
> so addresses like "0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359\n" is handled. I don't know how I generated it, but I guess it might be common :)

done

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -51,26 +120,200 @@ final class ScanQRCodeCoordinator: NSObject, Coordinator {
...
+            switch value {
+            case .address(let contract):
+                switch configuration {
+                case .resolution(let tokensDatastores, _):
+                    let actions: [ScanQRCodeAction]
+                    //NOTE: or meybe we need pass though all servers?
+                    guard let tokensDatastore = tokensDatastores.first(where: { $0.server == rpcServer }) else { return }
+
+                    //I guess if we have token, we shouldn't be able to send to it, or we should?
+                    if tokensDatastore.token(forContract: contract) != nil {
```
hboon, line 142: We should still allow `.sendToAddress` and `.watchWallet` because for the former, some token-holding contracts (I think Dai?) works by getting you to send ETH to it. For the latter, we might just want to look at transactions for the contract.

AlphaWallet/Browser/Coordinators/ScanQRCodeCoordinator.swift
```diff
@@ -2,18 +2,83 @@
...
+    var title: String {
+        switch self {
+        case .sendToAddress:
+            return R.string.localizable.qrCodeSendToAddressTitle()
+        case .addCustomToken:
+            return R.string.localizable.qrCodeAddCustomTokenTitle()
+        case .watchWallet:
+            return R.string.localizable.qrCodeWatchWalletTitle()
+        case .openInEtherscan:
+            return "Open in Etherscan"
```
hboon, line 35: Remember to localize

AlphaWallet/Tokens/ViewModels/SelectAssetViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    init(filterTokensCoordinator: FilterTokensCoordinator, tokens: [TokenObject], tickers: [RPCServer: [AlphaWallet.Address: CoinTicker]], filter: WalletFilter) {
+        self.filterTokensCoordinator = filterTokensCoordinator
+        self.tokens = tokens
+        self.tickers = tickers
+        self.filter = filter
+    }
+
+    private func filteredAndSortedTokens() -> [TokenObject] {
+        let displayedTokens = filterTokensCoordinator.filterTokens(tokens: tokens, filter: .type(.erc20))
```
hboon, line 62: Should the `.type(.erc20)` be replaced by the `filter` property instead?

hboon (changes requested):
Can you try with this EIP681 link:

ethereum:0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359/transfer?address=0x007bEe82BDd9e866b2bd114780a47f2261C684E3&uint256=1.3e18

If I scan it with the universal scanner, it shows the amount to transfer as `1.3e18`. but it should be 1.3. If you scan the same link when in the DAI card, it shows `1.3` as expected.

AlphaWallet/Tokens/ViewModels/SelectAssetViewModel.swift
vladyslav-iosdev: > Should the `.type(.erc20)` be replaced by the `filter` property instead?

sorry, u right.

vladyslav-iosdev:
> Can you try with this EIP681 link:
> 
> ethereum:0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359/transfer?address=0x007bEe82BDd9e866b2bd114780a47f2261C684E3&uint256=1.3e18
> 
> If I scan it with the universal scanner, it shows the amount to transfer as `1.3e18`. but it should be 1.3. If you scan the same link when in the DAI card, it shows `1.3` as expected.

@hboon done, now should be ok.

hboon (changes requested):
Looks good! A couple of things:

- [x] When scan an address-only QR code to add custom token, we should hide the tab bar when the add token screen is pushed onto the navigation controller (see screenshot)
- [x] Select Asset when scanning and choosing to send to an address should include the native crypto token (ETH, etc) in it
- [ ] Need to update the Podfile to point to the correct commit after https://github.com/AlphaWallet/QRCodeReaderViewController/pull/2 is merged before merging this PR
- [x] Squash into single commit

hboon (changes requested):
Looks good, but seems to be failing a test. Can you help fix it? After that we just need to wait for https://github.com/AlphaWallet/QRCodeReaderViewController/pull/2 to be ready.

vladyslav-iosdev:
> Looks good, but seems to be failing a test. Can you help fix it? After that we just need to wait for [AlphaWallet/QRCodeReaderViewController#2](https://github.com/AlphaWallet/QRCodeReaderViewController/pull/2) to be ready.

damn, right, i missed about test. i will return to it after finish bartercard task.

vladyslav-iosdev:
> Looks good, but seems to be failing a test. Can you help fix it? After that we just need to wait for [AlphaWallet/QRCodeReaderViewController#2](https://github.com/AlphaWallet/QRCodeReaderViewController/pull/2) to be ready.

@hboon  i pushed tests fixes.

hboon:
@vladyslav-iosdev I think the commit the Podfile should point to should be 30d1a2a7d167d0d207ae0ae3a4d81bcf473d7a65 instead, according to https://github.com/AlphaWallet/QRCodeReaderViewController/commits/alphawallet?

vladyslav-iosdev:
> @vladyslav-iosdev I think the commit the Podfile should point to should be 30d1a2a7d167d0d207ae0ae3a4d81bcf473d7a65 instead, according to https://github.com/AlphaWallet/QRCodeReaderViewController/commits/alphawallet?

oh, yep, u right. done.

hboon:
Sorry I missed this:

Can you remove this part `:branch=>'alphawallet'` from the Podfile? I don't know what would happen if we specify both the branch and commit hash.

vladyslav-iosdev:
> Can you remove this part `:branch=>'alphawallet'` from the Podfile? I don't know what would happen if we specify both the branch and commit hash.

done

hboon (approved):

PR #2062, opened 2020-07-22
after marriage proposal PR, redirect the user to aETH token if the current token is ETH and the name of the action is `depositAAVE` #2056

### LINKED ISSUE #2056: after marriage proposal PR, redirect the user to aETH token if the current token is ETH and the name of the action is `depositAAVE`
Observe this file https://github.com/AlphaWallet/TokenScript-Examples/blob/train-leaves/examples/action/ETH/ETH.xml
go to line 71 and see the hardcoded name is `depositeAAVE`

-> only PR to `train-leaves`

vladyslav-iosdev:
Closes #2056


James-Sangalli:
Dropped

PR #2086, opened 2020-08-03
Fix: crash when opening TokenScript views

hboon:
Any of this:

A. Tap on a token with TokenScript file with item-views. eg. ENS
B. Tap on a token with action-only TokenScript and tap on an action button e.g. xDai


James-Sangalli:
> Any of this:
> 
> A. Tap on a token with TokenScript file with item-views. eg. ENS
> B. Tap on a token with action-only TokenScript and tap on an action button e.g. xDai

@hboon I cannot reproduce any of these

hboon:
@James-Sangalli see if you can do it with iOS simulator 12.4. If it doesn't crash, try going back (so the views get destroyed).

hboon:
Maybe an optimisation. Strange though. But definitely a bug.

PR #2092, opened 2020-08-04
Expand on #2014 and automatically cycle through the value/market price etc every 5 secs #2082

### LINKED ISSUE #2082: Expand on #2014 and automatically cycle through the value/market price etc every 5 secs
New PR.

vladyslav-iosdev:
Closes #2082


hboon:
@vladyslav-iosdev does it handle going away correctly so it's deallocated without holding a strong reference?

PR #2094, opened 2020-08-04
Lack confirmation message when a developer opened/imported a TSML file into αW #2079

### LINKED ISSUE #2079: Lack confirmation message when a developer opened/imported a TSML file into αW
A developer has to go to advanced settings to see if the new file is recognised and in use - if the new file has the same name as one already in use there will be no clue if the import/open is successful.

currently:

1. open a tsml file (e.g. received from Telegram) with AlphaWallet
2. AlphaWallet launches innocently as αWallet but secretly takes the file in.

expected:

1. open a tsml file (e.g. received from Telegram) with AlphaWallet
2. AlphaWallet launches with a message like "xxx file imported with no error" or give some error.

@AW-STJ is following the process.

vladyslav-iosdev:
Closes #2079

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -357,12 +357,17 @@ extension AppCoordinator: CustomUrlSchemeCoordinatorDelegate {
 }
 
 extension AppCoordinator: AssetDefinitionStoreCoordinatorDelegate {
+
     func show(error: Error, for viewController: AssetDefinitionStoreCoordinator) {
         inCoordinator?.show(error: error)
     }
 
-    func addedTokenScript(forContract contract: AlphaWallet.Address, forServer server: RPCServer) {
+    func addedTokenScript(forContract contract: AlphaWallet.Address, forServer server: RPCServer, destinationFileInUse: Bool, filename: String) {
         inCoordinator?.addImported(contract: contract, forServer: server)
+
+        if !destinationFileInUse {
+            inCoordinator?.show(openedURL: filename)
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


vladyslav-iosdev:
I am not sure about implementation, need more details.

AlphaWallet/AppCoordinator.swift
```diff
@@ -357,6 +357,11 @@ extension AppCoordinator: CustomUrlSchemeCoordinatorDelegate {
 }
 
 extension AppCoordinator: AssetDefinitionStoreCoordinatorDelegate {
+
+    func didHandleOpenURL(in coordinator: AssetDefinitionStoreCoordinator, with result: OpenURLResult) {
```
hboon, line 361: It seems like there is already a `show(error:for:)` function in `AssetDefinitionStoreCoordinatorDelegate` (currently not called by the delegate) which we can call when there's an error.

There's also `addedTokenScript(forContract:, forServer:)` for when the override is successful.

So you can probably remove `didHandleOpenURL(in:with:)`, use those 2 instead?

AlphaWallet/TokenScriptClient/Coordinators/AssetDefinitionStoreCoordinator.swift
```diff
@@ -128,7 +166,8 @@ class AssetDefinitionStoreCoordinator: Coordinator {
             isTokenScriptOrXml = true
         case .unsupportedTokenScriptVersion:
             try? FileManager.default.removeItem(at: url)
-            delegate?.show(error: SchemaCheckError(msg: R.string.localizable.tokenScriptNotSupportedSchemaError()), for: self)
+            delegate?.didHandleOpenURL(in: self, with: .failure(OpenURLError.unsupportedTokenScriptVersion))
+
             return true
         case .unknownXml:
             try? FileManager.default.removeItem(at: url)
```
hboon, line 173: We'll have to see if `.unknownXML` and `.others` get triggered for XML files. We can AirDrop .css, files etc, but we really only want to show error (or success) messages for .xml or .tsml files.

vladyslav-iosdev: so for cases `.unknownXML` and `.others` i need to display error?

hboon: Leave it for now. Sounds complicated. Most important is to show a confirmation when it works.

Issue #1849, opened 2020-05-29
Research if/how to test if a private key (which is marked as user presence required) written into the Secure Enclave can tested for existence without prompting for user presence check


hboon:
Referring to the flag `SecAccessControlCreateFlags.userPresence`.

The public-private key pair is created in `SecureEnclave.createPrivateKey`.

This will require a device (not simulator) with Touch ID or Face ID to verify.

Just a snippet will do. It doesn't have to be integrated into the code base.

hboon:
Bump.

vladyslav-iosdev:
> @vladyslav-iosdev Bump.

@hboon i researched but find nothing according this issue

hboon:
> @hboon i researched but find nothing according this issue

As in, it's probably not possible?

hboon:
Good news, I asked someone who said this:

> I can confirm. I have code that does SecItemCopyMatching to check if a key already exists and it doesn’t prompt for biometry until you actually call something like SecKeyCreateEncryptedData

Can you help to dig further? :)

vladyslav-iosdev:
> Good news, I asked someone who said this:
> 
> > I can confirm. I have code that does SecItemCopyMatching to check if a key already exists and it doesn’t prompt for biometry until you actually call something like SecKeyCreateEncryptedData
> 
> Can you help to dig further? :)

ok, i will check it

vladyslav-iosdev:
@hboon can you try with this piece of code:
```
public func contains(_ key: String) throws -> Bool {
        var query = options.query()
        query[AttributeAccount] = key
        query[UseAuthenticationUI] = UseAuthenticationUIFail
        
        let status = SecItemCopyMatching(query as CFDictionary, nil)
        switch status {
        case errSecSuccess, errSecInteractionNotAllowed:
            return true
        case errSecItemNotFound:
            return false
        default:
            throw securityError(status: status)
        }
    }
```
i found it here:
https://github.com/kishikawakatsumi/KeychainAccess/issues/159

hboon:
@vladyslav-iosdev thanks! Will try it out.

FYI. The painful part of this is, in order to be sure, I have wipe device, restore from iCloud back to verify the key is no longer there.

PR #2097, opened 2020-08-10
Add in a check if keys for all wallets are not accessible on startup #2048

### LINKED ISSUE #2048: Add in a check if keys for all wallets are not accessible on startup
Probably have to iterate this a few times. Ideally something like:

1. Launch (not resume) app
2. Check keys are accessible
3. Inform user and "probably" go to the create new wallet screen

This is a bit tricky because we want to make sure the keys are really not accessible anymore (due to migration to a new device).

Related to: #1849

vladyslav-iosdev:
Closes #2048, #1849

### DIFF
--- a/AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
+++ b/AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
@@ -108,9 +108,13 @@ class AccountsViewController: UIViewController {
 
     private func delete(account: Wallet) {
         navigationController?.displayLoading(text: R.string.localizable.deleting())
-        keystore.delete(wallet: account) { [weak self] result in
+        let result = keystore.delete(wallet: account)
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.4) { [weak self] in
             guard let strongSelf = self else { return }
+
             strongSelf.navigationController?.hideLoading()
+
             switch result {
             case .success:
                 strongSelf.fetch()
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -82,12 +82,29 @@ class AppCoordinator: NSObject, Coordinator {
             }
         }
     }
+    //NOTE: This function is using to make sure that wallets in user defaults will be removed after restoring backup from iCloud. Realm files don't backup to iCloud but user defaults does backed up.
+    private func removeWalletsIfRealmFilesMissed() {
+        for wallet in keystore.wallets {
+            let migration = MigrationInitializer(account: wallet)
+
+            guard let path = migration.config.fileURL else { continue }
+
+            //NOTE: make sure realm files exists, if not then delete this wallets from user defaults.
+            if FileManager.default.fileExists(atPath: path.path) {
+                //no op
+            } else {
+                _ = keystore.delete(wallet: wallet)
+            }
+        }
+    }
 
     //This function exist to handle what we think is a rare (but hard to reproduce) occurrence that NSUserDefaults are not accessible for a short while during startup. If that happens, we delay the "launch" and check again. If the app is killed by the iOS launch time watchdog, so be it. Better than to let the user create a wallet and wipe the list of wallets and lose access
     private func startImpl() -> Bool {
         if hasRealmDatabasesForWallet && !keystore.hasWallets && !isRunningTests() {
             return false
         }
+
+        removeWalletsIfRealmFilesMissed()
 
         setupAnalytics()
         window.rootViewController = navigationController
@@ -98,7 +115,7 @@ class AppCoordinator: NSObject, Coordinator {
         resetToWelcomeScreen()
         setupAssetDefinitionStoreCoordinator()
         migrateToStoringRawPrivateKeysInKeychain()
-
+        
         if keystore.hasWallets {
             showTransactions(for: keystore.recentlyUsedWallet ?? keystore.wallets.first!)
         } else {
--- a/AlphaWallet/KeyManagement/EtherKeystore.swift
+++ b/AlphaWallet/KeyManagement/EtherKeystore.swift
@@ -52,7 +52,6 @@ open class EtherKeystore: Keystore {
     private let defaultKeychainAccessUserPresenceRequired: KeychainSwiftAccessOptions = .accessibleWhenUnlockedThisDeviceOnly(userPresenceRequired: true)
     private let defaultKeychainAccessUserPresenceNotRequired: KeychainSwiftAccessOptions = .accessibleWhenUnlockedThisDeviceOnly(userPresenceRequired: false)
     private let userDefaults: UserDefaults
-
     private var watchAddresses: [String] {
         set {
             let data = NSKeyedArchiver.archivedData(withRootObject: newValue)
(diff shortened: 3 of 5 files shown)


vladyslav-iosdev:
@hboon for me its not clear why do we need to paste it in MigrationInitializator.perform()
        i guess we need to perform it before we call
        if let wallet = keystore.recentlyUsedWallet ?? keystore.wallets.first, keystore.hasWallets {
            showTransactions(for: wallet)
        } else {
            resetToWelcomeScreen()
        }
        because wee need to remove unvailable wallets, and only then make decision what to do next. Or i mistake?

hboon:
> @hboon for me its not clear why do we need to paste it in MigrationInitializator.perform(). I guess we need to > perform it before we call.
> if let wallet = keystore.recentlyUsedWallet ?? keystore.wallets.first, keystore.hasWallets { showTransactions(for: > wallet) } else { resetToWelcomeScreen() }

The reason I suggested we put it in MigrationInitializer.perform() is so it it's easier to locate it. i.e. it is a kind of migration.

You are right, we should do it before this check ^. How about a static func in `MigrationInitializer`? It'll still be located there but called at the right place?

Also what about this? Simpler? :

> I just realised there's another way to fix the problem: since the Realm database files are not backed up too, when the app launch, run through the 3 lists of wallets and check if the Realm database file exist, if not, remove that wallet from the list).

vladyslav-iosdev:
Also i got a question, why do we call this function async? func `delete(wallet: Wallet, completion: @escaping (Result<Void, KeystoreError>) -> Void)` it called piece of code, `!UIApplication.shared.isProtectedDataAvailable` in background thread, but this one should be called only in main thread.

vladyslav-iosdev:
> @hboon for me its not clear why do we need to paste it in MigrationInitializator.perform(). I guess we need to > perform it before we call.
> if let wallet = keystore.recentlyUsedWallet ?? keystore.wallets.first, keystore.hasWallets { showTransactions(for: > wallet) } else { resetToWelcomeScreen() }
> 
> The reason I suggested we put it in MigrationInitializer.perform() is so it it's easier to locate it. i.e. it is a kind of migration.
> 
> You are right, we should do it before this check ^. How about a static func in MigrationInitializer? It'll still be located there but called at the right place?
> 
> Also what about this? Simpler? :
> 
> I just realised there's another way to fix the problem: since the Realm database files are not backed up too, when the app launch, run through the 3 lists of wallets and check if the Realm database file exist, if not, remove that wallet from the list).

yep, i realized that it's easier to implement.

hboon:
> Also i got a question, why do we call this function async? func `delete(wallet: Wallet, completion: @escaping (Result<Void, KeystoreError>) -> Void)` it called piece of code, `!UIApplication.shared.isProtectedDataAvailable` in background thread, but this one should be called only in main thread.

Do you mean why there is an async here in `EtherKeystore`?

```
    func delete(wallet: Wallet, completion: @escaping (Result<Void, KeystoreError>) -> Void) {
        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
            guard let strongSelf = self else {
```

I don't know. We inherited that code :) My guess is the developer felt/thought that the deleting work takes a bit of time and didn't want to block the UI.

Where does it call `UIApplication.shared.isProtectedDataAvailable` though?

hboon:
> yep, i realized that it's easier to implement.

Ok, would you think about it a bit and implement it if it's good? But can you also verify that the Realm database is created immediately when the user create/import/watch a database? This is to guard against this flow:

1. Create wallet (are the Realm database files created immediately?)
2. Immediately kills app (or low battery and iOS kills it)
3. Launch app
4. New code checks that Realm database is not there, delete the wallet address from the list (and thus losing access to the key)

vladyslav-iosdev:
> Where does it call `UIApplication.shared.isProtectedDataAvailable` though?

it get called in `init` method
`(try? LegacyFileBasedKeystore(analyticsCoordinator: analyticsCoordinator))?.delete(wallet: wallet) `

hboon:
> > Where does it call `UIApplication.shared.isProtectedDataAvailable` though?
> 
> it get called in `init` method
> `(try? LegacyFileBasedKeystore(analyticsCoordinator: analyticsCoordinator))?.delete(wallet: wallet)`

Oh. That's no good. Calling `UIApplication.shared.isProtectedDataAvailable` on a background thread seems to be harmless for now although the main thread checker complains about it.. Maybe wrap this line with a  dispatch to main in a separate PR?

> `(try? LegacyFileBasedKeystore(analyticsCoordinator: analyticsCoordinator))?.delete(wallet: wallet)`

Not the best approach, but `LegacyFileBasedKeystore` shouldn't matter much anymore since we have moved away from storing them that way for more than a year:

```
$ gl -S 'LegacyFileBasedKeystore'
commit 1e6a51b7feba2a401cf0cf56a0192d1a97a469ca
Date:   2019-07-04 16:29:35 +0800

    Support HD wallets, only using account 0, as well as raw private keys 
```

vladyslav-iosdev:
> Oh. That's no good. Calling `UIApplication.shared.isProtectedDataAvailable` on a background thread seems to be harmless for now although the main thread checker complains about it.. Maybe wrap this line with a dispatch to main in a separate PR?
> 
> > `(try? LegacyFileBasedKeystore(analyticsCoordinator: analyticsCoordinator))?.delete(wallet: wallet)`
> 
> Not the best approach, but `LegacyFileBasedKeystore` shouldn't matter much anymore since we have moved away from storing them that way for more than a year:

i guess it will not be a problem if we will create wallet/delete wallet in main instead of background thread. (backround thread improve nothing here, as i understand)

hboon:
> i guess it will not be a problem if we will create wallet/delete wallet in main instead of background thread. (backround thread improve nothing here, as i understand)

Hmm.. makes sense. Try? Remember not to change anything else. This is delicate.

vladyslav-iosdev:
> > i guess it will not be a problem if we will create wallet/delete wallet in main instead of background thread. (backround thread improve nothing here, as i understand)
> 
> Hmm.. makes sense. Try? Remember not to change anything else. This is delicate.

yep, ok! u know i just checked it, and found that creating wallet sync cause not a big issue, (but anyway, its issue), that wallet creates too fast now, and we don't see creating wallet indicator. adding here some delay make no sence because. Does it make sence create wallet sync instead of async?

hboon:
> Does it make sence create wallet sync instead of async?

No. Since we show that green success overlay upon creation, we don't need the creating wallet indicator if it's fast and synchronous (or if it's async and still fast).

vladyslav-iosdev:
> No. Since we show that green success overlay upon creation, we don't need the creating wallet indicator if it's fast and synchronous (or if it's async and still fast).

ok, i didn't touch it, and left as it is.

hboon:
Pushed after rebase master.

hboon (approved):

hboon:
@James-Sangalli this looks good to me, Would you help to take a look at this too? This describes the problem: https://github.com/AlphaWallet/alpha-wallet-ios/issues/2048#issuecomment-671774855

hboon:
@James-Sangalli bump

hboon:
Rebased master before doing a check again, to be sure.

PR #2100, opened 2020-08-14
set gas limit from node rather than hardcoding for token transactions

### LINKED ISSUE #2046: Built 3.05(346) ERC token transactions gas limit is default at 1000000
it should read from the node.

James-Sangalli:
Fixes #2046 

The problem was 3 things:

1. The data was not yet included in the configurator and therefore couldn't estimate properly
2. the value param was encoded incorrectly, running BigInt.description.hexEncoded produced 0x30 on a 0 value
3. The value param is used incorrectly, for erc20 transfers the value is the amount of tokens being sent instead of the attached native currency amount, doing so can and will produce an error because erc20 transfers that include a value are often reverted in the contract.

Created https://github.com/AlphaWallet/alpha-wallet-ios/issues/2101 as the code base is not good here.

### DIFF
--- a/AlphaWallet/EtherClient/Requests/EstimateGasRequest.swift
+++ b/AlphaWallet/EtherClient/Requests/EstimateGasRequest.swift
@@ -21,7 +21,7 @@ struct EstimateGasRequest: JSONRPCKit.Request {
             [
                 "from": from.description,
                 "to": to?.description ?? "",
-                "value": value.description.hexEncoded,
+                "value": "0x" + String(value, radix: 16) ?? "0x0",
                 "data": data.hexEncoded,
             ],
         ]
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -85,10 +85,18 @@ class TransactionConfigurator {
                 return token.contractAddress
             }
         }()
+        //TODO transaction.value should only ever be the attached native currency, not the erc20 amount as that is included in the data
+        let value: BigInt = {
+            switch transaction.transferType {
+            case .nativeCryptocurrency, .dapp, .ERC875TokenOrder: return transaction.value
+            case .ERC20Token, .ERC721Token, .ERC721ForTicketToken, .ERC875Token:
+                return 0;
+            }
+        }()
         let request = EstimateGasRequest(
             from: session.account.address,
             to: to,
-            value: transaction.value,
+            value: value,
             data: configuration.data
         )
         Session.send(EtherServiceRequest(server: session.server, batch: BatchFactory().create(request))) { [weak self] result in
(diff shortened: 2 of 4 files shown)


AlphaWallet/EtherClient/Requests/EstimateGasRequest.swift
```diff
@@ -21,7 +21,7 @@ struct EstimateGasRequest: JSONRPCKit.Request {
             [
                 "from": from.description,
                 "to": to?.description ?? "",
-                "value": value.description.hexEncoded,
+                "value": "0x" + String(value, radix: 16) ?? "0x0",
```
hboon, line 24: I think `String(_,radix:)` here will never be nil, so we can just do:

```
String(value, radix: 16).hexEncoded
```

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -85,10 +85,18 @@ class TransactionConfigurator {
                 return token.contractAddress
             }
         }()
+        //TODO transaction.value should only ever be the attached native currency, not the erc20 amount as that is included in the data
```
hboon, line 88: Not a TODO, but a comment for the "why" (which is good) right?

hboon: Do you want to extract it out into a property in `TransferType`?

So you can write here:

```
        let value: BigInt = transaction.nativeCryptoValue
```

But isn't `transaction.value` the native crypto value even if it's an ERC 20 transfer? The token amount is encoded into `data`? Or is it because the gas estimator only looks at the ETH value even if the data is not 0x0? (e.g. smart contract call — eg. ERC20 transfer — which also pays ETH) and we are trying to workaround the estimator?

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -278,6 +275,10 @@ class TransactionConfigurator {
                 completion(.failure(AnyError(Web3Error(description: "malformed tx"))))
             }
         }
+        /* the node can provide reliable gas limit estimates, this prevents running out of gas or defaulting to an
+        inappropriately high gas limit. This can also be an issue for native transfers which are 21k if send to an EOA
+        address but may be higher if sent to a contract address. */
+        estimateGasLimit()
```
hboon, line 281: If we move it here, and without checking against `gasLimitNotSet`, will this override the user's setting?

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
James-Sangalli: No, it's a TODO because it is incorrectly placing the erc20 amount in there.

AlphaWallet/EtherClient/Requests/EstimateGasRequest.swift
James-Sangalli: Cannot, if the string value is "0" it will convert it to 0x30 because 0x30 means 0 on the ascii chart...

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
hboon: Ah. I see. How did it ever work! Aargh.

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
James-Sangalli: No, it is set after the function call

PR #2104, opened 2020-08-17
allow zero value native transactions so data can be attached

### LINKED ISSUE #2098: Allow 0 amount TX in TX Advance Setting page
Allow 0 amount TX in TX Advance Setting page. Then advanced user can call contract with manually input the TX date.

James-Sangalli:
Closes #2098

### DIFF
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -283,8 +283,15 @@ class SendViewController: UIViewController, CanScanQRCode {
         let input = targetAddressTextField.value.trimmed
         targetAddressTextField.errorState = .none
         amountTextField.errorState = .none
-
-        guard let value = viewModel.validatedAmount(value: amountTextField.ethCost) else {
+        let checkIfGreaterThanZero: Bool
+        // allow users to input zero on native transactions as they may want to send custom data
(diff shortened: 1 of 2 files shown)


AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -283,8 +283,14 @@ class SendViewController: UIViewController, CanScanQRCode {
         let input = targetAddressTextField.value.trimmed
         targetAddressTextField.errorState = .none
         amountTextField.errorState = .none
-
-        guard let value = viewModel.validatedAmount(value: amountTextField.ethCost) else {
+        let checkIfGreaterThenZero: Bool
+        // allow users to input zero on native transactions as they may want to send custom data
+        if case .nativeCryptocurrency = transferType {
```
hboon, line 288: Can you replace the `if-case` with an explicit `switch-case`? Otherwise it always comes back to bite us.

AlphaWallet/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -283,8 +283,14 @@ class SendViewController: UIViewController, CanScanQRCode {
...
+            checkIfGreaterThenZero = false
```
hboon, line 289: "Then" -> "Than"

PR #2109, opened 2020-08-19
Activity. Disabled with feature flag

hboon:
WIP for Activity but let's merge it.

The functionality is disabled by the feature flag `Features.isActivityEnabled`.

Work on Activity will continue on a future PR.

When `Features.isActivityEnabled` is set to `true`, these becomes available:

* Base (overridable) TokenScript file for ERC20 and ERC721 is supported
* Native crypto support for send and receive activity item-views
* Activities appear in the Activities tab

The PR is fairly big. Here are the important parts:

EventActivity — The database object
EventsActivityDataStore — Access to the database
EventSourceCoordinatorForActivities — this fetches the events that will be mapped to activities
ActivitiesCoordinator — displays the Activities tab
ActivitiesViewController — the Activities tab. Shows the TokenScript activity `item-view`(s) and transactions
ActivityViewController — A single activity on the screen. Shows the TokenScript activity `view`

To try it out, set `Features.isActivityEnabled`. Probably have to modify `Constants.erc20ContractsSupportingActivities` too

I think we can remove `Constants.erc20ContractsSupportingActivities` before merging since all supported (ERC20 and ERC721) base TokenScript files are bundled into the app. We can use overriding for other tokens.


James-Sangalli (changes requested):
Two noticeable issues:

Simulator Screen Shot - iPhone 7 Plus - 2020-08-20 at 11 44 03
Simulator Screen Shot - iPhone 7 Plus - 2020-08-20 at 11 44 36

1. "invalid signature" on ETH card
2. Duplicate tokens for ALP

hboon:
@James-Sangalli 

> invalid signature" on ETH card

"aETH", not "ETH", right?  Can you try to sign and update the bundled file `ERC20-TokenScript.tsml` and see if it works? I think it's the ERC20s that don't have a TokenScript file which has this problem.

> Duplicate tokens for ALP

Can you tap on the token and then the token icon to verify if the contracts are different? The auto generated icons' (the one with the "ALP") background color are generated based on the contract [1] from a handful of colors, so different background colors means they are different contracts.

[1] https://github.com/AlphaWallet/alpha-wallet-ios/blob/42f1fd17bf46bb0a0750da2a5700a1715533c9d7/AlphaWallet/UI/TokenObject+UI.swift#L19

James-Sangalli:
> @James-Sangalli
> 
> > invalid signature" on ETH card
> 
> "aETH", not "ETH", right? Can you try to sign and update the bundled file `ERC20-TokenScript.tsml` and see if it works? I think it's the ERC20s that don't have a TokenScript file which has this problem.
> 
> > Duplicate tokens for ALP
> 
> Can you tap on the token and then the token icon to verify if the contracts are different? The auto generated icons' (the one with the "ALP") background color are generated based on the contract [1] from a handful of colors, so different background colors means they are different contracts.
> 
> [1]
> 
> https://github.com/AlphaWallet/alpha-wallet-ios/blob/42f1fd17bf46bb0a0750da2a5700a1715533c9d7/AlphaWallet/UI/TokenObject+UI.swift#L19

1. Yes, it is aETH, now the verification simply drags (might be related to a server outage)

2. They are different contracts, all on rinkeby

hboon:
> @hboon edit: invalid signature still on aETH, tested with curl and it passes.
> 
> curl -F file=@aETH.tsml https://aw.app/api/v1/verifyXMLDSig
> 
> {"result":"pass","subject":"CN=*.aw.app","keyName":"AlphaWallet","keyType":"SHA256withRSA","issuer":"CN=Let's Encrypt Authority X3,O=Let's Encrypt,C=US"}
> 
> So it has to do with this build. Note that the legacy TS files still validate signatures properly

Is this what you tried?:

> Can you try to sign and update the bundled file ERC20-TokenScript.tsml

hboon:
PS: I think I don't get a notification when we/you edit a GitHub comment and include an @mention

hboon:
@James-Sangalli pushed, with a separate commit so it's easier to see the fix

- [ ] Squash before merge

hboon:
Added support for ERC721 and base/overriding TokenScript 

@James-Sangalli Can you help to canonicalize the base ERC20 and ERC721 bundled?

I probably missed some things, I'll check again. But I think most of the features Activity is now included. A notable missing one is the native UI for the base TokenScript activities..

hboon:
@James-Sangalli good for review again. Starting up the app is slow with Activities enabled, it takes 3-4 seconds.

James-Sangalli:
@hboon 

> Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=10 "Migration is required due to the following errors:

hboon:
@James-Sangalli I have changed the schema only once within Activity quite a long time ago and have not recently. Can you see what it's referring to, just in case?

Otherwise deleting and reinstalling should fix it.

hboon:
@James-Sangalli can you give it another go? It's still slow, but much faster than previous attempts.

hboon:
But careful because this PR upgrades the Realm pod, switching to an older version of our codebase and building maybe require the app to be deleted first.

James-Sangalli (approved):
Phew, great stuff!

hboon:
Forced pushed after rebasing master

hboon:
Forced pushed after rebasing master

hboon:
@James-Sangalli anything I can sacrifice to merge this? :)

hboon:

PR #2111, opened 2020-08-19
Tap on cell containing "Add / Hide Tokens" in Wallet tab shouldn't select the 0th token #2110

### LINKED ISSUE #2110: Tap on cell containing "Add / Hide Tokens" in Wallet tab shouldn't select the 0th token
This is because the didSelect function doesn't check for the section number.

vladyslav-iosdev:
Closes #2110


James-Sangalli (approved):
@hboon FYI

hboon:
@James-Sangalli thanks!

hboon:
@vladyslav-iosdev would you split the fix and refactoring into separate PRs or commits in the future? Easier to review (and rollback if necessary).

PR #2115, opened 2020-08-25
Fix: TokenScript attribute fetching (actually most smart contract function calls) not working anymore, so TokenScript views display NaN #1652

### LINKED ISSUE #1652: TokenScript client displays values as NaN after opening and closing views a few times again
Similar to #1530 ? Tapping around, opening and closing Dai related TokenScript views for ~30 sec according to Victor's video.

hboon:
Fixes #1652

This was because too many calls are made. Since the Ethereum block time is > 10s as of now, the (simple) solution is to cache the values for a few seconds (3).

This is only reproducible on a device, not in the simulator

This took a long time to fix. Lots of debugging code written and very slow tweak-verify-build cycle because I was sent down the rabbit hole at https://github.com/AlphaWallet/web3swift/pull/1 (which is hard to test because I had to rebuild the project after each tweak in the pod) mistaking that it was the cause.

<p align=center style='font-size: 100%'>··•··</p>

To reproduce the problem, trigger 1000s of smart contract calls fairly quickly.

### DIFF
--- a/AlphaWallet/Tokens/Helpers/CallSmartContractFunction.swift
+++ b/AlphaWallet/Tokens/Helpers/CallSmartContractFunction.swift
@@ -4,14 +4,34 @@ import Foundation
 import PromiseKit
 import web3swift
 
-//TODO maybe we should cache promises that haven't resolved yet. This is useful/needed because users can switch between Wallet and Transactions tab multiple times quickly and trigger the same call to fire many times before any of them have been completed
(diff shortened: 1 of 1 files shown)


hboon:
James, please forgive me for the horrible branch name.

James-Sangalli (approved):
Well done!

zhangzhongnan928:
well done, finally solved.
Can I have a build with this fix?

PR #2122, opened 2020-09-01
Fix again: TokenScript attribute fetching (actually most smart contra…ct function calls) not working anymore, so TokenScript views display NaN

### LINKED ISSUE #1652: TokenScript client displays values as NaN after opening and closing views a few times again

hboon:
Fixes #1652

Follow up attempt to #2115 to fix #1652

The cause is still as identified previously:

> To reproduce the problem, trigger 1000s of smart contract calls fairly quickly.

It's still happening because:

A. Going to the Wallet tab triggers a refresh. This pulls the balance of every token in the Realm database
B. Going to the Transaction tab triggers a refresh. This pulls the balance of every token in the Realm database
C. Pull down to refresh in either tab has the same result as (A) and (B)

This is heavy for wallets with many tokens.

A highly probable way to verify that #1652 is fixed is to perform this sequence:

1. Tab on a token which has a TokenScript file, and tap a TokenScript action to show it's view
2. Verify if the attributes are loaded and displayed
3. Back to Wallet tab
4. Keep repeating 1-3 and see if the attributes go to N/A

The problem is step (3) makes the issue worse, because of (A)!

This PR throttles this batch refreshes. (A), (B) and (C) no longer get triggered within a 3 minute window. This isn't perfect because there'll inevitably be a wallet that holds enough tokens to trigger this with a single refresh. And within each batch, each requests is spread out over 0-20 seconds.

A better approach worth exploring is to fix and possibly improve web3swift so that batching works and we can see if that removes (or at least reduces) the need for such a limitation.

### DIFF
--- a/AlphaWallet/Tokens/Helpers/CallSmartContractFunction.swift
+++ b/AlphaWallet/Tokens/Helpers/CallSmartContractFunction.swift
@@ -10,7 +10,7 @@ var smartContractCallsCache = [String: (promise: Promise<[String: Any]>, timesta
 func callSmartContract(withServer server: RPCServer, contract: AlphaWallet.Address, functionName: String, abiString: String, parameters: [AnyObject] = [AnyObject](), timeout: TimeInterval? = nil) -> Promise<[String: Any]> {
     //We must include the ABI string in the key because the order of elements in a dictionary when serialized in the string is not ordered. Parameters (which is ordered) should ensure it's the same function
     let cacheKey = "\(contract).\(functionName) \(parameters) \(server.chainID)"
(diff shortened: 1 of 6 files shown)


hboon:
We can/should look at a better fix with web3swift later. Even if it doesn't fix this completely, it'll be a great improvement

James-Sangalli (approved):
Working on my side, cheers

zhangzhongnan928:
Let me test

PR #2154, opened 2020-09-21
Speed up app by not fetching all blocks when fetching transactions for token detection

hboon:
A. `GetContractInteractions.getContractList()` is called once per launch (called again if the user switches wallet) for ERC20 transactions and once for non-ERC20s [1]
B. `GetContractInteractions.getErc20Interactions()` is called at regular intervals

Before this PR, they both attempt to fetch data without specifying the `startBlock`.

This PR stores the last seen block from the results of each of these calls and only looks at newer blocks.

(B) in particular is really heavy and blocks the UI for *a few seconds* when running a decent-sized wallet

[1] The call for ERC20s transactions is actually a duplicate of (B), but this PR doesn't fix/refactor this.

### DIFF
--- a/AlphaWallet/Settings/Types/Config.swift
+++ b/AlphaWallet/Settings/Types/Config.swift
@@ -64,6 +64,51 @@ struct Config {
         return id
     }
 
+    private static func generateLastFetchedErc20InteractionBlockNumberKey(_ wallet: AlphaWallet.Address) -> String {
+        "\(Keys.lastFetchedAutoDetectedTransactedTokenErc20BlockNumber)-\(wallet.eip55String)"
+    }
+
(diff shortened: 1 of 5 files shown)


James-Sangalli (changes requested):
Thread 1: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=10 "Migration is required due to the following errors:
- Property 'EventActivity.transactionId' has been added." UserInfo={NSLocalizedDescription=Migration is required due to the following errors:
- Property 'EventActivity.transactionId' has been added., Error Code=10}

James-Sangalli (approved):
Build problems are from an intermediary build conflict with the DB. Not an issue for our users

Issue #2140, opened 2020-09-16
one same activity appears 3 times in a row


hboon:
same activitie appears 3 times on test-flight edition, according to Victor.

<img width=200 src="https://user-images.githubusercontent.com/56189/93288435-64a53280-f80e-11ea-8b4a-77f335012658.jpg">
<img width=200 src="https://user-images.githubusercontent.com/56189/93288437-65d65f80-f80e-11ea-8c54-3cd72f424351.jpg">

James-Sangalli:
@colourful-land good opportunity to try out the activity TokenScripts, let me know if you are up for it (have assigned you).

colourful-land:
The problem is not observed on the AppStore release 3.06 - called up Victor who said that the issue is on test-flight. Have to re-test again (I hope there is version info in the description).

colourful-land:
Can't reproduce the problem using your Ethereum address @zhangzhongnan928 

Never saw multiple-same-activity together after many scrolls. Closing this unless you can provide a screenshot with date so we can scrool to the same spot and compare.

![unnamed](https://user-images.githubusercontent.com/548435/93450794-bab2cc80-f919-11ea-8819-798e95adbb31.jpg)

colourful-land:
Victor provided this:
![record](https://user-images.githubusercontent.com/548435/93452957-63fac200-f91c-11ea-8b81-826add83f2a5.gif)

hboon:
@zhangzhongnan928 can you share the wallet address and the TokenScript file you used?

James-Sangalli:
@zhangzhongnan928 can you provide the tx hash? it is possible that multiple similar events are logged in the transaction itself

hboon:
Closing until it's reproducible again

PR #2155, opened 2020-09-25
Improve activities performance

hboon:
Especially from app launch till Activities tab is available.

Primarily by only ever, loading a subset of activities. 1000 for now, as well as transactions in the same period.


James-Sangalli (approved):
Wow, really nice. Why not make it a public build?

hboon:
@James-Sangalli there's this issue #2140. Otherwise, I don't mind.

PR #2182, opened 2020-10-05
Fix: pushing Advanced Settings screen and List of wallets screen did not hide the tab bar

hboon:
Before PR|After PR
-|-
| 
|


AlphaWallet/Settings/Coordinators/SettingsCoordinator.swift
```diff
@@ -39,6 +39,7 @@ class SettingsCoordinator: Coordinator {
     lazy var advancedSettingsViewController: AdvancedSettingsViewController = {
         let controller = AdvancedSettingsViewController()
         controller.delegate = self
+		controller.hidesBottomBarWhenPushed = true
```
James-Sangalli, line 42: opps!

James-Sangalli (approved):

AlphaWallet/Settings/Coordinators/SettingsCoordinator.swift
hboon: Ouch

PR #2209, opened 2020-10-16
Refactor transaction gas price limit estimating

### LINKED ISSUE #2101: Refactor transaction handling code

hboon:
Closes #2101 

* Estimate gas price and gas limit for all type of transactions
* Estimation of gas price and gas limit for send transactions is now only done in 1 place (not done for TokenScript transactions as well as import ERC875 transactions yet)
* Make sure that when the gas price and limit estimates are retrieved, it is only applied if the user hasn't set their own
* Only provide gas limit estimate if it hasn't already been specified in the transaction (currently only dapp-initiated transactions optionally set this)

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -152,7 +152,7 @@ class TransactionConfigurator {
     }
 
 // swiftlint:disable function_body_length
-    func load(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
+    func start(completion: @escaping (ResultResult<Void, AnyError>.t) -> Void) {
         switch transaction.transferType {
         case .dapp:
             configuration = TransactionConfiguration(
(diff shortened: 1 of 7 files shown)


James-Sangalli (changes requested):
@hboon I feared this PR. It seems there are some race conditions on the gas limit setting. It was just working before (hanging on a string) but is broken here. 

gaslimit

hboon:
@James-Sangalli can you describe what you saw? It's (trying to) treat gas limit and price the same way.

1. Request for estimates
2. While waiting, user can override if they want to
3. When the estimates are retrieved, use them if the user has not overridden it
4. Users can still override

It would be clearer and less error prone than the original version. Let's see if we can fix it?

James-Sangalli:
> @James-Sangalli can you describe what you saw? It's (trying to) treat gas limit and price the same way.
> 
> 1. Request for estimates
> 2. While waiting, user can override if they want to
> 3. When the estimates are retrieved, use them if the user has not overridden it
> 4. Users can still override
> 
> It would be clearer and less error prone than the original version. Let's see if we can fix it?

Sorry @hboon I should've clarified that I am referring to the gas limit defaulting to 100k, this is evidence that the gas limit is simply assigning the hardcoded value rather than getting the estimate

hboon:
> 2. While waiting, user can override if they want to

Yeah, it happens because we are still waiting for the estimates. The limit (and similarly price) should change after a while though. Does it?

(Maybe it was faster previously because the estimate for ERC20 token sends were done earlier, we can address this too)

James-Sangalli:
> > 1. While waiting, user can override if they want to
> 
> Yeah, it happens because we are still waiting for the estimates. The limit (and similarly price) should change after a while though. Does it?
> 
> (Maybe it was faster previously because the estimate for ERC20 token sends were done earlier, we can address this too)

It flat out stays at the hardcoded value, doesn't update (vs before)

hboon:
> It flat out stays at the hardcoded value, doesn't update (vs before)

Fixed. Was a bug. Have also `rebased master`. Would you help check again?

James-Sangalli (approved):

PR #2210, opened 2020-10-16
Retry brew update if failed.

samuelhbne:
Retry brew update command 3 times to avoid temporary build environment network issue.


James-Sangalli (approved):
Cheers

PR #2212, opened 2020-10-19
Fix: ERC721 transfer button not showing up after tapping on an ERC721 token instance

hboon:
This transfer button isn't displayed before this PR, due to a bug introduced when implementing Activity.


James-Sangalli (approved):
Oh dear! good spotting!

PR #2219, opened 2020-10-21
Refactor: remove most `Wallet?` references for current wallet to get rid of optionals

hboon:
No functional change.

If the current wallet is nil, something is very wrong and we should crash early rather than populate call sites with guard/`if let` which does nothing meaningful when the current wallet is nil.

Part of effort to reduce meaningless optionals in the code base.

### DIFF
--- a/AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
+++ b/AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
@@ -21,19 +21,16 @@ class AccountsViewController: UIViewController {
     private let config: Config
     private let keystore: Keystore
     private let balanceCoordinator: GetNativeCryptoCurrencyBalanceCoordinator
-    private let analyticsCoordinator: AnalyticsCoordinator?
-    lazy private var etherKeystore = try? EtherKeystore(analyticsCoordinator: analyticsCoordinator)
     weak var delegate: AccountsViewControllerDelegate?
     var allowsAccountDeletion: Bool = false
     var hasWallets: Bool {
         return !keystore.wallets.isEmpty
     }
 
-    init(config: Config, keystore: Keystore, balanceCoordinator: GetNativeCryptoCurrencyBalanceCoordinator, analyticsCoordinator: AnalyticsCoordinator?) {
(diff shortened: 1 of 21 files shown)


AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
```diff
@@ -21,19 +21,16 @@ class AccountsViewController: UIViewController {
...
     var hasWallets: Bool {
         return !keystore.wallets.isEmpty
     }
 
-    init(config: Config, keystore: Keystore, balanceCoordinator: GetNativeCryptoCurrencyBalanceCoordinator, analyticsCoordinator: AnalyticsCoordinator?) {
+    init(config: Config, keystore: Keystore, balanceCoordinator: GetNativeCryptoCurrencyBalanceCoordinator) {
         self.config = config
         self.keystore = keystore
         self.balanceCoordinator = balanceCoordinator
-        self.analyticsCoordinator = analyticsCoordinator
```
James-Sangalli, line 36: @hboon why is this in the PR?

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -142,15 +143,23 @@ open class EtherKeystore: Keystore {
...
-        do {
-            return try EtherKeystore(analyticsCoordinator: nil).recentlyUsedWallet
-        } catch {
-            return .none
+    var currentWallet: Wallet {
+        //Better crash now instead of populating callers with optionals
+        if let wallet = recentlyUsedWallet {
+            return wallet
+        } else if wallets.count == 1 {
+            return wallets.first!
```
James-Sangalli, line 151: Always knew you loved force unwrap ;)

James-Sangalli (approved):

AlphaWallet/Accounts/ViewModels/AccountsViewController.swift
hboon: Debated if I should include it. It was removed because it was added to provide wallet/keystore access. Not needed anymore because of this PR.

AlphaWallet/KeyManagement/EtherKeystore.swift
hboon: Haha. Jokes aside, that’s the least amount of change to make the PR work :)

PR #2236, opened 2020-10-29
Make pending ERC20/721 sends in Activity tab be displayed using the new Activity cells rather than old-style transactions

hboon:
Before PR|After PR
-|-
|


James-Sangalli:
@hboon what happens if the tx fails?

James-Sangalli:
Also, how will the user know it is still pending?

hboon:
> Also, how will the user know it is still pending?

Little grey icon shows it's pending. It's a little green tick if it's completed.

hboon:
> @hboon what happens if the tx fails?

No idea. Don't think it's handled correctly. Was it ever supported for iOS when we had the transactions tab? Any way, it's good to have. Created #2238 as/so we can handle it separately.

James-Sangalli (approved):

PR #2249, opened 2020-11-03
Toggling between ETH and USD in Send screen should show the same amounts #2243

### LINKED ISSUE #2243: Toggling between ETH and USD in Send screen should show the same amounts
1. Tap Ethereum (mainnet) in Wallet tab
2. Tap Send
3. Enter 1.01 ETH
4. Tap "ETH" to toggle to USD
5. Tap "USD" to toggle back to ETH

Expected
---
6a. 1.01

Observed
---
6b. 1.0099...

### LINKED ISSUE #2244: Toggling between ETH and USD in Send screen should not mess up the values
1. Tap Ethereum (mainnet) in Wallet tab
2. Tap Send
3. Enter 1.01 ETH
4. Tap "ETH" to toggle to USD
5. Tap "USD" to toggle back to ETH

Repeat 4 and 5 a few times, the USD will go totally wrong and if tapped again ETH = 0. It *looked* like the USD value multipled by the price, so USD is displayed as (assuming price is $370): 136900 (370x370). Maybe a race condition somewhere?

vladyslav-iosdev:
Closes #2243, Closes #2244

### DIFF
--- a/AlphaWallet/UI/Views/AmountTextField.swift
+++ b/AlphaWallet/UI/Views/AmountTextField.swift
@@ -78,32 +78,35 @@ class AmountTextField: UIControl {
     enum Currency {
         case cryptoCurrency(TokenObject)
         case usd
+    }
+
+    struct Pair {
+        var left: Currency
+        var right: Currency
+
+        mutating func swap() {
+            let currentLeft = left
+
+            left = right
+            right = currentLeft
+        }
+
+        var symbol: String {
+            switch left {
+            case .cryptoCurrency(let tokenObject):
+                return tokenObject.symbol
+            case .usd:
+                return Constants.Currency.usd
+            }
+        }
 
         var icon: Subscribable<TokenImage> {
-            switch self {
+            switch left {
             case .cryptoCurrency(let tokenObject):
                 return tokenObject.icon
             case .usd:
                 return .init((image: R.image.usaFlag()!, symbol: ""))
             }
-        }
-
-        var symbol: String {
-            switch self {
-            case .cryptoCurrency(let tokenObject):
-                return tokenObject.symbol
-            case .usd:
-                return "USD"
-            }
-        }
-    }
-
-    struct Pair {
-        let left: Currency
-        let right: Currency
-
-        func swapPair() -> Pair {
-            return Pair(left: right, right: left)
         }
     }
 
@@ -117,7 +120,7 @@ class AmountTextField: UIControl {
         textField.delegate = self
         textField.keyboardType = .decimalPad
         textField.leftViewMode = .always
-        textField.inputAccessoryView = makeToolbarWithDoneButton()
+        textField.inputAccessoryView = UIToolbar.doneToolbarButton(#selector(closeKeyboard), self)
         textField.textColor = R.color.black()!
         textField.font = DataEntry.Font.amountTextField
         textField.textAlignment = .right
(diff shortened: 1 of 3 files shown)


hboon:
Need to use the "closes" keyword before each issue number. I have edited the comment.

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -454,14 +502,62 @@ extension Double {
...
-        guard let decimalSeparator = Locale.current.decimalSeparator else { return self }
-        let period = "."
-        if decimalSeparator == period {
-            return self
+extension String {
+
+    var droppedTrailingZeros: String {
+        var string = self
+
+        while string.last == "0" || string.last == "." {
```
hboon, line 510: Would the "." check cause a problem with locales where the decimal separator is a comma?

AlphaWallet/UI/Views/AmountTextField.swift
```diff
@@ -145,50 +148,105 @@ class AmountTextField: UIControl {
         return label
     }()
 
+    private var cryptoToDollarRatePrevValue: Double?
```
hboon, line 151: Can you comment on what this new set of properties do and help, maybe with respect to the 2 issues? These few classes are still quite fragile so I'm hoping we can make minimal changes to it.

hboon (changes requested):

hboon:
Also, can you help remove the references to "qoin" in this PR?

vladyslav-iosdev:
@hboon can u explain me in what format do we need to display value in text field?

```
///Formats string value for display in text field.
    private func formatValueToDisplayValue(_ value: Double?) -> String {
        guard let amount = value else {
            return String()
        }

        switch currentPair.left {
        case .cryptoCurrency:
//            return StringFormatter().currency(with: amount, and: Constants.Currency.usd)
            return amount.toString(decimal: 2).droppedTrailingZeros
        case .usd:
            //NOTE: for values that greater then 0.01 threshold we use different formatters
            if amount > 0.01 {
//                return StringFormatter().alternateAmount(value: amount).droppedTrailingZeros
                return amount.toString(decimal: Constants.etherFormatterFractionDigits).droppedTrailingZeros
            } else {
                return amount.toString(decimal: 18).droppedTrailingZeros
            }
        }
    }
```

using this `return StringFormatter().currency(with: amount, and: Constants.Currency.usd)` for formatting can transform value to string like this `123,123.123` that later can't be parsed to string, but `return amount.toString(decimal: 2).droppedTrailingZeros` always converts value to string with format like `123123.123`

hboon:
> @hboon can u explain me in what format do we need to display value in text field?

It should be in the format "1,234.56" or "1.234,56" [1] depending on the decimal separator for the locale. We don't deal with huge amounts much, but have to be careful to parse the thousands separator correctly. I *think* it's always the opposite of the decimal separator, but let's try to let the system to parse it to be sure.

[1] 2 or 4 decimals places depending on fiat or crypto

hboon:
If it's better/simpler to fix this with a bigger refactoring of how we format numbers, maybe a simple fix for #2243 is to simply memoize the left-right pair the user has last type/edited?

hboon:
@vladyslav-iosdev remember to buzz me again if this is good for review

hboon (approved):

PR #2256, opened 2020-11-09
 Make camera button in address textfield follow textfield's background color #2253

### LINKED ISSUE #2253: Make camera button in address textfield follow textfield's background color
Careful that the same camera button image is used in the Wallet tab too

vladyslav-iosdev:
Closes #2253

### DIFF
--- a/AlphaWallet/UI/AddressTextField.swift
+++ b/AlphaWallet/UI/AddressTextField.swift
@@ -241,7 +241,7 @@ class AddressTextField: UIControl {
     }
 
     private func makeTargetAddressRightView() -> UIView {
-        let scanQRCodeButton = Button(size: .normal, style: .borderless)
+        let scanQRCodeButton = Button(size: .normal, style: .system)
         scanQRCodeButton.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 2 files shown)


AlphaWallet/UI/AddressTextField.swift
```diff
@@ -245,6 +245,7 @@ class AddressTextField: UIControl {
         scanQRCodeButton.translatesAutoresizingMaskIntoConstraints = false
         scanQRCodeButton.setImage(R.image.qr_code_icon(), for: .normal)
         scanQRCodeButton.addTarget(self, action: #selector(openReader), for: .touchUpInside)
+        scanQRCodeButton.backgroundColor = .clear
```
hboon, line 248: I thought setting `backgroundColor` on `UIButton` isn't so predictable. Is it necessary to fix this?

vladyslav-iosdev: > I thought setting `backgroundColor` on `UIButton` isn't so predictable. Is it necessary to fix this?

i suppose we can replace it with adding new case for ButtonStyle, (but behind the scene it still will bet background color value for UIButton)

hboon (approved):
Not too sure. But seems to work :)

PR #2267, opened 2020-11-11
Add Open in AlphaWallet DApp Browser in iOS Activities #1100

### LINKED ISSUE #1100: Add Open in AlphaWallet DApp Browser in iOS Activities
img_3268

If the user is browsing a DApp, he quickly selects Open in AlphaWallet DApp Browser, instead of copy paste.

vladyslav-iosdev:
Closes #1100

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -688,6 +688,15 @@
 		8769888D24C6ED04002BF62B /* TransactionInProgressCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8769888C24C6ED04002BF62B /* TransactionInProgressCoordinator.swift */; };
 		8782035D2431E66600792F12 /* FilterTokensCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8782035C2431E66600792F12 /* FilterTokensCoordinator.swift */; };
 		8782035F2431FBC300792F12 /* ShowAddHideTokensViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8782035E2431FBC300792F12 /* ShowAddHideTokensViewModel.swift */; };
+		878EE8EE255ADA49000210DE /* ShareContextHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 878EE8ED255ADA49000210DE /* ShareContextHandler.swift */; };
+		878EE8F5255ADA49000210DE /* AlphaWalletShare.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 878EE8EB255ADA49000210DE /* AlphaWalletShare.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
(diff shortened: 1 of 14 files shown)


hboon (changes requested):
- [x] Remember to fix the test suite.
- [ ] It crashes when I launch the extension from Safari in Simulator for iOS 12.4

vladyslav-iosdev:
> * [ ]  It crashes when I launch the extension from Safari in Simulator for iOS 12.4

i tried to reproduce it but for me its works ok, first time i did it on real device, second time tested on simulator 12.4 version.

AlphaWalletShare/ShareViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+            return
+        }
+
+        let valueResolver = DefaultItemProviderValueResolver()
+        extensionItem.resolveAttachments(valueResolver: valueResolver) { attachment in
+            if let attachment = attachment {
+                self.showOpenURL { result in
+                    switch result {
+                    case .success:
+                        self.storage.attachement = attachment
```
hboon, line 45: Do you think we can encode the URL to open directly into the custom URL, then we slip the temporary storage? eg. `awallet://share_content?u=example.com` (with % encoding)

AlphaWalletShare/ShareViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+
+        guard let extensionItem = context.inputItems.first as? NSExtensionItem else {
+            context.cancelRequest(withError: AnyError.canceled)
+            return
+        }
+
+        let valueResolver = DefaultItemProviderValueResolver()
+        extensionItem.resolveAttachments(valueResolver: valueResolver) { attachment in
+            if let attachment = attachment {
+                self.showOpenURL { result in
```
hboon, line 42: Can we open the app right away instead of showing an actionsheet? Looks like it'll work if we do a `DispatchQueue.main.async(after:completion:)`?

AlphaWallet/Share/ShareAction.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Created by Vladyslav Shepitko on 11.11.2020.
+//
+
+import Foundation
+
+enum ShareAction: String {
+
+    static let scheme = "awallet"
+
+    case content = "share_content"
```
hboon, line 14: Can you rename most of the references to "share" to "openIn"? Eg. `ShareAction` and "share_content"? A bit more intention revealing.

hboon (changes requested):
Can you squash the 2 commits into 1?

vladyslav-iosdev:
@hboon can u review this PR?

hboon (changes requested):
When I launch the extension from Safari, it switches to AlphaWallet, but when I go back to Safari, Safari doesn't respond anymore.

Doing this seems to help on the simulator.

```
DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
    self.open(url: url)
    context.completeRequest(returningItems: nil)
}

```

Can you see if this helps on the device too, or if there's a better way?

hboon (approved):
Thanks!

PR #2275, opened 2020-11-13
Fix: user enters a fractional amount in Send fungible screen like "0.1" and it is taken as "0.0999999.." instead

### LINKED ISSUE #2270: Amount entered into Send fungible (native/ERC 20) screen is not taken as it is
Amount entered into Send fungible (native/ERC 20) screen is not taken as it is

Maybe it's a bug introduced in #2249 ?

This also cause the problem where the user sees and knows there is a balance of "0.1" but is not allowed to enter "0.1"

(screenshot is taken with actionsheet PR, but issue is reproducible on master)

Screenshot 2020-11-12 at 2 26 38 PM

hboon:
Fixes #2270 

This reverts commit 58901a81a5a130ad39ea8d6bb3717073d2feb90f.

### DIFF
--- a/AlphaWallet/Foundation/StringFormatter.swift
+++ b/AlphaWallet/Foundation/StringFormatter.swift
@@ -14,18 +14,6 @@ final class StringFormatter {
         formatter.isLenient = true
         return formatter
     }()
-
-    private let alternateAmountFormatter: NumberFormatter = {
-        let formatter = NumberFormatter()
-        formatter.currencySymbol = ""
-        formatter.minimumFractionDigits = Constants.etherFormatterFractionDigits
-        formatter.maximumFractionDigits = Constants.etherFormatterFractionDigits
-        formatter.roundingMode = .down
-        formatter.numberStyle = .currency
-
-        return formatter
-    }()
-    
     /// Converts a Double to a `currency String`.
     ///
     /// - Parameters:
@@ -54,14 +42,5 @@ final class StringFormatter {
     /// - Returns: `String` representation.
     func formatter(for double: Double) -> String {
         return String(format: "%f", double)
-    }
-
-    func alternateAmount(value: Double) -> String {
-        //For some reasons formatter adds trailing whitespace
-        if let value = alternateAmountFormatter.string(from: NSNumber(value: value)) {
-            return value.trimmingCharacters(in: .whitespacesAndNewlines)
-        } else {
-            return "\(value)"
-        }
     }
 }
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -214,6 +214,7 @@ class SendViewController: UIViewController {
 
         amountLabel.font = viewModel.textFieldsLabelFont
         amountLabel.textColor = viewModel.textFieldsLabelTextColor
+        amountTextField.currentPair = viewModel.amountTextFieldPair
         amountTextField.isAlternativeAmountEnabled = false
         amountTextField.selectCurrencyButton.isHidden = viewModel.currencyButtonHidden
         amountTextField.selectCurrencyButton.expandIconHidden = viewModel.selectCurrencyButtonHidden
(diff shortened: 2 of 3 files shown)


vladyslav-iosdev:
@hboon can u explain me steps to reproduce this issue, maybe i did something wrong to get appropriate result?

hboon:
1. Tap Ethereum in Wallet tab
2. Tap Send
3. Enter 0.1
4. Paste address
5. Tap Send

Expected
---
6a. -0.1

Observed
---
6b. -0.100000006 (or 0.0)

hboon:
I stepped in the code earlier and noticed this might be the problem, but haven't looked into it:

```
    var ethCost: String {
        get {
            return ethCostRawValue?.toString(decimal: 18) ?? "0"
```

vladyslav-iosdev:
can u send me some eth to address 0x5d4Fdf90c45a7310e7DeAB4471038C067c2BdfCE ?

vladyslav-iosdev:
> I stepped into the code earlier and noticed this might be the problem, but haven't looked into it:
> 
> ```
>     var ethCost: String {
>         get {
>             return ethCostRawValue?.toString(decimal: 18) ?? "0"
> ```

there was not any issues with this before.
```
func toString(decimal: Int) -> String {
        let value = self == 0.0 ? 0 : max(0, decimal)
        
        return String(format: "%.\(value)f", self).droppedTrailingZeros
    }
```

hboon:
Done. For Ropsten

hboon:
> > I stepped into the code earlier and noticed this might be the problem, but haven't looked into it:
> > ```
> >     var ethCost: String {
> >         get {
> >             return ethCostRawValue?.toString(decimal: 18) ?? "0"
> > ```
> 
> there was not any issues with this before.
> 
> ```
> func toString(decimal: Int) -> String {
>         let value = self == 0.0 ? 0 : max(0, decimal)
>         
>         return String(format: "%.\(value)f", self).droppedTrailingZeros
>     }
> ```

This standalone snippet prints `0.100000000000000006` for me:

```
extension Character {
	var toString: String {
		return String(self)
	}
}


extension String {

	var droppedTrailingZeros: String {
		var string = self
		let decimalSeparator = Locale.current.decimalSeparator ?? "."

		while string.last == "0" || string.last?.toString == decimalSeparator {
			if string.last?.toString == decimalSeparator {
				string = String(string.dropLast())
				break
			}
			string = String(string.dropLast())
		}

		return string
	}
}

extension Double {

	var stringValue: String {
		String(self)
	}

	func toString(decimal: Int) -> String {
		let value = self == 0.0 ? 0 : max(0, decimal)
		
		return String(format: "%.\(value)f", self).droppedTrailingZeros
	}
}

let ethCostRawValue = 0.1
print(ethCostRawValue.toString(decimal: 18))
```

vladyslav-iosdev:
i suppose its because that entered value 
`let ethCostRawValue = 0.1` equals to next
ethCostRawValue	Double	0.10000000000000001
and function not able to convert it in right way

vladyslav-iosdev:
@hboon what if we replace `Double` type with `Decimal` for `Amount Text Field`, i suppose it could help us to avoid this precision loss

hboon:
Possibly

vladyslav-iosdev:
can u help me to resolve issue? changes i made based on #2243  but branch `revert-pr-to-fix-send-fungible-amounts-entered-by-user-not-match` doesn't containt required changes. Canu push changes to #2243?

hboon:
I’ll check the PR, but it case you still need a reply: we can throw away #2243 if there’s a separate PR that fixes it.

vladyslav-iosdev:
> I’ll check the PR, but it case you still need a reply: we can throw away #2243 if there’s a separate PR that fixes it.

Ok, i got it. Besides, I was wrong with prev my message (i meant i did those changes based on master branch changes that contains changes from #2243)

hboon:
Just noticed, when I meant throw away #2243, I mean throw away *this* PR :)

hboon:
Superseded by #2279

PR #2280, opened 2020-11-16
In Activity tab, show "All <symbol>" instead of an very long string of numbers that garbles the display for ERC20 approvals for very large amounts

hboon:
Doesn't look like there's a standard number being used for incredibly large amounts, so this PR just picks an arbitrarily large number.

Before PR|After PR
-|-
|


hboon:
Tests work locally. No idea what Travis is barfing over.

James-Sangalli (approved):

PR #2285, opened 2020-11-17
Toggling between ETH and USD in Send screen should not mess up the values #2244

### LINKED ISSUE #2243: Toggling between ETH and USD in Send screen should show the same amounts

### LINKED ISSUE #2244: Toggling between ETH and USD in Send screen should not mess up the values

vladyslav-iosdev:
Closes #2244, closes #2243


vladyslav-iosdev:
@hboon i updated commit with removing trailing zeros from text field, worked for me. Please check and let me know if this pr is ok.

hboon:
Hmm. I thought the issue has been fixed, no?

vladyslav-iosdev:
> Hmm. I thought the issue has been fixed, no?

yes it is, i made some updates for displaying large numbers, for case when we enter amount like 10000, then do convert to usd and then back , value i text field become 10,000.00, so this update remove `.00` from string.

vladyslav-iosdev:
maybe it was better to create another issue?

hboon:
> maybe it was better to create another issue?

 Yeah. Thanks :) I'll close the other 2 issues. I missed those earlier.

PR #2298, opened 2020-11-19
Display ENS name or local wallet name in Wallet tab #2266

### LINKED ISSUE #2266: Display ENS name or local wallet name in Wallet tab's nav bar title if available, otherwise fallback to "Wallet"
Like in https://zpl.io/bJmwEox. Can you see first if it can be done in a way that is not too fragile or hard to maintain?

Note that users can set a "local wallet name" which overrides the ENS name. in Settings > Change Wallet > Tap and hold a wallet.

Screenshot 2020-11-11 at 3 49 03 PM

vladyslav-iosdev:
Closes #2266


hboon (approved):
Thanks!

Have:

* Fixed test suite
* Keep the large titles for now until we add blockies
* Rebased master

PR #2300, opened 2020-11-21
Use Actionsheet for claiming paid MagicLinks

### LINKED ISSUE #1478: Some transaction cannot adjust gas price
Like MagicLink, NFT transfer and most of the TokenScript transections.

Should we use the same transection page for all the transections.

hboon:
With this PR, all transactions now use the same actionsheet UI. Closes #1478


James-Sangalli (approved):
Looks good, thanks!

PR #2313, opened 2020-11-24
add Alfacash exchange

seichris:


hboon (changes requested):
Thanks.

Would you help to forced push again, just modifying `AlphaWallet/Browser/ViewModel/Dapps.swift`, excluding the other changes?

Apple does not allow us to ship with a dapps directory on iOS, so we had to removed it a while go. But we'll still keep the list in the code base and hopefully be able to show it again, one day.

seichris:
Thanks @hboon !

Created a new, clean PR https://github.com/AlphaWallet/alpha-wallet-ios/pull/2318

PR #2314, opened 2020-11-24
Do not show "ENS" label in Actionsheet for transaction confirmation under Recipient if there is no ENS name resolved (yet) #2310

### LINKED ISSUE #2310: Do not show "ENS" label in Actionsheet for transaction confirmation under Recipient if there is no ENS name resolved (yet)
If we tap on Recipient in the actionsheet, we see (A):

After a while, it becomes (B):

Let's make it be (B) unless an ENS name has already been resolved (or provided) and become (A) if/when it's resolved.

vladyslav-iosdev:
Closes #2310

### DIFF
--- a/AlphaWallet/Transfer/ViewControllers/TransactionConfirmationViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/TransactionConfirmationViewController.swift
@@ -98,7 +98,7 @@ class TransactionConfirmationViewController: UIViewController {
 
     var canBeDismissed = true
     weak var delegate: TransactionConfirmationViewControllerDelegate?
-
+    // swiftlint:disable function_body_length
     init(viewModel: TransactionConfirmationViewModel) {
         self.viewModel = viewModel
         super.init(nibName: nil, bundle: nil)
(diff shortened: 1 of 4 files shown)


AlphaWallet/Transfer/ViewControllers/TransactionConfirmationViewController.swift
```diff
@@ -453,14 +454,14 @@ extension TransactionConfirmationViewController {
             for (sectionIndex, section) in viewModel.sections.enumerated() {
                 let header = TransactionConfirmationHeaderView(viewModel: viewModel.headerViewModel(section: sectionIndex))
                 header.delegate = self
-                var children: [UIView] = []
+//                var children: [UIView] = []
```
hboon, line 457: Can you remove the commented out code in the commit?

AlphaWallet/Transfer/ViewControllers/TransactionConfirmationViewController.swift
```diff
@@ -453,14 +454,14 @@ extension TransactionConfirmationViewController {
...
                 switch section {
                 case .gas:
                     header.setEditButton(section: sectionIndex, self, selector: #selector(editTransactionButtonTapped))
                 case .amount:
                     break
                 }
-                header.childrenStackView.addArrangedSubviews(children)
+//                header.childrenStackView.addArrangedSubviews(children)
```
hboon, line 464: Same here

hboon (changes requested):
If we add a delay to slow down the reverse resolution in `ENSReverseLookupCoordinator` and go proceed to try to send to 0xd8da6bf26964af9d7eed9e03e53415d37aa96045 (which should resolves to `vitalik.eth` in a while), this screen is still shown for a while until the ENS name appears:

vladyslav-iosdev:
> If we add a delay to slow down the reverse resolution in ENSReverseLookupCoordinator and go proceed to try to send to 0xd8da6bf26964af9d7eed9e03e53415d37aa96045 (which should resolves to vitalik.eth in a while), this screen is still shown for a while until the ENS name appears:

oh, i didn't check this case, working on it.

hboon:
@vladyslav-iosdev looks good. Is it go to review and merge?

hboon:
@vladyslav-iosdev build error. Maybe the `children` var was deleted when rebasing?

hboon (approved):
Thanks!

PR #2315, opened 2020-11-24
Make Fastlane for uploading to App Store connect work again #2306

### LINKED ISSUE #2306: Make Fastlane for uploading to App Store connect work again
Referring to https://docs.fastlane.tools/actions/upload_to_app_store/ (didn't check if that's the latest doc)

We are quite behind on `Fastlane` integration. Before we forked the code base, it was already set up, but I'm not sure if it ever worked. Let's try to get it to work for us.

PS: I noticed `submit_for_review` in `Deliverfile` is `true`. It (or it's replacement setting) should be set to `false`

vladyslav-iosdev:
Closes #2306

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -67,7 +67,7 @@
 		296105931FA2AA2100292494 /* UnsignedTransaction.swift in Sources */ = {isa = PBXBuildFile; fileRef = 296105921FA2AA2100292494 /* UnsignedTransaction.swift */; };
 		296105951FA2DEF000292494 /* TransactionDirection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 296105941FA2DEF000292494 /* TransactionDirection.swift */; };
 		296106CC1F776FD00006164B /* WalletCoordinatorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 296106CB1F776FD00006164B /* WalletCoordinatorTests.swift */; };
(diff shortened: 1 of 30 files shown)


AlphaWalletTests/Foundation/DecimalNumberFormatterTest.swift
```diff
@@ -38,10 +38,8 @@ class DecimalFormatterTest: XCTestCase {
         XCTAssertEqual(frDecimalFormatter.number(from: ",25"), NSNumber(value: 0.25))
     }
     func testNumberToStingInFr() {
-        //NOTE: For FR seems like it uses \u{202F} scalar for space instead of \u{00A0} like others do.
```
hboon, line 41: Can you add this fix in a separate PR?

fastlane/Fastfile
```diff
@@ -48,7 +48,7 @@ platform :ios do
     run_tests(
       workspace: "AlphaWallet.xcworkspace",
       scheme: "AlphaWalletTests",
-      device: "iPhone X",
+      device: "iPhone Xʀ",
```
hboon, line 51: Why does switching to a different device help?

hboon (changes requested):

fastlane/Fastfile
vladyslav-iosdev: yep, it seems like i didn't understand task.

vladyslav-iosdev:
@hboon can u explain what exactly doesn't work here, and how can i reproduce it?

hboon:
@vladyslav-iosdev sorry, was vague. Have updated the 1st comment in the issue #2306

PR #2327, opened 2020-11-26
Refactor: reduce use of `EthereumAddress` #2323

### LINKED ISSUE #2323: Refactor: reduce use of `EthereumAddress`
There are places where `EthereumAddress` have to be used. But let's try to reduce that, e.g passing them across code we own.

eg.:

https://github.com/AlphaWallet/alpha-wallet-ios/blob/0513af2c5b981d3a14e5dd23d323f80aa517347c/AlphaWallet/Tokens/Coordinators/GetENSOwnerCoordinator.swift#L29

vladyslav-iosdev:
Closes #2323


hboon:

hboon (approved):

PR #2330, opened 2020-11-26
Make Fastlane for uploading to App Store connect work again #2306

### LINKED ISSUE #2306: Make Fastlane for uploading to App Store connect work again

vladyslav-iosdev:
Closes #2306

### DIFF
--- a/fastlane/Deliverfile
+++ b/fastlane/Deliverfile
@@ -11,7 +11,7 @@ username "thetrustwallet@gmail.com" # your Apple ID user
 
 force true
 
-submit_for_review true
+submit_for_review false
 automatic_release false
 
 submission_information({


hboon:
Oh, does `fastlane deliver` work as it is?

vladyslav-iosdev:
> Oh, does `fastlane deliver` work as it is?

actually i don't know, i am not sure how to check that case

hboon:
Haha. Don't worry about it for now then. I'll look into it when there's a chance if you haven't. Will close this PR for now.

PR #2368, opened 2020-12-08
Fix: non-ascii character in variable name

hboon:
Can't discern from the diff, but some ruby code to print out the ascii values:

```ruby
before = 'сontainerView'
p before.each_byte.to_a
after = 'containerView'
p "     " +  after.each_byte.to_a.to_s
```

Output:

```
[209, 129, 111, 110, 116, 97, 105, 110, 101, 114, 86, 105, 101, 119]
"     [99, 111, 110, 116, 97, 105, 110, 101, 114, 86, 105, 101, 119]"
```


James-Sangalli (approved):
hmm

hboon:
Swift, for some strange reason, support unicode identifiers. Maybe so developers can show variable names which include fancy emojis.

PR #2377, opened 2020-12-09
Push instead of present modal the Wallet Address screen (the one with QR code) #2375

### LINKED ISSUE #2375: Push instead of present modal the Wallet Address screen (the one with QR code)
* From Fungible send's Receive button
* From Settings

vladyslav-iosdev:
Closes #2375


James-Sangalli (changes requested):
@vladyslav-iosdev happy to merge after fixing the broken test case that resulted from this PR

James-Sangalli (approved):

PR #2399, opened 2020-12-15
add bnb logo

guagualvcha:
1-liner summary:
Add BNB logo to alpha wallet.

Anything particular to test:


hboon:
@guagualvcha thanks for the PR.

I noticed from the Android PR https://github.com/AlphaWallet/alpha-wallet-android/pull/1695/files that its use the same icon for both the mainnet and testnet. Here's how it would look like on iOS, because we currently only show the blockchain's name for testnets on iOS (look for ROPSTEN and BINANCE in the 2nd and 4th row), it might not be obvious which is the Binance mainnet and testnet from the icons (see 1st screenshot). Do you think it's better if we just use the icon you have in the PR for just mainnet and let the autogenerated version remain (as in the 2nd screenshot).

Or, if you can amend the PR to include another icon for testnet, that would work too.

&nbsp;|&nbsp;
-|-
|

guagualvcha:
hi. @hboon. Thanks for your nice suggestion.  Unfortunately, we do not have a logo for testnet,  I am fine with the auto-generated  logo. And is it available that we rename the "Binance(BNB)" to "Binance Testnet(BNB)"? I think it would be better if it is feasible, but fine without it.

hboon:
@guagualvcha Ok. Have created #2405 and #2406.

The "Binance (BNB)" is picked up from the token name and symbol so #2406 will change the yellow label to "BINANCE TESTNET". instead.

Thanks again for the PR.

PR #2400, opened 2020-12-16
Refactor to fix unnecessary/wrong access to "current chain id"

hboon:
Current chain (ID) should only be used by dapp and WalletConnect as the other parts of the app support multichain.

### DIFF
--- a/AlphaWallet/Browser/Coordinators/QRCodeResolutionCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/QRCodeResolutionCoordinator.swift
@@ -85,24 +85,25 @@ private enum CheckEIP681Error: Error {
 }
 
 final class QRCodeResolutionCoordinator: Coordinator {
+    enum Usage {
+        case all(tokensDatastores: [TokensDataStore], assetDefinitionStore: AssetDefinitionStore)
+        case importWalletOnly
+    }
 
-    private let tokensDatastores: [TokensDataStore]?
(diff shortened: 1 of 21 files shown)


AlphaWallet/Browser/Coordinators/QRCodeResolutionCoordinator.swift
```diff
@@ -282,6 +281,18 @@ extension QRCodeResolutionCoordinator: ScanQRCodeCoordinatorDelegate {
...
+    private func serverFromEip681LinkOrDefault(_ serverInLink: RPCServer?) -> RPCServer? {
+        let server: RPCServer
+        if let serverInLink = serverInLink {
+            return serverInLink
+        }
+        if config.enabledServers.count == 1 {
+            //Specs https://eips.ethereum.org/EIPS/eip-681 says we should fallback to the current chainId, but since we support multiple chains at the same time, we only fallback if there is exactly 1 enabled network
+            return config.enabledServers.first!
+        }
+        return nil
```
James-Sangalli, line 294: else? :)

James-Sangalli (approved):

AlphaWallet/Browser/Coordinators/QRCodeResolutionCoordinator.swift
hboon: @James-Sangalli nah. There are more than one if(s). So they are guard/early return style. No `else` for those.

PR #2439, opened 2021-01-07
Hotfix: Sending tokens 10s only sends 1 because trailing zeros are stripped

hboon:
For posterity. Already merged into master with #2436

Supersedes #2438


hboon:
Not to be merged

hboon:
Submitted hotfix as 3.13 (369)

PR #2451, opened 2021-01-12
Speed up app launch #2374

### LINKED ISSUE #2374: Speed up app launch
We have done this before, but good to revisit again. See if we can make it faster without drastic changes

vladyslav-iosdev:
Closes  #2374


AlphaWallet/InCoordinator.swift
```diff
@@ -225,7 +225,10 @@ class InCoordinator: NSObject, Coordinator {
 
     private func oneTimeCreationOfOneDatabaseToHoldAllChains() {
         let migration = MigrationInitializer(account: wallet)
-        migration.oneTimeCreationOfOneDatabaseToHoldAllChains(assetDefinitionStore: assetDefinitionStore)
+
+        DispatchQueue.global().async {
```
hboon, line 229: We can't move the call for `oneTimeCreationOfOneDatabaseToHoldAllChains` to the background because it'll be very fragile. All other database access would fail or return wrong results until this is complete, and the database might be in an invalid state. I think the function is mostly a no-op if there's nothing to migrate and the migration only runs at most once, and only for users who started using the app a very long time ago. So maybe no need to optimize this.

AlphaWallet/TokenScriptClient/Models/EventsDataStore.swift
```diff
@@ -63,13 +55,37 @@ class EventsDataStore: EventsDataStoreProtocol {
...
+    func add(events: [EventInstanceValue], forTokenContract contract: AlphaWallet.Address) -> Promise<Void> {
+        if events.isEmpty {
+            return .value(())
+        }
+
+        return Promise { seal in
+            do {
+                let realm = self.realm.threadSafe
+                try realm.write {
+                    let eventsToSave = events.map { EventInstance(value: $0) }
```
hboon, line 81: So `EventInstance(value: $0)` uses KVC? So it might be possible for it to break when the properties no longer match. Can you see if you can write a test class that helps to prevent this error in the future? It doesn't have to be a test case that checks at runtime (e.g. `ServersCoordinatorTests.testServerListIsComplete`), it could be be something that just fails to build if you just list out the properties for both and add a comment that the programmer should make sure they are the same. Eg.

//a comment here to describe what this test case helps to guard against
let v = EventInstanceValue(contract: ...)
let _ = v.chainId
let _ = v.blockNumber
//other properties
//repeat for EventInstance with the same list of properties as above

AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -861,3 +867,9 @@ class TokensDataStore {
     }
 }
 // swiftlint:enable type_body_length
+
+extension Realm {
+    var threadSafe: Realm {
+         try! Realm(configuration: self.configuration)
```
hboon, line 873: Is this a slow operation? Or are we ok as long as we create it and batch read/write to it?

hboon (commented):
Would you help to write a short description of the approach? I suppose it's along the lines of create value objects and then creating the corresponding Realm objects as a second step, and the second step is the one we can probably move to the background after this PR?

Are events (i.e. `EventInstance` in the Activity tab) the only Realm object affected in the PR? Not TokenObjects?

I squashed the commits into 1 locally when reviewing. Can you help to clean it up? (I noticed wip commits).

AlphaWallet/InCoordinator.swift
vladyslav-iosdev: yep, ok i got it.

vladyslav-iosdev:
> Would you help to write a short description of the approach? I suppose it's along the lines of create value objects and then creating the corresponding Realm objects as a second step, and the second step is the one we can probably move to the background after this PR?
> 
> Are events (i.e. `EventInstance` in the Activity tab) the only Realm object affected in the PR? Not TokenObjects?
> 
> I squashed the commits into 1 locally when reviewing. Can you help to clean it up? (I noticed wip commits).

i added value types like EventInstance, TransactionInstance, LocalizedOperationObjectInstance.
i was need to create these structure equvivalents of realm types because when we trying to pass realm objects bettwen threads 
First step we create value type objects (after loading), pass them bettwen all promises, and so on. and right before saving them, create their realm instances.

AlphaWallet/Tokens/Types/TokensDataStore.swift
vladyslav-iosdev: seems like it ok, but i will try to play with other init method for realm 

`Realm(configuration: <#T##Realm.Configuration#>, queue: <#T##DispatchQueue?#>)
`

to avoid creating threadSafe object

AlphaWallet/TokenScriptClient/Models/EventsDataStore.swift
vladyslav-iosdev: it seems like 

```
@objc dynamic var json: String = "{}" {
        didSet {
            _data = EventActivity.convertJsonToDictionary(json)
        }
    }
```
will never get called, because it sets only in `init` method.

hboon:
O

> > Would you help to write a short description of the approach? I suppose it's along the lines of create value objects and then creating the corresponding Realm objects as a second step, and the second step is the one we can probably move to the background after this PR?
> > Are events (i.e. `EventInstance` in the Activity tab) the only Realm object affected in the PR? Not TokenObjects?
> > I squashed the commits into 1 locally when reviewing. Can you help to clean it up? (I noticed wip commits).
> 
> i added value types like EventInstance, TransactionInstance, LocalizedOperationObjectInstance.
> i was need to create these structure equvivalents of realm types because when we trying to pass realm objects bettwen threads
> First step we create value type objects (after loading), pass them bettwen all promises, and so on. and right before saving them, create their realm instances.

Great. Ok, minimum changes to this. Let's get #2469 in so we can make a build, and then this merge this.

hboon:
This good for review again?

vladyslav-iosdev:
> This good for review again?

yes, but let me resolve conflicts first.

vladyslav-iosdev:
> This good for review again?

not yet, seems like improvements work, but not for 100 percents. i got one suggestion, in some places we use `EthereumAddress(address: contract)` and seems like this is very expansive operation. What if we replace it with init with data  `EthereumAddress(contract.data)!`? (where contract is AlphaWallet.Address). i suppose it will be less safe because we not use `toChecksumAddress`

hboon:
> > This good for review again?
> 
> not yet, seems like improvements work, but not for 100 percents. i got one suggestion, in some places we use `EthereumAddress(address: contract)` and seems like this is very expansive operation. What if we replace it with init with data `EthereumAddress(contract.data)!`? (where contract is AlphaWallet.Address). i suppose it will be less safe because we not use `toChecksumAddress`

If what you have (after cleaning up) is an improvement, we can merge it first. Otherwise it gets harder to  merge and harder to revert after merging.

hboon:
> > This good for review again?
> 
> not yet, seems like improvements work, but not for 100 percents. i got one suggestion, in some places we use `EthereumAddress(address: contract)` and seems like this is very expansive operation. What if we replace it with init with data `EthereumAddress(contract.data)!`? (where contract is AlphaWallet.Address). i suppose it will be less safe because we not use `toChecksumAddress`

Yes, we should continue to use checksums. If we move the creation of the Address/EthereumAddress objects to the background, the problem should go away? Maybe a separate PR?

hboon:
@vladyslav-iosdev sorry, one of my recent PRs (for grouping transactions) messed this up horribly. Do you want to work on this again?

vladyslav-iosdev:
> @vladyslav-iosdev sorry, one of my recent PRs (for grouping transactions) messed this up horribly. Do you want to work on this again?

Yep sure, i will check it later.

hboon:

hboon (approved):
Didn't check. Trust it hasn't changed much :)

PR #2457, opened 2021-01-14
Add send all funds button #2450

### LINKED ISSUE #2450: Add send all funds button
Can probably track down an approach from https://github.com/AlphaWallet/alpha-wallet-android/pull/1704

Easier for tokens, bit more work for native crypto. Careful that xDai has a fixed gas price (look for `estimateGasPriceForXDai`)

vladyslav-iosdev:
Closes #2450

### DIFF
--- a/AlphaWallet/Transfer/Coordinators/SendCoordinator.swift
+++ b/AlphaWallet/Transfer/Coordinators/SendCoordinator.swift
@@ -127,6 +127,11 @@ extension SendCoordinator: ScanQRCodeCoordinatorDelegate {
     }
 }
 
+struct FungiblesTransactionAmount {
+    var value: String
+    var isAllFunds: Bool = false
+}
+
 extension SendCoordinator: SendViewControllerDelegate {
     func openQRCode(in controller: SendViewController) {
         guard navigationController.ensureHasDeviceAuthorization() else { return }
@@ -142,7 +147,7 @@ extension SendCoordinator: SendViewControllerDelegate {
             confirmType: .signThenSend,
             keystore: keystore,
             assetDefinitionStore: assetDefinitionStore,
-            amount: amount,
+            amount: FungiblesTransactionAmount(value: amount, isAllFunds: viewController.isAllFunds),
             ethPrice: ethPrice
         )
         let coordinator = TransactionConfirmationCoordinator(navigationController: navigationController, session: session, transaction: transaction, configuration: configuration, analyticsCoordinator: analyticsCoordinator)
--- a/AlphaWallet/Transfer/ViewModels/TransactionConfirmationViewModel.swift
+++ b/AlphaWallet/Transfer/ViewModels/TransactionConfirmationViewModel.swift
@@ -117,7 +117,7 @@ extension TransactionConfirmationViewModel {
             }
         }
 
-        private let amount: String
+        private var amount: FungiblesTransactionAmount
         private var balance: String?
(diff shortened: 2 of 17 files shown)


AlphaWallet/Transfer/ViewModels/TransactionConfirmationViewModel.swift
```diff
@@ -158,12 +158,25 @@ extension TransactionConfirmationViewModel {
                     let newAmountShort = EtherNumberFormatter.short.string(from: balance - configurator.transaction.value)
                     newBalance = R.string.localizable.transactionConfirmationSendSectionBalanceNewTitle(newAmountShort, viewModel.symbol)
                 }
+
+                if self.amount.allFunds {
+                    //NOTE: we need to handle balance updates, and refresh `amount` balance - gas
+                    let diff: BigInt? = viewModel.value - (configurator.currentConfiguration.gasPrice * configurator.currentConfiguration.gasLimit)
```
hboon, line 164: `diff` can be non-optional

hboon: Also `diff` isn't intention-revealing. (it's not a difference of values)

AlphaWallet/Transfer/Coordinators/SendCoordinator.swift
```diff
@@ -127,6 +127,11 @@ extension SendCoordinator: ScanQRCodeCoordinatorDelegate {
     }
 }
 
+struct FungiblesTransactionAmount {
+    var value: String
+    var allFunds: Bool = false
```
hboon, line 132: More idiomatic: `isAllFunds`

hboon (changes requested):
I tried this:

1. In Wallet tab
2. Tap ETH for Ropsten
3. Tap "All Funds"
4. Tap Paste (address)
5. Tap Send
6. When it hits `if self.amount.allFunds {`, `allFunds` is false

hboon:
@vladyslav-iosdev this good for review again?

hboon (changes requested):
If I do this:

1. Tap send Ropsten ETH
2. Tap All Funds
3. Paste address
4. Tap Send
5. Tap Confirm
6. It displays the error "insufficient funds for gas * price + value"

After rebasing master, I added a commit to show what might be wrong. There's a comment in it:

```
xxx balance: 894161570414202023 sending: 893972570414202023
xxx balance: 894161570414202023 sending: 893972570414202023
xxx form transaction with value: 894161570414202023
xxx balance: 894161570414202023 sending: 894037679172441023
```

hboon (changes requested):
- [ ] See this (the `Amount`) when sending ERC20:



- [ ] If the ETH balance is zero (and possibly when balance less than gas), sending ETH will crash. Not sure if it happens while sending ERC20s

vladyslav-iosdev:
> * [ ]  See this (the `Amount`) when sending ERC20:
> 
> 
> 
> * [ ]  If the ETH balance is zero (and possibly when balance less than gas), sending ETH will crash. Not sure if it happens while sending ERC20s

it happens here `func hash(transaction: UnsignedTransaction) throws -> Data {` because of force unwrap, updated code to avoid it.

hboon:
@vladyslav-iosdev remember to reassign me for review if PRs are ready :)

hboon (approved):

PR #2465, opened 2021-01-19
Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller #1748

### LINKED ISSUE #1748: Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller

vladyslav-iosdev:
Closes #1748

### DIFF
--- a/AlphaWallet/Accounts/Coordinators/AccountsCoordinator.swift
+++ b/AlphaWallet/Accounts/Coordinators/AccountsCoordinator.swift
@@ -136,9 +136,7 @@ class AccountsCoordinator: Coordinator {
 
             controller.addAction(cancelAction)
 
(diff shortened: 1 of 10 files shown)


hboon:

hboon:
@vladyslav-iosdev Would you help exclude or move to a separate commit/PR changes like:

```
- present(viewController, animated: true, completion: nil)
+ present(viewController, animated: true)
```

Easier and faster to review :)

hboon:
Those cleanups are useful too.

hboon (approved):
Great. Looks like the only functional change is that on iPad, transactions opened from the Transactions tab now open in full screen mode instead of a pop up. But that wouldn't show up too since the Transactions tab is replaced by the Activities tab.

PR #2468, opened 2021-01-19
Resolve memory leak issue #2460

### LINKED ISSUE #2460: Resolve memory leak issue
name: Memory leaks
about: while switching between wallets we can observe that some view controllers don't get delete

vladyslav-iosdev:
Fixes #2460

### DIFF
--- a/AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
+++ b/AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
@@ -165,6 +165,7 @@ class ActivitiesCoordinator: Coordinator {
         activities.sort { $0.blockNumber > $1.blockNumber }
         updateActivitiesIndexLookup()
         reloadViewController(reloadImmediately: false)
+
         for (activity, tokenObject, tokenHolder) in activitiesAndTokens {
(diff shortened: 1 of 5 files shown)


AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
```diff
@@ -267,13 +269,12 @@ class ActivitiesCoordinator: Coordinator {
     }
 
     //Important to pass in the `TokenHolder` instance and not re-create so that we don't override the subscribable values for the token with ones that are not resolved yet
-    private func refreshActivity(tokenObject: TokenObject, tokenHolder: TokenHolder, activity: Activity, isFirstUpdate: Bool = true) {
```
vladyslav-iosdev, line 270: @hboon i didn't pay attention on comment first, but do we break logic if we pass `AssetAttributeValues` fetched from TokenHolder, instead pass it directly?

this change was made to avoid memory leak caused by `TokenHolder`

hboon: Yeah. It probably would :)

hboon:
Forced push to fix `stongSef` :)

hboon (approved):

PR #2481, opened 2021-01-21
Should selectively auto disconnect all WalletConnect sessions when restarting wallet #2474

### LINKED ISSUE #2474: Should selectively auto disconnect all WalletConnect sessions when restarting wallet
Currently, with #2469, we disconnect all WalletConnect sessions when `InCoordinator.restart(for:in:)` is called.

https://github.com/AlphaWallet/alpha-wallet-ios/blob/89dfa3a4c9ef24af9b288ae70a5dc7aaa72c7391/AlphaWallet/InCoordinator.swift#L507

But this function is called for a few reasons:

A. User switches wallet — disconnect all is good here
B. User changes the enabled networks — should only disconnect those that are for now disabled networks
C. User switches language UI — don't disconnect

vladyslav-iosdev:
Closes #2474

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -146,8 +146,8 @@ class InCoordinator: NSObject, Coordinator {
         super.init()
     }
 
-    func start() { 
-        
+    func start() {
+
         showTabBar(for: wallet)
         checkDevice()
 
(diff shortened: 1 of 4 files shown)


hboon:

hboon (approved):

PR #2490, opened 2021-01-25
Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller #1748

### LINKED ISSUE #1748: Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller

vladyslav-iosdev:
Closes #1748


hboon (changes requested):
Shall we do this 1 flow in each PR at a time? :)

hboon:
@vladyslav-iosdev could you break this up so it just affect 1 screen, or at most 1 flow for each PR? Easier to check that the flow isn't broken.

PR #2491, opened 2021-01-25
Added Eporio DAPP

alexsyo:


hboon (approved):
Thanks for contributing.

Unfortunately the app store forbade us from listing dapps, so this is no longer used. This list is still maintained in this repo, but in the meantime this will not be visible in the app.

PR #2492, opened 2021-01-26
Please add TaiChi Network #2489

### LINKED ISSUE #2489: Please add TaiChi Network
https://github.com/Taichi-Network/docs/blob/master/sendPriveteTx_tutorial.md

vladyslav-iosdev:
Closes #2489

### DIFF
--- a/AlphaWallet/Browser/Factory/WKWebViewConfiguration.swift
+++ b/AlphaWallet/Browser/Factory/WKWebViewConfiguration.swift
@@ -72,7 +72,7 @@ extension WKWebViewConfiguration {
                
                const addressHex = "\(address.eip55String)"
                const rpcURL = "\(server.rpcURL.absoluteString)"
-               const chainID = "\(server.chainID)"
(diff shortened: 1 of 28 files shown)


hboon:
@vladyslav-iosdev let me think and maybe clarify more about the feature first. It seems like a better way to do this is to provide a switch/toggle to use TaiChi either in the send screen/actionsheet or in the Settings tab?

hboon:
And feel free to tweak the issue title and especially the commit message:

This:

> Add TaiChi Network 

is better than:

> Please add TaiChi Network

Much easier to read in the git commit logs :)

hboon:
Hi @vladyslav-iosdev, I'm sorry, could you open a new PR and use this approach instead? https://github.com/AlphaWallet/alpha-wallet-ios/issues/2489#issuecomment-769544479. Maybe we need to add a flag (an enum maybe?) to the `UnsignedTransaction` and friends so we can discover that it should be signed with a different RPC node?

hboon:
Remember to buzz me if this is good again :)

PR #2511, opened 2021-02-15
Push `RequestCoordinator` root view controller instead presend it modally

### LINKED ISSUE #1748: Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller

vladyslav-iosdev:
Closes #1748


hboon (changes requested):
The animation is sliding from the bottom though. Could you make it appear from right to left like it's pushed?

hboon (approved):
Oh, scratch that previous comment. Working as expected.

PR #2515, opened 2021-02-15
Push `SendViewController` instead present it modally

### LINKED ISSUE #1748: Update view controllers to be either presented modally (still full screen) or pushed onto a navigation controller

vladyslav-iosdev:
Closes #1748

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -541,28 +541,7 @@ class InCoordinator: NSObject, Coordinator {
         let tokenStorage = tokensStorages[server]
 
         switch (type, session.account.type) {
-        case (.send, .real):
-            let coordinator = PaymentCoordinator(
-                    flow: type,
-                    session: session,
-                    keystore: keystore,
-                    storage: tokenStorage,
-                    ethPrice: nativeCryptoCurrencyPrices[server],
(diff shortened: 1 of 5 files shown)


AlphaWallet/InCoordinator.swift
```diff
@@ -891,13 +870,8 @@ extension InCoordinator: PaymentCoordinatorDelegate {
...
 
     func didCancel(in coordinator: PaymentCoordinator) {
-        switch coordinator.flow {
-        case .request:
-            coordinator.navigationController.setNavigationBarHidden(true, animated: false)
-            coordinator.navigationController.popToRootViewController(animated: true)
-        case .send:
-            coordinator.navigationController.dismiss(animated: true)
-        }
+        coordinator.navigationController.setNavigationBarHidden(true, animated: false)
```
hboon, line 873: After sending a transaction and closing the informational screen, the send screen (with the amount and address text boxes are still shown). Can you make it so that it goes to the Activity tab so the user can see the pending transaction? It's done in `func didFinish(_ result: ConfirmResult, in coordinator: PaymentCoordinator) {` a few lines above here.

hboon (changes requested):
Looks great. Just 1 thing to fix.

hboon (approved):

PR #2520, opened 2021-02-17
Fix: dapp web3 provider needs to invoke Dapp's callbacks for `eth_call` when an error has occurred

### LINKED ISSUE #2516: with build3.15(374) DApps cannot read wallet balance in DApp browser
DApps can load the wallet address, but cannot read anything from the node 
Tested with Uniswap, CoFiX

P.S. It seems DApps are unable to read blockchain data through AlphaWallet

hboon:
Closes #2516.

This fixes the problem where the dapp makes an `eth_call` which fails and blocks on it and get stuck if the `eth_call` fails.

Pseudo code for dapp:

```
let tokenName = await contract.getName() //uses eth_call underneath
//do something else. Never reach here since we don't return an error, the promise above don't reject
```


hboon:
Forced pushed after rebase master. Still good for review.

vladyslav-iosdev (approved):

PR #2527, opened 2021-02-18
Disable update `chainstate` for disabled RPC servers #2526

### LINKED ISSUE #2526: Disable update chain state for disabled servers

vladyslav-iosdev:
Closes #2526

### DIFF
--- a/AlphaWallet/InCoordinator.swift
+++ b/AlphaWallet/InCoordinator.swift
@@ -213,9 +213,9 @@ class InCoordinator: NSObject, Coordinator {
         let tokensStorage = tokensStorages[server]
         let etherToken = TokensDataStore.etherToken(forServer: server)
         tokensStorage.tokensModel.subscribe { [weak self, weak tokensStorage] tokensModel in
-            guard let strongSelf = self else { return }
+            guard let strongSelf = self, let tokensStorage = tokensStorage else { return }
             guard let tokens = tokensModel, let eth = tokens.first(where: { $0 == etherToken }) else { return }
-            guard let tokensStorage = tokensStorage else { return }
+            
             if let ticker = tokensStorage.coinTicker(for: eth) {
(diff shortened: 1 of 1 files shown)


hboon (changes requested):
Ah, now I remember why this wouldn't work :) (it's ok for BC because it only works on 1 chain).

Try this, disable Ethereum mainnet and run the app. It crashes because it tries to fetch the price for mainnet. It's used for display the fiat value in Settings > Change/Add Wallet. Can't remember if it's used elsewhere.

vladyslav-iosdev:
> Ah, now I remember why this wouldn't work :) (it's ok for BC because it only works on 1 chain).
> 
> Try this, disable Ethereum mainnet and run the app. It crashes because it tries to fetch the price for mainnet. It's used for display the fiat value in Settings > Change/Add Wallet. Can't remember if it's used elsewhere.

ok, i will check it

AlphaWallet/InCoordinator.swift
```diff
@@ -304,6 +304,8 @@ class InCoordinator: NSObject, Coordinator {
     private func setupResourcesOnMultiChain() {
         oneTimeCreationOfOneDatabaseToHoldAllChains()
         setupTokenDataStores()
+        setupNativeCryptoCurrencyPrices()
```
hboon, line 307: Why are these 2 lines needed? They just init the lazy vars, right?

vladyslav-iosdev: we recreate them every time when restart with a new wallet. it needed because of missing rpc server, (when disable or enable some)

hboon: Oh, because `InCoordinator` is not re-created upon switching servers/wallets. Got it.

hboon (approved):

Issue #2498, opened 2021-02-04
Allow WalletConnect to be used on the app's Web3 browser


AW-STJ:
Observed that Android and iOS WalletConnect behaves differently when used in the app's browser.

Android connects the wallet directly when tapped on WalletConnect button. iOS wallet asks for the wallet options and shows QR Code, which cannot be used on the application. 

Example Dapp: Furucombo.app

![image](https://user-images.githubusercontent.com/61957841/106877241-a6d30000-672c-11eb-8d74-ad4582e2bf7b.png)

hboon:
Several factors here:

Due to limitations of iOS, Wallet Connect works differently from Android in this case. Instead, the supported way for the user to do for iOS is:

1. Tap/stay on the Mobile tap

<img width='200' src='https://user-images.githubusercontent.com/56189/108334070-8adf5c00-720c-11eb-8d1b-4b3cbd871109.png'>

2. Tap on the 1, 2, 3, 4, find AlphaWallet and tap on it:

<img width='200' src='https://user-images.githubusercontent.com/56189/108334087-8f0b7980-720c-11eb-8bf8-38f3c43ee120.png'>

3. AlphaWallet will prompt the user to start the WalletConnect session

This functionality provided by iOS + WalletConnect is referred to by WalletConnect as "mobile linking". There's a bug in our implementation though, have created 2 PRs to fix them.

Also, it looks dapps will have to upgrade their SDK to get the latest list of wallet apps. Eg. if we visit https://example.walletconnect.org, it provides a much more extension list, including AlphaWallet.

https://furucombo.app seem to only work with WalletConnect and not web3 directly? In that case, you can't use it with a single iOS device, i.e. need to scan QR code until they update to the latest WalletConnect SDK (and those 2 PRs get merged and deployed to the App store)

hboon:
Done. In a way (see comment above)

PR #2528, opened 2021-02-18
Fix: parse WalletConnect links picked up from mobile linking

hboon:
Related to #2498

This is the WalletConnect mobile link referred to:

When the user taps on the AlphaWallet icon which is a link to a universal link with a WalletConnect link embedded in it,

Before PR: the link is not parsed correctly (I think either an old version or the one given in the docs doesn't URL encode the WalletConnect link)
After PR: the link is now parsed correctly


hboon:
#2498 can be closed after both #2528 and #2529 are closed

vladyslav-iosdev (approved):

PR #2529, opened 2021-02-18
Fix: handle WalletConnect links when the mobile link is launched in dapp browser

hboon:
Related to #2498 and #2528

Make sure that when WalletConnect and mobile link button is used in the dapp browser, we connect to WalletConnect properly instead of opening the universal link  in the dapp browser


hboon:
#2498 can be closed after both #2528 and #2529 are closed

vladyslav-iosdev (approved):

PR #2540, opened 2021-02-19
Implement deep links for xDAI Honeyswap #2537

### LINKED ISSUE #2537: Implement deep links for xDAI Honeyswap
1. Add the 'Swap with Honeyswap' option to xDAI token's the kebab menu
2. Take users to https://app.honeyswap.org/
3. When you are going to Honeyswap, you should switch to xDai network automatically. 

xDai_alphawallet_HoneySwap

vladyslav-iosdev:
Closes #2537


hboon (changes requested):
@vladyslav-iosdev it's always swapping from xDai (so only for the xDai native crypto token, on the xDai chain), so you can probably implement it in a simpler manner like #2530. If possible, set the target (received) token to be "ETH"?

vladyslav-iosdev:
> @vladyslav-iosdev it's always swapping from xDai (so only for the xDai native crypto token, on the xDai chain), so you can probably implement it in a simpler manner like #2530. If possible, set the target (received) token to be "ETH"?

@hboon  i will double check it, but according their sources it seems like there is no `inputCurrency `

hboon (approved):

PR #2573, opened 2021-03-01
Make analytics coordinator non optional

hboon:
Includes commits from #2572


vladyslav-iosdev (approved):

hboon:

PR #2591, opened 2021-03-03
Make Ramp (by crypto with fiat) available for other tokens #2587

### LINKED ISSUE #2587: Make Ramp (by crypto with fiat) available for other tokens
Originally implemented for xDai only with #2530.

Technically, have to implement it similarly to how you implemented support for the various swap sites?

Keys mentioned in https://github.com/AlphaWallet/alpha-wallet-ios/issues/2507#issuecomment-781161721

vladyslav-iosdev:
Closes #2587


hboon (changes requested):
@vladyslav-iosdev I have force pushed after rebasing master and added a few commits. Can you help to review them and squash + merge if they look ok to you?

PR #2607, opened 2021-03-04
build3.17（381）send "All Fund" will add "," in numbers which breaks the send function #2590

### LINKED ISSUE #2590: build3.17（381）send "All Fund" will add "," in numbers which breaks the send function
IMG_2325
IMG_2324

vladyslav-iosdev:
Fixes  #2590

### DIFF
--- a/AlphaWallet/Core/Features.swift
+++ b/AlphaWallet/Core/Features.swift
@@ -4,5 +4,5 @@ import Foundation
 
 enum Features {
     static let isActivityEnabled = true
-    static let isSendAllFundsFungibleEnabled = false
+    static let isSendAllFundsFungibleEnabled = true
 }
--- a/AlphaWallet/Foundation/EtherNumberFormatter.swift
+++ b/AlphaWallet/Foundation/EtherNumberFormatter.swift
@@ -16,6 +16,14 @@ final class EtherNumberFormatter {
        return formatter
     }()
 
+    static let shortPlain: EtherNumberFormatter = {
+       let formatter = EtherNumberFormatter()
+       formatter.maximumFractionDigits = Constants.etherFormatterFractionDigits
+       formatter.groupingSeparator = ""
+        
+       return formatter
+    }()
+
     static var plain: EtherNumberFormatter = {
        let formatter = EtherNumberFormatter()
        formatter.groupingSeparator = ""
@@ -35,8 +43,12 @@ final class EtherNumberFormatter {
     /// Thousands separator.
     var groupingSeparator = ","
 
+    let locale: Locale
+
     /// Initializes a `EtherNumberFormatter` with a `Locale`.
     init(locale: Locale = .current) {
+        self.locale = locale
+
         decimalSeparator = locale.decimalSeparator ?? "."
         groupingSeparator = locale.groupingSeparator ?? ","
     }
@@ -112,7 +124,7 @@ final class EtherNumberFormatter {
         if fractionalString.isEmpty {
             return integerString
         }
-        return "\(integerString).\(fractionalString)"
+        return "\(integerString)\(decimalSeparator)\(fractionalString)".droppedTrailingZeros
     }
 
     private func integerString(from: BigInt) -> String {
--- a/AlphaWallet/Foundation/StringFormatter.swift
+++ b/AlphaWallet/Foundation/StringFormatter.swift
@@ -32,18 +32,25 @@ final class StringFormatter {
     ///   - double: double to convert.
     ///   - currencyCode: code of the currency.
     /// - Returns: Currency `String` representation.
-    func currency(with value: Double, and currencyCode: String) -> String {
+    func currency(with value: Double, and currencyCode: String = "") -> String {
         let formatter = currencyFormatter
         formatter.currencyCode = currencyCode
         //Trimming is important because the formatter output for `1.2` becomes "1.2 " (with trailing space) when region = Poland
-        return formatter.string(from: NSNumber(value: value))?.trimmed ?? "\(value)"
+        return (formatter.string(from: NSNumber(value: value))?.trimmed ?? "\(value)").droppedTrailingZeros
     }
-
-    func currency(with value: NSDecimalNumber, and currencyCode: String) -> String {
+    /// Converts a NSDecimalNumber to a `currency String`.
+    ///
+    /// - Parameters:
+    ///   - double: double to convert.
+    ///   - currencyCode: code of the currency.
+    /// - Returns: Currency `String` representation.
+    func currency(with value: NSDecimalNumber, and currencyCode: String = "", usesGroupingSeparator: Bool = true) -> String {
         let formatter = currencyFormatter
         formatter.currencyCode = currencyCode
+        formatter.usesGroupingSeparator = usesGroupingSeparator
+
         //Trimming is important because the formatter output for `1.2` becomes "1.2 " (with trailing space) when region = Poland
-        return formatter.string(from: value)?.trimmed ?? "\(value)"
+        return (formatter.string(from: value)?.trimmed ?? "\(value)").droppedTrailingZeros
     }
     /// Converts a Double to a `String`.
     ///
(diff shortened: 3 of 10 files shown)


hboon (changes requested):
@vladyslav-iosdev when I tap "All Funds" for an ERC20 token with a balance of `2000`, it still includes the comma. I have added a commit with some debug logs. This is printed out:

> xxx here3: 2,100

Could you help to check?

Also the commit message, maybe:

> Send "All Fund" will add "," in numbers which breaks the send function #2590

Less noise :)

hboon:
Also tests appears to be broken

vladyslav-iosdev:
@hboon yep, it adds comma separator, but it doesn't brake send function. user is still allowed to enter values formatted in its current locale. (with comma, or white space). Do i need to remove decimal separator from string returned from `allFunds`?

hboon:
@vladyslav-iosdev yes. Better to not include the comma. Gets messy and might be confusing for people with used to different locales.

hboon:
@vladyslav-iosdev actually, hold it for a while. Let me test this a bit :)

hboon:
@vladyslav-iosdev it doesn't seem to work when the region is set to Poland. The textfield stays at 0 even if the balance is "2000"

hboon:
Yeah, I think it's better to not include the separator (probably that's the reason)?

hboon:
I pushed a commit to disable "All Funds" to master (to do a TestFlight), so remember to rebase master to undo the flag in this PR.

hboon (approved):
Still seeing this: https://github.com/AlphaWallet/alpha-wallet-ios/pull/2607#issuecomment-792417620

hboon (changes requested):
Still seeing this https://github.com/AlphaWallet/alpha-wallet-ios/pull/2607#issuecomment-792417620

vladyslav-iosdev:
@hboon i set region to Poland, and checked with balance value equal to 10000
here what i got in console for selected crypto:
```
2021-03-09 11:00:30.366597+0200 AlphaWallet[4770:858550] xxx eth all: Balance(value: 10000000000000000000000) -> 10000
2021-03-09 11:00:30.367290+0200 AlphaWallet[4770:858550] xxx here1: 10000
2021-03-09 11:00:30.368054+0200 AlphaWallet[4770:858550] xxx here2: Optional(10000)
2021-03-09 11:00:30.368316+0200 AlphaWallet[4770:858550] xxx here4: 10000
2021-03-09 11:00:30.388202+0200 AlphaWallet[4770:858550] xxx eth all: Balance(value: 10000000000000000000000) -> 10000
```

and for selected fiat currency:
```
2021-03-09 11:01:30.792816+0200 AlphaWallet[4770:858550] xxx eth all: Balance(value: 10000000000000000000000) -> 10000
2021-03-09 11:01:30.793524+0200 AlphaWallet[4770:858550] xxx here1: 10000
2021-03-09 11:01:30.794362+0200 AlphaWallet[4770:858550] xxx here2: Optional(10000)
2021-03-09 11:01:30.813892+0200 AlphaWallet[4770:858550] xxx eth all: Balance(value: 10000000000000000000000) -> 10000
```

hboon:
@vladyslav-iosdev does the text field show 10000 too?

vladyslav-iosdev:
> @vladyslav-iosdev does the text field show 10000 too?

yep, here how it looks

https://user-images.githubusercontent.com/55975226/110451464-e5156380-80cc-11eb-96ef-2cf83fb79801.MP4

hboon:
@vladyslav-iosdev Ah. I see it depends.

Can you try to watch the wallet `0x007bEe82BDd9e866b2bd114780a47f2261C684E3` and modify the code base so you can show the send fungible screen with a watched account?

Then tap on the xAW token. It's balance is `2000.00002...`. When I tap on `All Funds`, the textfield remains 0. I see this was logged:

```
2021-03-10 01:01:36.166734+0800 AlphaWallet[55331:43064023] xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
2021-03-10 01:01:36.167025+0800 AlphaWallet[55331:43064023] xxx here1: 2000.00002022471910112
2021-03-10 01:01:36.167227+0800 AlphaWallet[55331:43064023] xxx here2: nil
2021-03-10 01:01:36.167344+0800 AlphaWallet[55331:43064023] xxx here3: 
```

The problem seems to be `xxx here2: nil`, maybe strip out the thousands separator (but watch out for the decimal separator) before using the formatter might help?

hboon (changes requested):
There seems to be a problem with sending ERC20s. Not the entire balance is sent. If you go through with sending `xAW` token for the wallet `0x007bEe82BDd9e866b2bd114780a47f2261C684E3` and tap the Send button (but stop before hitting Confirm), it logs:

> xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
> xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
> xxx here1: 2000.00002022471910112
> xxx here2: Optional(2000.000020224719)
> xxx here3: 2000
> xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
> xxx erc20 value: 2000000000000000000000

Notice the last line says `2000000000000000000000`. It should be `2000000020224719101120`. But seems to work correctly for native crypto.

Also the Amount isn't displayed correctlyin the actionsheet for transaction confirmation (only for ERC20, native crypto works correctly here too):

vladyslav-iosdev:
@hboon yep, i find these issue, wanted to ask u for advice, what if we will use `EtherNumberFormatter.short` instead of `EtherNumberFormatter.plain`, because for plain formatter we don't drop fraction digits. I suppose that 4 fraction digits will be anough for displaying? what do u think?

hboon:
> xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
> xxx erc20 value: 2000000000000000000000

@vladyslav-iosdev but it doesn't seem to be because of fractional digits in the 2 lines above? The 1st preserves it and the second is BigInt/BigUInt. Unless you are saying in between the 2 lines there is some conversion to and fro that causes this?

vladyslav-iosdev:
> xxx erc20 all: 2000000020224719101120 -> 2000.00002022471910112
> xxx erc20 value: 2000000000000000000000

the reason why this happens is that we using `optionalDecimalValue: NSDecimalNumber?` as `EtherNumberFormatter.plain.decimal(from: self)` can't return formatted value, because  decimal separator is `,` and we gon value with `,` decimal separtor. To fix this we can update function 
`func string(from number: BigInt, decimals: Int) -> String`  from `EtherNumberFormatter` with this 

`return "\(integerString)\(decimalSeparator)\(fractionalString)".droppedTrailingZeros`
instead of
`return "\(integerString).\(fractionalString)".droppedTrailingZeros`

and it will return us value formated in users locale.

vladyslav-iosdev:
With these updates i got next values:

```
2021-03-11 10:59:27.636998+0200 AlphaWallet[16546:7320054] xxx here1: 
2021-03-11 10:59:27.637140+0200 AlphaWallet[16546:7320054] xxx here2: nil
2021-03-11 10:59:27.637217+0200 AlphaWallet[16546:7320054] xxx here4: 

2021-03-11 10:59:27.638511+0200 AlphaWallet[16546:7320054] xxx erc20 all: 2000000020224719101120 -> 2000
2021-03-11 10:59:29.944760+0200 AlphaWallet[16546:7320054] xxx erc20 all: 2000000020224719101120 -> 2000
2021-03-11 10:59:29.944960+0200 AlphaWallet[16546:7320054] xxx here1: 2000
2021-03-11 10:59:29.945203+0200 AlphaWallet[16546:7320054] xxx here2: Optional(2000)
2021-03-11 10:59:29.945288+0200 AlphaWallet[16546:7320054] xxx here4: 2000
2021-03-11 10:59:29.951438+0200 AlphaWallet[16546:7320054] xxx erc20 all: 2000000020224719101120 -> 2000

```

and next displaying on screen
Screenshot 2021-03-11 at 11 00 31
Screenshot 2021-03-11 at 11 00 46

hboon:
Ok. That seems to look right :) thanks

hboon:
@vladyslav-iosdev is this good to merge after review or WIP?

hboon:
Oh, sorry, I misunderstood your earlier message.

Can this:

> xxx eth all: 2000000020224719101120 -> 2000 -> 2000
> xxx eth all send: 2000
> xxx erc20 value: 2000000000000000000000

become:

> xxx eth all: 2000000020224719101120 -> 2000 -> 2000
> xxx eth all send: 2000
> xxx erc20 value: 2000000020224719101120 

(so that the last line is changed).

Similarly for native crypto.

Or is there a major change involved?

vladyslav-iosdev:
> Or is there a major change involved?

@hboon Let me check it

hboon (changes requested):
Almost there!

When the region is set to Poland and we transfer fractional native crypto (eg.` "0,099"`), the amount is displayed as "0.0" ETH even though the transaction is formed with the correct amount.

I think it's just this line: https://github.com/AlphaWallet/alpha-wallet-ios/pull/2607/files#diff-b585d1f2b6627a06352687e7fb5314bf45733759cb9fe6511a034140b173383fR192

as `Double(amount.value)` doesn't support locales.

hboon:
Force pushed after rebase master.

hboon (approved):

PR #2614, opened 2021-03-05
WalletConnect using mobile linking crashes if the chain is disabled #2609

### LINKED ISSUE #2609: WalletConnect using mobile linking crashes if the chain is disabled
1. Disable Ethereum mainnet
2. Dapp Browser tab (Maybe mobile Safari will work too)
3. Visit https://example.walletconnect.org
4. Tap "Connect to WalletConnect"
5. Tap "3"
6. Tap "AlphaWallet"

Expected:

7. It works

Observed:

7. Crashes

I think the if we change step 2 to scan a QR code, the connection attempt gets silently ignored.

If we can figure out the chain, can we make this display an error message? Like "\(server.name) is disabled. Please enable it in Settings > Select Active Networks". As a follow up PR, we can make it better like improve the message and have a button to open that screen.

vladyslav-iosdev:
Closes #2609

### DIFF
--- a/AlphaWallet/Settings/Types/Constants.swift
+++ b/AlphaWallet/Settings/Types/Constants.swift
@@ -203,6 +203,8 @@ public struct Constants {
         (name: "HKS", contract: AlphaWallet.Address(string: "0x75cB2Cc380d1387a79eE64b1b7C9Fa051139a319")!),
     ]
 
+    static let ensContractOnMainnet = AlphaWallet.Address.ethereumAddress(eip55String: "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85")
+
     static let defaultEnabledServers: [RPCServer] = [.main, .xDai]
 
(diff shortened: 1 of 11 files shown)


AlphaWallet/Settings/Types/Constants.swift
```diff
@@ -58,6 +58,7 @@ public struct Constants {
     public static let burnAddressString = "0x000000000000000000000000000000000000dEaD"
     static let nullAddress = AlphaWallet.Address(uncheckedAgainstNullAddress: "0x0000000000000000000000000000000000000000")!
     static let nativeCryptoAddressInDatabase = nullAddress
+    static let ensContractOnMainnet = nullAddress
```
vladyslav-iosdev, line 61: @hboon can u please check this one, it was mised in prev PR.

hboon: Got it in. Thanks

hboon (approved):

PR #2615, opened 2021-03-05
Improve WalletConnect #2496

### LINKED ISSUE #2496: Improve WalletConnect
1. We need to add an item in the Settings. Let me explain why: Many times users are coming from different wallets where this is a place where they can find WalletConnect (in the Settings). And without knowing about the smart camera, it looks like we don't provide the WalletConnect feature. 

So we will have two methods to use WalletConnect. Launch a camera as you already made it, OR go to Settings and open the WalletConnect item.

Screenshot 2021-01-28 at 16 15 29

Zeplin:
https://zpl.io/a3AlQ41

----

2. The list of all WalletConnect active connections should go with icons. I created a library with HQ 100 icons for the most popular dapps. 

Screenshot 2021-01-28 at 16 21 06

Zeplin:
https://zpl.io/V0X0z9x (icons)
https://zpl.io/bL6DJOd

----

3. When you open a certain connection, you should have a icon and the name at the top instead of WalletConnect logo. 
Screenshot 2021-01-28 at 14 53 10

Zeplin:
https://zpl.io/a3AlQrW

----

4. Can we unify action sheets?

Screenshot 2021-01-28 at 16 18 07

Zeplin:
https://zpl.io/2vEWDjr

vladyslav-iosdev:
Closes #2496


hboon:
Can you summarise what this PR is about?

vladyslav-iosdev:
- adds wallet connect link to settings
- allows to scan qr code from wallet connect sesstions
- updates server selection while connecting to wallet connect session (push servers view controller, instead present default action sheet)
- adds bridge to promise for sessions coordinator
- adds loading wallet connect session icon
- fixes memory leaks

hboon:

hboon (approved):
@vladyslav-iosdev I noticed a little animation glitch when pushing and popping the screen the list of WalletConnect sessions in the Settings tab. I added a commit that seems to fix it. Can you verify if it helps and then squash + merge if it's good?

Also would you modify the commit message to include these after the original line: https://github.com/AlphaWallet/alpha-wallet-ios/pull/2615#issuecomment-791372757

PR #2623, opened 2021-03-09
Switching `eth_sign` to sign personal as web3.js v1 (unlike v0.20.x) and probably other modern libraries expect so

### LINKED ISSUE #2617: Signing doesn't work
IMG_2381

IMG_2376

https://v.cent.co/tweet/298270194231549952

hboon:
Fixes #2617 

https://web3js.readthedocs.io/en/v1.2.7/web3-eth-accounts.html#sign

This breaks compatibility with https://xcontract.herokuapp.com/sign. Will have to observe if there'll be reports of this breaking other dapps.

### DIFF
--- a/AlphaWallet/Transfer/Coordinators/SignMessageCoordinator.swift
+++ b/AlphaWallet/Transfer/Coordinators/SignMessageCoordinator.swift
@@ -3,7 +3,7 @@
 import Foundation
 import UIKit
 import CryptoSwift
-import Result 
+import Result
 
 enum SignMessageType {
     case message(Data)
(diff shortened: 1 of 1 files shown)


vladyslav-iosdev (approved):

zhangzhongnan928:
Screen Shot 2021-03-18 at 12 12 06 am


Joe (AZTEC)Today at 00:04
Hmmm, it sounds like that above issue is not fixed in the wallet-connect code path
https://github.com/AlphaWallet/alpha-wallet-ios/blob/a9d5997bdd696deda5e6ded9d68592bf79564748/AlphaWallet/WalletConnect/Types/WalletConnectServerRequest.swift#L54

PR #2646, opened 2021-03-16
Closes wallet connect issues

### LINKED ISSUE #2642: If we reject a WalletConnect connection, we are stuck in the list of sessions screen
Not sure if this only happens if there are no existing connections

### LINKED ISSUE #2643: When switching servers for a new WalletConnect session, the default/chosen server should be checked

### LINKED ISSUE #2644: Can't detect when the server selection screen is closed (via back button) in the WalletConnect open session screen.
Was trying to log an event by replacing the `.cauterize()` to `.catch { _ in NSLog("xxx this doesn't get logged") }` at:

https://github.com/AlphaWallet/alpha-wallet-ios/blob/1fe24dfcc5870d991267f5af565144174ac8aad7/AlphaWallet/WalletConnect/Coordinator/WalletConnectToSessionCoordinator.swift#L84

Can you help to see what's wrong? and replace this `cauterize()` with a `catch { _ in //no-op }`?

vladyslav-iosdev:
Closes #2644, Closes #2643, Fixes #2642

### DIFF
--- a/AlphaWallet/Settings/Coordinators/ServersCoordinator.swift
+++ b/AlphaWallet/Settings/Coordinators/ServersCoordinator.swift
@@ -97,7 +97,7 @@ extension ServersCoordinator: ServersViewControllerDelegate {
 private class ServersCoordinatorBridgeToPromise {
 
     private let navigationController: UINavigationController
-    private let (promiseToReturn, seal) = Promise<RPCServer>.pending()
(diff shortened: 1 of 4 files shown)


hboon (approved):

hboon:
Took the liberty to add a summary line to the commit message (so it works better with `git log` and `git log -S`).

PR #2653, opened 2021-03-18
Show error message if dapp selects a chain for WalletConnect that is disabled in the app #2652

### LINKED ISSUE #2652: Show error message if dapp selects a chain for WalletConnect that is disabled in the app
1. Disable mainnet
2. Visit https://example.walletconnect.org
3. Tap "Connect to WalletConnect"

Expected:

4. Show error: "<Ethereum> is disabled. You can enable it in Settings > Select Active Networks" (replace with `RPCServer.name`)

Observed (screenshot):

Screenshot 2021-03-18 at 12 29 14 PM

vladyslav-iosdev:
Closes #2652

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -118,7 +118,7 @@ class AppCoordinator: NSObject, Coordinator {
     }
 
     /// Return true if handled
-    func handleOpen(url: URL) -> Bool {
+    @discardableResult func handleOpen(url: URL) -> Bool {
         let handled = urlSchemeCoordinator.handleOpen(url: url)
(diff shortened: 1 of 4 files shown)


hboon (changes requested):
Still seems to work like before the PR.

vladyslav-iosdev:
> Still seems to work like before the PR.

it appears when we tap `connect` button on appeared connection alert. Maybe i misunderstood something?

hboon:
Oh. Will check again. Thanks

hboon (approved):
@vladyslav-iosdev I was thinking to show the error and not show the actionsheet at all, but this works too. Thanks!

Issue #2655, opened 2021-03-19
Decimal and thousands separator not using the one specified in Settings > Advanced > Change Language


hboon:
Refer: #2596

Can be observed in the Wallet tab with Spanish chosen in the in-app settings. (it still uses . as the decimal separator if there's what your iOS setting is)

It works in iPhone X 12.4 Simulator, it doesn't work on iPhone 11 Pro 14.5.

Maybe it's the iOS version number?

vladyslav-iosdev:
@hboon as i found using Config.locale that was added in prev. commits should be used for tests only. Because it causes locale issues, like in https://github.com/AlphaWallet/alpha-wallet-ios/issues/2662.  I suppose we need to push off Locale.current and not from Config.locale

PR #2666, opened 2021-03-22
Wrong amount in Send actionsheet #2661

### LINKED ISSUE #2599: Check if CoinGecko API is good for our purpose
We already use CoinGecko for ETH prices, it looks like we might be able to use them for other tokens as well as historical prices:

https://www.coingecko.com/api/documentations/v3#/coins/get_coins__id__market_chart

I think there's enough data for #2274

### LINKED ISSUE #2661: Wrong amount in Send actionsheet
I tried to send 5 DAI, but it displays my total balance under "Amount".

### LINKED ISSUE #2662: No way to enter decimals when sending funds
Maybe only for Europeans?

vladyslav-iosdev:
Closes #2662, Closes  #2661

### DIFF
--- a/AlphaWallet/Settings/Types/Config.swift
+++ b/AlphaWallet/Settings/Types/Config.swift
@@ -36,7 +36,7 @@ struct Config {
     }
 
     static var locale: Locale {
-        if let identifier = getLocale() {
+        if let identifier = getLocale(), isRunningTests() {
             return Locale(identifier: identifier)
         } else {
             return Locale.current
--- a/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
+++ b/AlphaWallet/Transfer/ViewControllers/SendViewController.swift
@@ -331,7 +331,12 @@ class SendViewController: UIViewController {
                 contract: transactionType.contractForFungibleSend,
                 data: nil
         )
-        delegate?.didPressConfirm(transaction: transaction, in: self, amount: amountTextField.ethCost, shortValue: allFundsFormattedValues?.allFundsShortValue)
+        
+        delegate?.didPressConfirm(transaction: transaction, in: self, amount: amountTextField.ethCost, shortValue: shortValueForAllFunds)
+    }
+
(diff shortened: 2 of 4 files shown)


hboon:
So this PR would fix #2655 too?

AlphaWallet/Settings/Types/Config.swift
```diff
@@ -36,7 +36,7 @@ struct Config {
     }
 
     static var locale: Locale {
```
hboon, line 38: I know you mentioned this in #https://github.com/AlphaWallet/alpha-wallet-ios/issues/2655#issuecomment-803947279, but I don't understand.

When do we use `Config.locale` and when do we use `Config.getLocale()`? In other words, why do we want `Config.locale` to return `Locale.current` when the user has overridden the language in in-app Settings > Advanced > Change Language?

hboon (changes requested):
I see this when iOS region is Singapore+English and in-app setting is Polish, after app is restarted. Notice it shows "0,1" in the amount textfield, but "0.1" in the actionsheet.

Screenshot 2021-03-23 at 1 00 22 PM

hboon:
@vladyslav-iosdev also see if https://github.com/AlphaWallet/alpha-wallet-ios/issues/2662#issuecomment-804614729 and the comments after that might be relevant?

hboon:
@all-contributors please add @vladyslav-iosdev for code

vladyslav-iosdev:
> #2655

yep, it should

hboon:
@vladyslav-iosdev it works on iOS 12.4 simulator, but I tried on an iPhone 11 Pro (device) with iOS 14.5 and changing the locale in-app to and after restarting the app, the keyboard and textfields still show "." as the decimal separator. Does it work for you?

hboon:
@vladyslav-iosdev going to merge this in and I'll create a separate PR to hide (and undo) the in-app language settings for now until we sort it out.

hboon (approved):

PR #2677, opened 2021-03-26
Move TaiChi option to advanced settings section #2658

### LINKED ISSUE #2658: Move TaiChi option in Settings somewhere else.
Maybe in Settings > Advanced?

At least until we figure out if we should move it into somewhere like in the actionsheet, etc

vladyslav-iosdev:
Closes  #2658


hboon (changes requested):
Remember to remove the commented out code :)

hboon (approved):

PR #2683, opened 2021-03-30
Disable test case for TokenScript signature verification (because signature verification web API isn't working properly)

hboon:


hboon:
@vladyslav-iosdev bump.

PR #2687, opened 2021-03-30
Fetch fiat value for native crypto and ERC20 tokens #2599

### LINKED ISSUE #2274: Add Wallet Summary [Samoa]
Screenshot 2021-03-08 at 11 02 38

-----
This screen will be a replacement for the current "Change/Add Wallet". 

- It will be the first screen you see when launching AlphaWallet (already imported wallets). 
- There are two ways to enter this screen
a) from the wallet tab, you tap on a blockie 
b) You Open Settings -> Change/Add Wallet. So the same as we have now.

-----
Zeplin:
https://zpl.io/aBQO6xK

-----
If you open the summary of all wallets, it will work the same as Watch-only but for all wallets you imported (watch-only and that on-chain)
Screenshot 2020-11-12 at 11 53 28

### LINKED ISSUE #2599: Check if CoinGecko API is good for our purpose

vladyslav-iosdev:
Closes #2599

### DIFF
--- a/AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
+++ b/AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
@@ -1,0 +1,326 @@
+//
+//  CoinTickersFetcher.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 27.01.2021.
+//
+
+import PromiseKit
+import Moya
+import SwiftyJSON
+
+struct TokenMappedToTicker: Hashable {
+    let symbol: String
+    let contractAddress: AlphaWallet.Address
+    let server: RPCServer
+
+    init(tokenObject: TokenObject) {
+        symbol = tokenObject.symbol
+        contractAddress = tokenObject.contractAddress
+        server = tokenObject.server
+    }
+
+    init(symbol: String, contractAddress: AlphaWallet.Address, server: RPCServer) {
+        self.symbol = symbol
+        self.contractAddress = contractAddress
+        self.server = server
+    }
+}
+
+protocol CoinTickersFetcherType {
+    func fetchPrices(forTokens tokens: ServerDictionary<[TokenMappedToTicker]>) -> Promise<[AddressAndRPCServer: CoinTicker]>
+    func fetchChartHistories(addressToRPCServerKey: AddressAndRPCServer) -> Promise<[ChartHistory]>
+}
+
+fileprivate struct MappedCoinTickerId: Hashable {
+    let tickerId: String
+    let contractAddress: AlphaWallet.Address
+    let server: RPCServer
+}
+
+class CoinTickersFetcher: CoinTickersFetcherType {
+    private enum AnyError: Error {
+        case alreadyFetchingPrices
+    }
+
+    private static var fetchSupportedTokensPromise: Promise<[Ticker]>?
+    private static var coinGeckoTickers: [Ticker] = []
+
+    private let pricesCacheLifetime: TimeInterval = 60 * 60
+    private let dayChartHistoryCacheLifetime: TimeInterval = 60 * 60
+    private var isFetchingPrices = false
+    private var lastFetchedTickerIds: [String]?
+    private var lastFetchedDate: Date?
+
+    private static let queue: DispatchQueue = DispatchQueue(label: "com.CoinTickersFetcher.updateQueue")
+
+    private let provider: MoyaProvider<AlphaWalletService>
+    private let config: Config
+    private var tickers: [AddressAndRPCServer: CoinTicker] = [:]
+    private var historyCache: [CoinTicker: [ChartHistoryPeriod: (history: ChartHistory, fetchDate: Date)]] = [:]
+
+    init(provider: MoyaProvider<AlphaWalletService>, config: Config) {
+        self.provider = provider
+        self.config = config
+    }
+
+    //Important in implementation to not cache the returned promise (which is used to further fetch prices). We only want to cache the promise/request for fetching supported tickers
+    private static func fetchSupportedTickers(config: Config, provider: MoyaProvider<AlphaWalletService>, shouldRetry: Bool = true) -> Promise<[Ticker]> {
+        if let promise = fetchSupportedTokensPromise { return promise }
+
+        let promise: Promise<[Ticker]> = firstly {
+            provider.request(.tokensThatHasPrices(config: config))
+        }.map(on: CoinTickersFetcher.queue, { response -> [Ticker] in
+            return try response.map([Ticker].self, using: JSONDecoder())
+        }).get(on: CoinTickersFetcher.queue, { tickers in
+            CoinTickersFetcher.coinGeckoTickers = tickers
+        }).recover { _ -> Promise<[Ticker]> in
+            if shouldRetry {
+                return fetchSupportedTickers(config: config, provider: provider, shouldRetry: false)
+            } else {
+                return .value([])
+            }
+        }
+        fetchSupportedTokensPromise = promise
+        return promise
+    }
+
+    private func fetchSupportedTickers() -> Promise<[Ticker]> {
+        Self.fetchSupportedTickers(config: config, provider: provider)
+    }
+
+    func fetchPrices(forTokens tokens: ServerDictionary<[TokenMappedToTicker]>) -> Promise<[AddressAndRPCServer: CoinTicker]> {
+        firstly {
+            fetchTickers(forTokens: tokens)
+        }.get { [weak self] tickers, tickerIds in
+            self?.tickers = tickers
+            self?.lastFetchedTickerIds = tickerIds
+            self?.lastFetchedDate = Date()
+        }.map {
+            $0.tickers
+        }
+    }
+
+    private static let fetchChartHistoryTimeLimit: TimeInterval = 5
+
+    func fetchChartHistories(addressToRPCServerKey: AddressAndRPCServer) -> Promise<[ChartHistory]> {
+        let promises: [Promise<ChartHistory>] = ChartHistoryPeriod.allCases.map {
+            fetchChartHistory(force: false, period: $0, for: addressToRPCServerKey)
+        }
+        return when(fulfilled: promises)
+    }
+
+    private func fetchChartHistory(period: ChartHistoryPeriod, ticker: CoinTicker) -> Promise<ChartHistory> {
+        firstly {
+            provider.request(.priceHistoryOfToken(config: config, id: ticker.id, currency: Constants.Currency.usd, days: period.rawValue))
+        }.map(on: CoinTickersFetcher.queue, { response -> ChartHistory in
+            try response.map(ChartHistory.self, using: JSONDecoder())
+        }).recover(on: CoinTickersFetcher.queue, { _ -> Promise<ChartHistory> in
+            .value(.empty)
+        })
+    }
+
+    private func cacheChartHistory(result: ChartHistory, period: ChartHistoryPeriod, for ticker: CoinTicker) {
+        guard !result.prices.isEmpty else { return }
+        var newHistory = historyCache[ticker] ?? [:]
+        newHistory[period] = (history: result, fetchDate: Date())
+        historyCache[ticker] = newHistory
+    }
+
+    func fetchChartHistory(force: Bool, period: ChartHistoryPeriod, for key: AddressAndRPCServer, shouldRetry: Bool = true) -> Promise<ChartHistory> {
+        firstly {
+            getCachedChartHistory(period: period, for: key)
+        }.then { values -> Promise<ChartHistory> in
+            let ticker = values.ticker
+            if let value = values.history, !force {
+                return .value(value)
+            } else {
+                return firstly {
+                    self.fetchChartHistory(period: period, ticker: ticker)
+                }.get(on: CoinTickersFetcher.queue, {
+                    self.cacheChartHistory(result: $0, period: period, for: ticker)
+                })
+            }
+        }.recover { _ -> Promise<ChartHistory> in
+            if shouldRetry {
+                return self.fetchChartHistory(force: force, period: period, for: key, shouldRetry: false)
+            } else {
+                struct FetchChartHistoryError: Error {}
+                throw FetchChartHistoryError()
+            }
+        }
+    }
+
+    private func getCachedChartHistory(period: ChartHistoryPeriod, for key: AddressAndRPCServer) -> Promise<(ticker: CoinTicker, history: ChartHistory?)> {
+        struct TickerNotFound: Error {
+        }
+        if let ticker = tickers[key] {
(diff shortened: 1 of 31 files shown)


vladyslav-iosdev:
@hboon do i need to update token view controller with fetched tickers, as well as send token screen?

hboon:
>@hboon do i need to update token view controller with fetched tickers, as well as send token screen? ...

@vladyslav-iosdev update them to?

hboon:
In case you want me to check it out Vlad, I can't build it because web3swift points to a local branch :)

hboon:
> @hboon do i need to update token view controller with fetched tickers, as well as send token screen?

But in case this is what I'm thinking of: try to keep the changes minimal so we can roll it back or add on later.

vladyslav-iosdev:
> In case you want me to check it out Vlad, I can't build it because web3swift points to a local branch :)

oh, sorry i totally missed for pointing to local branch

vladyslav-iosdev:
> In case you want me to check it out Vlad, I can't build it because web3swift points to a local branch :)

@hboon pointing to local branch has removed

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+}
+
+protocol CoinTickersResolverType {
+    var tickers: [AddressToPRCServerKey: CoinTicker] { get }
+
+    func coinTickers(enabledObject: [TokenMappedToTicker]) -> Promise<Void>
+    func chartHistory(force: Bool, period: ChartHistoryPeriod, for key: AddressToPRCServerKey) -> Promise<ChartHistory>
+}
+
+class CoinTickersResolver: CoinTickersResolverType {
```
hboon, line 37: Maybe `CoinTickersFetcher` and `CoinTickersFetcherType` would be better?

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+
+        init(id: String, contractAddress: AlphaWallet.Address, server: RPCServer) {
+            self.coinTickerId = id
+            self.contractAddress = contractAddress
+            self.server = server
+        }
+    }
+
+    func coinTickers(enabledObject: [TokenMappedToTicker]) -> Promise<Void> {
+        Promise<Void> { seal in
```
hboon, line 101: I think we can just turn this into a `.map {}` instead of wrapping with another `Promise`.

AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -642,42 +646,12 @@ class TokensDataStore {
...
-
-    private func getPriceToUpdate() -> AlphaWalletService? {
-        switch server {
-        case .main:
-            return .priceOfEth(config: config)
-        case .xDai:
-            return .priceOfDai(config: config)
-        case .kovan, .ropsten, .rinkeby, .poa, .sokol, .classic, .callisto, .goerli, .artis_sigma1, .artis_tau1, .binance_smart_chain, .binance_smart_chain_testnet, .custom, .heco, .heco_testnet, .fantom, .fantom_testnet, .avalanche, .avalanche_testnet, .polygon, .mumbai_testnet:
-            return nil
+        coinTickersResolver.coinTickers(enabledObject: objects).done { _ in
```
hboon, line 651: `Promise<Void>` doesn't need an arg for the closure.

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+        self.symbol = symbol
+        self.contractAddress = contractAddress
+        self.server = server
+    }
+}
+
+protocol CoinTickersResolverType {
+    var tickers: [AddressToPRCServerKey: CoinTicker] { get }
+
+    func coinTickers(enabledObject: [TokenMappedToTicker]) -> Promise<Void>
```
hboon, line 33: Would you rename `enabledObject`? I guess it came from `TokensDataStore`. But not so intention-revealing here.

AlphaWallet/Tokens/Types/TokensDataStore.swift
hboon: Actually, if we don't intend to act on the promise, maybe `coinTickers(enabledObject:)` should just have no return value?

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+        guard !isFetchingPrices else {
+            return .init(error: AnyError.alreadyFetchingPrices)
+        }
+
+        let config = self.config
+
+        isFetchingPrices = true
+
+        return firstly {
+            fetchCoinTikerIds()
```
hboon, line 174: Typo.

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+            self.contractAddress = contractAddress
+            self.server = server
+        }
+    }
+
+    func coinTickers(enabledObject: [TokenMappedToTicker]) -> Promise<Void> {
+        Promise<Void> { seal in
+            coinTickersPromise(enabledObject: enabledObject).done { [weak self] tickers in
+                self?.tickers = tickers
+                self?.loadChartHistoryForResolverTickers()
```
hboon, line 104: Does this mean that this shouldn't be public, since it can be called externally (again)?

```
    func chartHistory(force: Bool, period: ChartHistoryPeriod, for key: AddressToPRCServerKey) -> Promise<ChartHistory> {
```

AlphaWallet/Core/CoinTicker/CoinTickersResolver.swift
```diff
@@ -0,0 +1,243 @@
...
+                return .value(self.tickers)
+            })
+        }).ensure(on: CoinTickersResolver.queue, { [weak self] in
+            self?.isFetchingPrices = false
+        })
+    }
+}
+
+private extension CoinTickersResolver {
+    struct CoinTickerId: Codable {
```
hboon, line 210: How about just `Ticker`? The `Id` doesn't communicate its purpose well, `Ticker` isn't perfect, but it's a private type, so..

hboon (changes requested):
Is PR purely a backend change, fetching from CoinGecko, but not displaying them in the UI?

Seems like I hit a 429/rate limit quite easily with CoinGecko API, just the 2nd time I launched the app. I think I had about `6000` instances `CoinTickerId` in an array. Since it's 1 call per token, maybe we can debounce (`RateLimiter`?) the calls and perhaps cache them?

I'll continue checking out the code again after we have a fix for that? :)

hboon (commented):
I think we still need to debounce the requests with `RateLimiter` :)

Screenshot 2021-04-01 at 5 03 45 PM

vladyslav-iosdev:
> Is PR purely a backend change, fetching from CoinGecko, but not displaying them in the UI?

if u mean about fetching charts- yes.

AlphaWallet/Core/Types/RateLimiter.swift
```diff
@@ -46,3 +46,62 @@ class RateLimiter {
...
+
+        execute {
+            value = block()
+        }
+
+        return value
+    }
+}
+
+public final class TimedLimiter: SyncLimiter {
```
hboon, line 67: This seems similar to `RateLimiter`. Are there any differences?

AlphaWallet/Core/Types/RateLimiter.swift
```diff
@@ -46,3 +46,62 @@ class RateLimiter {
...
+
+    // MARK: - Properties
+    public let limit: TimeInterval
+    public private(set) var lastExecutedAt: Date?
+
+    private let syncQueue = DispatchQueue(label: "com.samsoffes.ratelimit", attributes: [])
```
hboon, line 73: Remember to change the label if we keep this :)

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
```diff
@@ -0,0 +1,310 @@
...
+                CoinTickersFetcher.queue.asyncAfter(deadline: .now() + Self.fetchChartHistoryTimeLimit + 1) {
+                    strongSelf.fetchChartHistoryForAlearyResolvedTickers()
+                }
+            }.cauterize()
+        }
+    }
+
+    private func loadAllPeriodsHistory(for key: AddressToPRCServerKey) -> Promise<Void> {
+        return Promise<Void> { seal in
+            let group = DispatchGroup()
```
hboon, line 136: If we keep this, can we abstract this line (and related lines) that are more management-related plumbing, then logic?

```
let group = DispatchGroup()
```

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
```diff
@@ -0,0 +1,310 @@
...
+            }
+        }.catch { _ in
+            //no-op
+        }
+    }
+
+    private static let fetchChartHistoryTimeLimit: TimeInterval = 5
+    private let loadChartHistoryLimiter = TimedLimiter(limit: CoinTickersFetcher.fetchChartHistoryTimeLimit)
+
+    func fetchChartHistoryForAlearyResolvedTickers() {
```
hboon, line 121: Typo: "Already"

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
hboon: Maybe we can use `when(resolved:)` here instead too. Higher level.

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
```diff
@@ -0,0 +1,310 @@
...
+
+        var valueToSet = self.history[ticker] ?? [:]
+        valueToSet[period] = result
+
+        self.history[ticker] = valueToSet
+        return .value(result)
+    }
+
+    func fetchChartHistory(force: Bool, period: ChartHistoryPeriod, for key: AddressToPRCServerKey) -> Promise<ChartHistory> {
+        return firstly {
```
hboon, line 176: Can you describe how the fetching and caching works? I'm especially interested if we are caching and especially for data that is not `ChartHistoryPeriod.day` and when/how we purge the cache.

Caching aside, do we make 5 calls per token? (1 for each case of `ChartHistoryPeriod`)?

I reckon the API rate limits by IP address, so in an event, like a conference, we might only have a very small set of IP addresses with several tends or hundreds of devices :)

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
hboon: Is the logic here like this?:

```
//This runs not more often than every 5 secs
func fetchChartHistoryForAlearyResolvedTickers() {
    foreach token {
        fetch immediately
    }
    when all done, wait 6 secs
    fetchChartHistoryForAlearyResolvedTickers()
}

fetchChartHistoryForAlearyResolvedTickers()
```

Podfile.lock
```diff
@@ -326,4 +326,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 82fcf9b1524e8d73c1036e7362d98856b45da675
 
-COCOAPODS: 1.10.1
```
hboon, line 329: Remember to run `bundle` first

AlphaWallet/Core/Types/RateLimiter.swift
vladyslav-iosdev: yep it acts in same way as RateLimiter, but this one avoids using timer creating, (and that is more expensive operation).

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
vladyslav-iosdev: `Caching aside, do we make 5 calls per token? (1 for each case of ChartHistoryPeriod)?` yes, we make api calls to fetch history values.  one time per 5 seconds for single server. Once it done we cache fetched values in dictionary.

vladyslav-iosdev: but, seems like its need to be updated. because when we have 5 or more enabled networks it will cause to many requests.

AlphaWallet/Core/Types/RateLimiter.swift
hboon: I see. If it does the same thing but the implementation is better, would you replace the internals of RateLimiter in a separate commit/PR and use it still? No point having 2 classes if they do the same thing.

hboon:
Remember to buzz me if you want a review again

vladyslav-iosdev:
Sorry while working on bartercard i totally missed about this issue. Wanted to ask u a question: do we need to handle token prices update on send token screen and on token screen?

vladyslav-iosdev:
i disabled fetching for history for now.

hboon:
>Sorry while working on bartercard i totally missed let’s meabout this issue. Wanted to ask u a question: do we need to handle token prices update on send token screen and on token screen?

@vladyslav-iosdev
>Sorry while working on bartercard i totally missed about this issue. Wanted to ask u a question: do we need to handle token prices update on send token screen and on token screen?

@vladyslav-iosdev let’s just only fetch but not display the prices in the UI in this PR, so no changes to the UI, basically what you have done so far (the last time I reviewed). Both to keep it more self-contained and also I’m wondering if the rate limits will affect how we offer the price data to the UI. So let’s work that out first.

hboon:
If you are busy with Bartercard, I can work on tweaking the price fetching next week and you can take over the UI (if you have the time, later). Let me know.

vladyslav-iosdev:
> If you are busy with Bartercard, I can work on tweaking the price fetching next week and you can take over the UI (if you have the time, later). Let me know.

sorry i missed your comment, yep, currently in work with auth0, so u can work on it.

hboon:
@vladyslav-iosdev no problem. I'll look into it. You figured out most of it already any way. Thanks.

hboon:
I've tweaked this so:

* list of token IDs supported by CoinGecko is only fetched once and shared across chains
* current price of tokens is only fetched once and shared across chains
* chart history of tokens is now cached
* chart history of tokens is no longer fetched automatically (so it is not fetched in this PR)
* added a TODO to call to fetch chart history of tokens for a single token when the token screen is shown

@vladyslav-iosdev would you help to take a look? Can't add you as reviewer since you are the author of this PR. If this is good, we can merge it and then move on to work on the UI.

PR #2714, opened 2021-04-16
Make enabling testnets and mainnets be exclusive

### LINKED ISSUE #2710: Make 'Testnet' more obvious and separated from 'Mainnet'
**Clickable prototype:**
https://invis.io/GU10Q3YYSZB7

**Zeplin:**
https://zpl.io/254Zpjo

1. You have a toggle to enable Testnets. This toggle disables Mainnet and enable Testnet. And other way around. Only one can be enabled at the time. 
2. After you hit the toggle, it warns you about what testnets are
4. We say that they have no value at all time.

testnet_flow

hboon:
Closes #2710 

This is to prevent users from confusing them and hopefully lessen the possibility of users being scammed by thinking tesetnet balances have non-zero fiat value.


hboon:
A TODO in #2716 for some follow up tweaks after Samoa

vladyslav-iosdev (approved):

PR #2715, opened 2021-04-16
add mushrooms finance to alpha wallet iOS

mushroomsforest:


hboon (approved):
Thanks for contributing.

mushroomsforest:
> 
> 
> Thanks for contributing.
> 
> Unfortunately the app store forbade us from listing dapps, so this list is not currently used. It is still maintained in this repo, but in the meantime this will not be visible in the app.

so for iOS alpha wallet users, how do they interact with dapps?

hboon:
@mushroomsforest we don't display the list in the AlphaWallet iOS, but the users can still enter the URL into the dapp browser and use them as expected.

PR #2739, opened 2021-04-30
POC swift JS Core. Loading of scripts, libraries and Node Modules

nicktaras:
Closes #

https://trello.com/c/vqRdsZQS/98-swift-node-js-complete-task-to-convert-node-js-crypto-module-into-a-swift-playground-file-poc-complete-using-a-non-minified-vers

Added a Project Directory with Playground Proof of Concept to load Javascript / Node JS modules into Swift.


hboon:
@nicktaras Ah, can I say that the main thing to get it to work is to prepend the extra "namespace" (`nodejs`) introduced generated while building, requiring no changes to the JavaScript code and JavaScript build scripts?

```
context.evaluateScript("nodejs.crytpo.Cipher.toString()")
```

vs what we tried previously:

```
context.evaluateScript("CryptoJS.HmacSHA1('Message', 'Key'))")
```

nicktaras:
@hboon - Yep, that's correct. Browserify was key to create the js bundle, but also expose the namespace 'node' as the entry point to the library.

hboon:
@nicktaras thanks! Do we already have the JavaScript code to verify TokenScript signature so we can embed and try it out? Otherwise we can close this for now.

nicktaras:
Hi @hboon, sorry for the late reply. Do you mean logic to check the tokenScript xml node that contains the signature?

hboon:
@nicktaras the logic that extracts the signature and verify that it’s valid. I don’t know if it’s been implemented already. This is why we are doing this right? :)

hboon:
Closing for now

PR #2746, opened 2021-05-11
Decode ERC20 approval transactions #2394

### LINKED ISSUE #2394: Decode ERC20 approval transactions
Look for `DecodedFunctionCall.FunctionType.erc20Transfer`, which decodes ERC20 transfers.

This can be tested through transactions posted from dapp and WalletConnect

### LINKED ISSUE #2751: Decode transactions when downloading them from Etherscan
Continuing from #2746

Noticed we forgot to do this earlier and was only decoding transactions while we send them.

It is handled in `TransactionInstance.createOperationForTokenTransfer()`.

Replacing the bulk of the code with something like this should help:

```
//hhh forced unwrap (fix)
let (operations: operations, isErc20Interaction: isErc20Interaction) = Transaction.decodeOperations(fromData: Data(hex: transaction.input), from: AlphaWallet.Address(string: transaction.from)!, contractOrRecipient: AlphaWallet.Address(string: transaction.to), tokensDataStore: tokensStorage)
if !operations.isEmpty {
    NSLog("xxx2 was decoded as \(operations[0].type) for hash: \(transaction.hash)")
} else {
    NSLog("xxx2 not decoded for hash: \(transaction.hash)")
}
var operations2: [LocalizedOperationObjectInstance] = operations.map { LocalizedOperationObjectInstance(object: $0) }
return .value(operations2)
```

The result should show up as an Activity in Activities tab (and transaction in Transaction tab) and show an appropriate Activity screen when tapped.

vladyslav-iosdev:
Closes #2394, Closes #2751


hboon:
@vladyslav-iosdev I often do this by hunting for the right token on Etherscan, picking a wallet and then watch that wallet. Eg. Look for DAI/SAI token.

hboon (approved):
Looks good. I have added 2 commits to clean things up, you can merge after squashing them if they look good.

Noticed we forgot to do this earlier and was only decoding transactions while we send them, so something to follow up: #2751

hboon (approved):

PR #2752, opened 2021-05-12
Add Levinswap to Dapps.Swift Alphawallet

LucaSilverhand:
Add Levinswap to Dapps.swift


hboon:
@LucaSilverhand thanks for contributing. I have rebased the PR and removed the unnecessary (file mode) changes. I can't push to your fork so I have created #2797.

Unfortunately the app store forbade us from listing dapps, so this is no longer used. This list is still maintained in this repo, but in the meantime this will not be visible in the app.

PR #2753, opened 2021-05-12
Build the Info tab for fungibles (native crypto + ERC20) and move the most recent 3 transactions into the Activity tab #2748

### LINKED ISSUE #2748: [Samoa] Build the Info tab for fungibles (native crypto + ERC20) and move the most recent 3 transactions into the Activity tab
No alert tab.

Original issue: #2629

Refer to umbrella issue: #2747

### LINKED ISSUE #2749: [Samoa] Modify the Activity sub-tab in token screen to show relevant Activities instead, with the correct UI and data
Refer to umbrella issue: #2747

vladyslav-iosdev:
Closes #2748, closes #2749


hboon (commented):
Looks great! A couple of questions:

* What if there is no price data for a particular token?
* What if price data failed to load?
* The labels like "ath" etc should need to be tweaked (I know it's a WIP)
* When the user touch down and drag their finger over the charts, is it possible to short the price and date/time at that point?

hboon:
@vladyslav-iosdev BTW, before it's too late, do you want to keep the old `TokenViewController` around, if it doesn't complicate the code much? I'm thinking it might be easier for the other project when we sync again.

vladyslav-iosdev:
> * What if there is no price data for a particular token?
> * What if price data failed to load?

for now it will display all fields with "-" values, the same for charts, 'no chart data' will be displayed.

vladyslav-iosdev:
@hboon as for me, current implementation for tickers, isn't perfect, would be better to update it as subscribable values.

vladyslav-iosdev:
> * When the user touch down and drag their finger over the charts, is it possible to short the price and date/time at that point?

sure, i will check it.

hboon:
>@hboon as for me, current implementation for tickers, isn't perfect, would be better to update it as subscribable values. 

@vladyslav-iosdev yes. I didn’t change it because we could see how the UI works. Can do it later. It’s ok.

vladyslav-iosdev:
@hboon can u provide more detailed information, what fields from ticker do we need to display, and what formatting specification for them?

hboon:
@vladyslav-iosdev Looking at https://projects.invisionapp.com/share/7TZHRQ8W265#/screens/438941755, I think you can ignore:

* 24-H return
* Profit/Loss
* portfolio share
* average cost
* paid fees

They can't be computed unless we have the purchase data.

We should be able to get market cap and total supply if it isn't already available in the CoinGeck calls we made.

Let me know if you need more?

vladyslav-iosdev:
@hboon can u describe how to resolve perfomance fields? what logic for their calculation should be

hboon:
How about take the oldest price of each period - current?

vladyslav-iosdev:
> How about take the oldest price of each period - current?

ok, i got it. also what about percentage value for it? how can i get it?
Screenshot 2021-05-17 at 09 33 58

hboon:
Current / the_period_oldest_value * 100. Is that what are referring to?

vladyslav-iosdev:
> Current / the_period_oldest_value * 100. Is that what are referring to?

thx, yes

hboon:
Remember to buzz me if you want a review again.

vladyslav-iosdev:
@hboon during developing i found that creating XMLHandler takes to many time, can we cache  XMLHandler instances ?  does it can brake any logic?

hboon:
`XMLHandler` is just a wrapper around `PrivateXMLHandler`. The latter is already cached. Creating `XMLHandler` shouldn't take too much time.

vladyslav-iosdev:
> `XMLHandler` is just a wrapper around `PrivateXMLHandler`. The latter is already cached. Creating `XMLHandler` shouldn't take too much time.

ok, i got it. Can u check this PR?

hboon:
Ok!

hboon:
Looks good except the numbers in the screenshot look strange. For example 1 day is "76.83 USD" maybe, then it can't be 97.01%, because ETH is around 2500-2600 USD these 2 days. Similarly, it was around 2300-2800 the past 7 days, and so on


Screenshot 2021-06-10 at 11 57 54 PM

vladyslav-iosdev:
@hboon check this pr,  values should be ok now.

hboon (changes requested):
- [ ] Can you describe how the Performance 1 Year is computed with numbers (you can make some up) as example? (I'm still wondering what it should actually show, not only what is being shown in the PR)
- [x] I think the fonts like "Stats" and "Market Cap" doesn't seem to match the design
- [x] Can you change format the Market Cap 248392873791 to become 248.39B? Maybe hardcode (or if there's a way to do it with formatters) so that B(illion), M(illion) are supported? And < 1M as it is, but all rounded to 2 decimal places since it's fiat
- [x] Let's use "USD" instead of "$" for Market Cap like we do for 1 Year Low/High (I know the design is a bit inconsistent there).
- [x] The token icon should be smaller

hboon:
BTW, we should be using `1.10.1`. You'll have it if you run `bundle` to install gems in the project root.

vladyslav-iosdev:
> * Can you describe how the Performance 1 Year is computed with numbers (you can make some up) as example? (I'm still wondering what it should actually show, not only what is being shown in the PR)

we computing value change according the values we got from the api.  here is description how it works

hboon (changes requested):
Ok, I'm not too sure about the computations, but they don't look blantantly wrong. Could you help remove/hide the "Total Supply" and "Max Supply" rows? But we should figure out how to add them in a separate PR.

vladyslav-iosdev:
> Total Supply 

as well as `Max Supply` we are getting from json, without any calculation on our side.

vladyslav-iosdev:
@hboon Bump

hboon (approved):
Can you help rebase master and merge?

vladyslav-iosdev:
@hboon i left pr unsquashed for reasons that this pr is too big, and it will be easier to find commits with problems

PR #2759, opened 2021-05-14
Fix: detecting of transactions for ERC20 approvals broken when downloading transaction history

hboon:
* Also fix naming "sender" -> "spender" for ERC20 approvals


AlphaWallet/EtherClient/TrustClient/Models/RawTransaction.swift
```diff
@@ -121,13 +121,13 @@ extension TransactionInstance {
             }
         }
 
-        if let data = transaction.input.data(using: .ascii), let functionCall = DecodedFunctionCall(data: data) {
```
hboon, line 124: Key change is here.

PR #2766, opened 2021-05-17
Back button for fungible token screen doesn't work like native #2765

### LINKED ISSUE #2765: Back button for fungible token screen doesn't work like native
In the fungible token screen (screenshot), tapping the back button doesn't show a list of (1) view controllers in the navigation stack. Also panning from the left edge of the screen doesn't go back.

It looks like it's because of https://github.com/AlphaWallet/alpha-wallet-ios/blob/5aefc4eee83bf4e0926031ecfcc0b54f5ab6bec9/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift#L400

It looks like it's for ee701f1959384b81cf3de777c16d13e3ea5afe06. Do we still need it?

vladyslav-iosdev:
Fixes #2765

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift
@@ -398,7 +398,7 @@ class SingleChainTokenCoordinator: Coordinator {
         viewController.configure(viewModel: viewModel)
 
         viewController.navigationItem.leftBarButtonItem = UIBarButtonItem.backBarButton(selectionClosure: {
-            navigationController.popToRootViewController(animated: true)
+            navigationController.popViewController(animated: true)
         })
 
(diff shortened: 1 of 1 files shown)


hboon:
Haven’t test yet, but does this actually make pan/swipe to go back as well as tap and hold the back button work?

vladyslav-iosdev:
> Haven’t test yet, but does this actually make pan/swipe to go back as well as tap and hold the back button work?

oh, you right, its not, but as i know it disabled for whole application. do i need to enable it back? it could cause the new issues in the app

hboon:
Ah. Nah. Just keep it. Was hoping the fix wasn’t necessary anymore. But when you have the time and inclination, you can help look into it again. No rush. Just not natural to not be able to swipe back :)

hboon:
Closing for now.

PR #2784, opened 2021-05-21
Pasting something into AddressTextField should clear error state displayed if it's not an error #2781

### LINKED ISSUE #2781: Pasting something into AddressTextField should clear error state displayed if it's not an error
eg. paste something and it may show "Empty Clipboard" or "Invalid Address", then if we paste a valid address, the red text should disappear.

1. Paste non-address
2. "Invalid Address"
3. Paste valid address

Expected:

4. "Invalid Address" disappears
5. ENS name appears

Observed:

4. "Invalid Address" still remains
5. ENS name appears

vladyslav-iosdev:
Closes #2781


hboon (approved):
Being liberal with approval. Assuming you tested :)

PR #2790, opened 2021-05-21
[Samoa] New Wallet summary and selection screen #2756

### LINKED ISSUE #2756: [Samoa] New Wallet summary and selection screen
- [ ] Accessible from the blockie in the Wallet tab
- [ ] Also accessible from Settings > "Change / Add Wallet"
- [ ] Update look for list of wallets
- [ ] Show total amount, summing up all wallets (this might be tricky? If so maybe, have a PR that sums up the total first, without UI?). Requires #2802

https://app.zeplin.io/project/5d0356be2f25b319dcc5944b/screen/6045ef3df032b81670d585f4

vladyslav-iosdev:
Closes #2756


AlphaWallet/AppCoordinator.swift
```diff
@@ -175,8 +205,25 @@ class AppCoordinator: NSObject, Coordinator {
...
+    }
+
+    //NOTE: Not sure how to call it right, rename with more appropriate name.
+    private enum AccountsStartPoint {
+        case withAccountsPresentation
+        case skipAccounts
+    }
+
+    @discardableResult private func showTransactions(wallet: Wallet, animated: Bool, some: AccountsStartPoint = .withAccountsPresentation) -> InCoordinator {
+        switch some {
```
hboon, line 219: Can we make this not reload the UI if the current account is tapped? Something like https://github.com/AlphaWallet/alpha-wallet-ios/blob/f8f132f5ecfb7cfde5d640996d40c9da3ad5d2c4/AlphaWallet/Settings/Coordinators/SettingsCoordinator.swift#L223

hboon (commented):
Looks good. So currently, this changes the flow, but the contents screens are unmodified, right?

vladyslav-iosdev:
> Looks good. So currently, this changes the flow, but the contents screens are unmodified, right?

yep, currently i am fixing memory leaks. once i done i'll start on modifying screen

hboon (commented):
Could you make the percentages (the % symbol) align on their right? (because the check and chevron have different widths).

And the fonts too?

Looks good otherwise. Thanks!

Screenshot 2021-05-25 at 3 49 30 PM

hboon (approved):
Oh. Had to remove the percentage because we can't compute for inactive wallets until #2802 right?

Looks good!

vladyslav-iosdev:
> Oh. Had to remove the percentage because we can't compute for inactive wallets until #2802 right?

yes, you right. #2802 is ready but i need #2756 to finish it. Is #2802 is ok to merge?

hboon (approved):
I'll check out #2802. But don't worry about it. If you are ok with this PR as it is now, you can merge it first in case #2802 has more changes.

vladyslav-iosdev:
> BTW, would you check if cancelling an import wallet works in the updated list of wallets screen? 

yep, seems like it doesn't work correctly.

PR #2791, opened 2021-05-23
Add swap and contract execution grouping in Activity tab

hboon:
Show a "Swap" placeholder and "Contract Execution" placeholder in Activity tab. Swaps are groups of transctions which include at least 1 receive and 1 send transaction.

Several follow-ups needed:

* Performance tweak to run some processing in a background queue has been disabled to make app not crash. This has to be re-introduced
* The number of events used for the tab has been reduced from 1000 to 100 to keep performance barely acceptable
* Upon initial import of wallet, events aren't always shown in the display right away
* Most of the "Sent ETH" with "-0 ETH" should also be "Contract Execution"(s)

This isn't the final form (short-term) because not all the activities are shown properly due to performance.

@colourfreak FYI. Added an ugly "Contract Execution" and "Swap" header in Activity tab for now. We have to improve this visually. But wanted to get this into the code base and out of the way first and not block progress.


hboon:
Forgot to check for my "hhh". Good for review again.

vladyslav-iosdev (approved):

hboon:
Forced push to fix a syntax error:

```
let isSwap = isSwap(...
```

Xcode 12.4 (which we use for Travis) doesn't like this, but it builds with 12.5.

PR #2806, opened 2021-05-25
[Samoa] Compute fiat value for each wallet while they are inactive #2802

### LINKED ISSUE #2802: [Samoa] Compute fiat value for each wallet while they are inactive
Not sure how it's possible:

We can either do it

A. The true/right way, be able to compute fiat for each inactive wallet. Seems like a major change
B. Cache the tokens (based on CoinGecko's token ID) and their count somewhere, say a .json file and be able to compute even if it is inactive.

This is going to be used:

- [x] to display the total (across all wallets) in Wallet tab for follow up to #2755
- [x] display total in "list of wallets" tab #2756

vladyslav-iosdev:
Closes #2802

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -766,6 +766,7 @@
 		876C80CF267398F900B16595 /* AddRPCServerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 876C80CE267398F900B16595 /* AddRPCServerViewController.swift */; };
 		876C80D12673992300B16595 /* AddRPCServerViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 876C80D02673992300B16595 /* AddRPCServerViewModel.swift */; };
 		87713EAE264ABF2800B1B9CB /* DecodedFunctionCall+Decode.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87713EAD264ABF2800B1B9CB /* DecodedFunctionCall+Decode.swift */; };
+		8777A47E26660CAA005285BD /* PrivateTokensDataStoreType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8777A47D26660CAA005285BD /* PrivateTokensDataStoreType.swift */; };
 		877D00AF25ADF60A008E22CC /* TransactionConfiguratorTransactionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 877D00AE25ADF60A008E22CC /* TransactionConfiguratorTransactionsTests.swift */; };
 		8782035D2431E66600792F12 /* FilterTokensCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8782035C2431E66600792F12 /* FilterTokensCoordinator.swift */; };
 		8782035F2431FBC300792F12 /* ShowAddHideTokensViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8782035E2431FBC300792F12 /* ShowAddHideTokensViewModel.swift */; };
@@ -791,6 +792,10 @@
 		87A3022724C02212000DF32E /* TransactionConfirmationViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87A3022624C02212000DF32E /* TransactionConfirmationViewController.swift */; };
 		87B651F7256D4BFE000EF927 /* ClaimPaidOrderCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B651F6256D4BFE000EF927 /* ClaimPaidOrderCoordinator.swift */; };
 		87B6526A256FBF36000EF927 /* SwapTokenService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B65269256FBF36000EF927 /* SwapTokenService.swift */; };
+		87BB63DC265CFF2700FF702A /* WalletBalanceCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BB63DB265CFF2700FF702A /* WalletBalanceCoordinator.swift */; };
+		87BB63DE265E21C900FF702A /* WalletSummary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BB63DD265E21C900FF702A /* WalletSummary.swift */; };
+		87BB63E0265E21E400FF702A /* WalletBalance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BB63DF265E21E400FF702A /* WalletBalance.swift */; };
+		87BB63E2265E759700FF702A /* PrivateBalanceFetcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BB63E1265E759700FF702A /* PrivateBalanceFetcher.swift */; };
 		87BBF9862563DD7600FF4846 /* WalletConnect+UserDefaults.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BBF9692563DD7500FF4846 /* WalletConnect+UserDefaults.swift */; };
 		87BBF9872563DD7600FF4846 /* WalletConnectSessionDetailsViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BBF96B2563DD7500FF4846 /* WalletConnectSessionDetailsViewModel.swift */; };
 		87BBF9882563DD7600FF4846 /* WallerConnectRawViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87BBF96C2563DD7500FF4846 /* WallerConnectRawViewModel.swift */; };
@@ -810,6 +815,7 @@
 		87C10A6525ED1105008E9B1B /* SignatureConfirmationDetailsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87C10A6425ED1105008E9B1B /* SignatureConfirmationDetailsViewController.swift */; };
 		87C447AA255970C5009DF2D2 /* ActiveWalletSessionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87C447A9255970C5009DF2D2 /* ActiveWalletSessionView.swift */; };
 		87C650C325F2408E007B02CB /* ServerUnavailableCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87C650C225F2408E007B02CB /* ServerUnavailableCoordinator.swift */; };
+		87C65F532660DD2E00919819 /* WalletBalanceFetcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87C65F522660DD2E00919819 /* WalletBalanceFetcher.swift */; };
 		87C8018C24350174007648CF /* AddHideTokenSectionHeaderViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87C8018B24350174007648CF /* AddHideTokenSectionHeaderViewModel.swift */; };
 		87CA8495253DDFF200BF8443 /* TransitionButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CA8490253DDFF100BF8443 /* TransitionButton.swift */; };
 		87CDD49025F100AE0009B6AC /* Array.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87CDD48F25F100AE0009B6AC /* Array.swift */; };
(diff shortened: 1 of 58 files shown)


hboon (changes requested):
@vladyslav-iosdev could you describe, briefly, the approach?

There's some build errors.

vladyslav-iosdev:
Yes, sure. We are using subscribable wallets model to handle wallets chaning. it fires once we add, remove wallets.
For each Wallet we create fetcher object that performs fetching wallet balance for each server (it also possible to set only for `.main` if needed). `BalanceFetcher` has subscribable balance model, that triggers once we get wallet balance. Time interval for balance fetching `15 secs`. Every time when we got the wallets balance we combine it with fetched tickers. We fetch tickers for native crypto for each server. Combined balance and ticker become to `WalletBalance` where we can get its right usd value. (what according to the apprecation value, i guess we can get only value for `.main` server for it.) The `walletSummary` is sum of `WalletBalance` fetched in prev steps. `WalletBalanceCoordinator` allows user to subscribe for specific wallets balance, and also for wallet summary.

vladyslav-iosdev:
> @vladyslav-iosdev could you describe, briefly, the approach?
> 
> There's some build errors.

i will check.

vladyslav-iosdev:
> There's some build errors.

Oh, i missed to compile the project once refactored code

hboon:
Oh. That means it only checks the fiat value for the native crypto, but not other tokens, especially ERC20s?

hboon:
If we are displaying a total, we'll want to include especially the ERC20s. That's why I mention there might be complications if we try to read all the databases for each inactive wallet to compute this total. Sorry, I wasn't clear earlier.

vladyslav-iosdev:
> Oh. That means it only checks the fiat value for the native crypto, but not other tokens, especially ERC20s?

yes, right. let check this one more time.

hboon:
You can go ahead and merge once you fix it. So I don't block you.

AlphaWallet/Core/Coordinators/WalletBalanceCoordinator.swift
```diff
@@ -0,0 +1,297 @@
...
+}
+
+extension Wallet: Hashable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(address.eip55String)
+    }
+}
+
+private class WalletBalanceFetcher: NSObject {
+    private static let updateBalanceInterval: TimeInterval = 15
```
hboon, line 191: Shall we extend this to a larger value, maybe 60s? Worried that it will blow the limits if there are many wallets with many tokens.

hboon (commented):
Looks good generally. I didn't go into the details, but we can get this merged in once you are done and fix any additional problems later.

BTW, don't forget:

> //TODO: Rename

hboon:
I did notice that many classes have to be introduced though. I'm thinking we should introduce more boundaries — by breaking up some of our code into modules. Do you have experience with using SwiftPM in the project that already use CococaPods (like ours). Any gotchas?

hboon:
I asked and someone said:

> as long as what you are putting in the swift packages does not depend on any CocoPods works fine

I forgot about this case. We would need this. So maybe SPM is out?

AlphaWallet/Core/Coordinators/WalletBalanceCoordinator.swift
vladyslav-iosdev: sure we can do that. as all actions performs in background it shouldn't load the CP too mutch.

vladyslav-iosdev:
> I did notice that many classes have to be introduced though. I'm thinking we should introduce more boundaries — by breaking up some of our code into modules. Do you have experience with using SwiftPM in the project that already use CococaPods (like ours). Any gotchas?

yes i have (but not too much :). What parts do you want to move to separate module?

vladyslav-iosdev:
@hboon can u provide me more details what this apprecation/deprecation value is? and how to calcutate it? and the same for wallet summary
Screenshot 2021-05-28 at 10 52 30

hboon:
> yes i have (but not too much :)

Oh, I'm not blaming the PR, but the code base currently encourages this (not always bad) and don't enforce the boundaries (so that other code would be too tightly coupled with the classes which are private to this new functionality).

> What parts do you want to move to separate module?

Not within this PR, but I think would be good if for example, a PR like this would have the bulk of the code in a separate module and only minimal (sometimes still changing many classes/files) changes to other parts of the code. i.e. details of fetching the balances might be in the new module, but displaying is in the main code base. Naturally you can't do that yet, I think there needs to be quite a bit of refactoring in the entire code base. Created #2816

vladyslav-iosdev:
> > @hboon can u provide me more details what this apprecation/deprecation value is? and how to calcutate it? and the same for wallet summary
> 
> Oh, let's take it as compared to `1d` ago. And tweak it if we have feedback, i.e.
> 
> ```
> fiat1dAgo = sum(tokenN * price1dN)
> fiat = sum(tokenN * priceN)
> (fiat - fiat1dAgo) / fiat1dAgo * 100%
> ```

as i understand for this we also need to fetch change histore for each token to get (1d ago) value, right?

hboon:
Oh. How did we display the change for each token in the Wallet tab? If there’s a value embedded in that single call, let’s just use that instead.

vladyslav-iosdev:
> Oh. How did we display the change for each token in the Wallet tab? If there’s a value embedded in that single call, let’s just use that instead.

so we just need to summarize those appreciation values for each token, right?

hboon:
Yes, taking into account the balance. Is that mathematically possible?

vladyslav-iosdev:
> Yes, taking into account the balance. Is that mathematically possible?

i am not sure, if results could be right. For each token we are displaying its 24 hours appreciation value we received in tiker.

hboon:
Something along the lines of sum(change * balance)/total should work?

vladyslav-iosdev:
> Something along the lines of sum(change * balance)/total should work?

let me try it

hboon (changes requested):
It crashed for me shortly after I tried these 2 ways:

A. git rebase master then run with my current wallets
B. delete the app and run this PR (without rebase) and import a wallet

I can reproduce with with wallet 0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362

> 2021-06-01 12:46:29.979804+0800 AlphaWallet[45995:709556] *** Terminating app due to uncaught exception 'RLMException', reason: 'Attempting to create an object of type 'TokenObject' with an existing primary key value '0x06012c8cf97BEaD5deAe237070F9587f8E7A266d-1'.'
> *** First throw call stack:
> (
> 	0   CoreFoundation                      0x00007fff20422fba __exceptionPreprocess + 242
> 	1   libobjc.A.dylib                     0x00007fff20193ff5 objc_exception_throw + 48
> 	2   Realm                               0x0000000105efaf4d RLMAddObjectToRealm + 829
> 	3   RealmSwift                          0x00000001074eb7a6 $s10RealmSwift0A0V3add_6updateyAA6ObjectC_AC12UpdatePolicyOtF + 1286
> 	4   AlphaWallet                         0x0000000102247e62 $s11AlphaWallet22PrivateTokensDataStoreC9addCustom33_405B4813A88B7C6B938907DB2BA70474LL5token10completionyAA8ERCTokenV_yyctFyycfU_ + 1266
> 	5   AlphaWallet                         0x0000000101ffa820 $sIeg_IeyB_TR + 48
> 	6   libdispatch.dylib                   0x00007fff2010670d _dispatch_call_block_and_release + 12
> 	7   libdispatch.dylib                   0x00007fff201078df _dispatch_client_callout + 8
> 	8   libdispatch.dylib                   0x00007fff2010de15 _dispatch_lane_serial_drain + 715
> 	9   libdispatch.dylib                   0x00007fff2010e98c _dispatch_lane_invoke + 400
> 	10  libdispatch.dylib                   0x00007fff20118f81 _dispatch_workloop_worker_thread + 772
> 	11  libsystem_pthread.dylib             0x00007fff6033f499 _pthread_wqthread + 314
> 	12  libsystem_pthread.dylib             0x00007fff6033e467 start_wqthread + 15
> )
> libc++abi: terminating with uncaught exception of type NSException
> 
> Process finished with exit code 0

hboon (changes requested):
Doesn't look right here.

Notice the total is 8182.05 USD, but the appreciation is much higher at 68041.00 USD

You can try reproducing it with:

0x80cc263cb3fA1bE2AeC748b3811261c1E2a1ba8d
0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362

Screenshot 2021-06-01 at 4 45 20 PM

hboon:
BTW, we only check for chains that are enabled, right?

vladyslav-iosdev:
> BTW, we only check for chains that are enabled, right?

yes, it works for all wallets, for all tokens, for all enabled chains

vladyslav-iosdev:
ok, yep, i also facing that issue, once i done with Bartercard i return back to this

hboon:
Also noticed this in the console:

> [Assert] Unsupported use of UIKit view-customization API off the main thread. -setBackgroundColor: sent to <AlphaWallet.WalletSummaryTableViewCell: 0x7f9ec0ca1c00; baseClass = UITableViewCell; frame = (0 46; 375 111); autoresize = W; layer = <CALayer: 0x6000056affa0>>

hboon:
> ok, yep, i also facing that issue, once i done with Bartercard i return back to this

Sure. No problem.

vladyslav-iosdev:
> setBackgroundColor

ah made updates but not in all places, now it should be ok.

vladyslav-iosdev:
> Notice the total is 8182.05 USD, but the appreciation is much higher at 68041.00 USD

@hboon i calculated apprecation value for each wallet according to sum(change * balance)/total

hboon:
@vladyslav-iosdev could you dump out (only) the applicable values and let's take a look to see if we can figure it out?

vladyslav-iosdev:
> @vladyslav-iosdev could you dump out (only) the applicable values and let's take a look to see if we can figure it out?

can u please explain what do you meant here, (you a talking about calculation wallet price for values that have 24 appreciation value)?

hboon:
I meant can you paste some actual numbers here and we can figure out the calculation, if you want help with it. Otherwise not necessary.

hboon (commented):
Looks good.

Noticed there are 2 minus in the summary section "( - -25.05 %)"

Screenshot 2021-06-08 at 4 30 21 PM

hboon:
Is there anything else you expect to change for the PR?

vladyslav-iosdev:
> Looks good.
> 
> Noticed there are 2 minus in the summary section "( - -25.05 %)"
> 
> Screenshot 2021-06-08 at 4 30 21 PM

i will check double -- and that all, i hope :).

hboon:
We can also make list of wallets screen accessible from Settings use the one from Wallet tab + blockie. But maybe a separate PR? Big enough already :)

vladyslav-iosdev:
> We can also make list of wallets screen accessible from Settings use the one from Wallet tab + blockie. But maybe a separate PR? Big enough already :)

you are talking about this screen?
Screenshot 2021-06-08 at 11 42 43

vladyslav-iosdev:
> We can also make list of wallets screen accessible from Settings use the one from Wallet tab + blockie. But maybe a separate PR? Big enough already :)

or you mean to display the same configuration for `accountsViewController` with summary?

hboon:
> > We can also make list of wallets screen accessible from Settings use the one from Wallet tab + blockie. But maybe a separate PR? Big enough already :)
> 
> or you mean to display the same configuration for `accountsViewController` with summary?

Yeah, this.

vladyslav-iosdev:
@hboon is there some missed things to merge this pr?

hboon:
Not sure. Is this and the comments after it resolved?

> https://github.com/AlphaWallet/alpha-wallet-ios/pull/2806#issuecomment-856574862

vladyslav-iosdev:
> Not sure. Is this and the comments after it resolved?
> 
> > [#2806 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/pull/2806#issuecomment-856574862)

i will double check, but i fixed it before.

hboon (changes requested):
Seeing "NaN" in the screenshot

Screenshot 2021-07-06 at 2 22 55 PM

Other than that, I guess let's get this merged and get some job satisfaction. We might need to tweak the calculation later, but at least it doesn't look blatantly wrong :)

vladyslav-iosdev:
> Seeing "NaN" in the screenshot

ok, let me check NaN first ;)

vladyslav-iosdev:
@hboon now it should be ok

hboon (approved):
Looks good! Can you fix the test suite before merging?

PR #2874, opened 2021-06-08
Disabled eventlog test in test suite to make tests more predictable

hboon:
So that when Travis is red, it's really something wrong.


hboon:
Bump. In case you missed it.

vladyslav-iosdev (approved):

PR #2878, opened 2021-06-08
Still seems to be crashing sometimes when enabling/disable networks #2877

### LINKED ISSUE #2877: Still seems to be crashing sometimes when enabling/disable networks
Observed a few times but can't reproduce. But could be related to previous crashes (which was fixed) which is accessing a `ServerDictionary` with a chain which is no longer around?

vladyslav-iosdev:
Closes #2877


vladyslav-iosdev:
@hboon just take a look what a beautiful cycle. The Mona Lisa in iOS development ;)
Screenshot 2021-06-08 at 15 38 16

hboon:
Hahaa

hboon:
Buzz me if I'm blocking you.

vladyslav-iosdev:
> Buzz me if I'm blocking you.

i found that there also some ref cycles in web3swift trying to resolve it

hboon:
Ah ok, but be careful with that, especially don't break batch calls. e.g. sending multiple calls with 1 RPC calls.

Issue #2890, opened 2021-06-12
ERC721 balance and display issues


zhangzhongnan928:
Wallet: 0xbc8dAfeacA658Ae0857C80D8Aa6dE4D487577c63
Network: rinkeby
721 contract: 0x00000D5B7AaBD11768CC14f8474e5B34f8D39F2B

1. The balance on Etherscan is 13,  on Android is 13, on iOS is 15, on AutographNFT is 14.

2. Many tokens are not displayed on iOS https://user-images.githubusercontent.com/33795543/121791129-dee1d680-cc29-11eb-94c2-a00b27964fab.MOV

3. The 15 Tickers should be 15 AMX or 15 AMXs or 15 AMX Tokens

hboon:
> 3. The 15 Tickers should be 15 AMX or 15 AMXs or 15 AMX Tokens

@zhangzhongnan928 other than the balance being wrong, are you saying the title:

"13 Alchemyy Remix Tokens (AMX)" should be displayed as: "13 AMX" instead?

(or 13 AMXs or 13 AMX Tokens)

zhangzhongnan928:
> > 3. The 15 Tickers should be 15 AMX or 15 AMXs or 15 AMX Tokens
> 
> 
> 
> @zhangzhongnan928 other than the balance being wrong, are you saying the title:
> 
> 
> 
> "13 Alchemyy Remix Tokens (AMX)" should be displayed as: "13 AMX" instead?
> 
> 
> 
> (or 13 AMXs or 13 AMX Tokens)

no the small words of "13 Tickers"
![image](https://user-images.githubusercontent.com/33795543/121824869-99d5a700-ccf2-11eb-9d77-1c737c66d8a6.jpeg)

hboon:
- [ ] #2895
- [ ] #2896
- [ ] #2897

PR #2896, opened 2021-06-14
Fix: NFT images for token instances not showing up if they are SVG

hboon:
Part of #2890 

Converted to use WKWebview for token instance images, removing all conversions from SVG -> UIImage

Might be a performance problem. Will have to try it out.


hboon:
Close issue once all PRs listed in https://github.com/AlphaWallet/alpha-wallet-ios/issues/2890#issuecomment-860487761 merge

vladyslav-iosdev (approved):

PR #2900, opened 2021-06-15
Access ipfs: URLs via https://ipfs.io for ERC721 meta JSON uri and images in them

hboon:
Otherwise the ERC721 images that are ipfs: aren't displayed in the app


hboon:
Bump. In case you missed it.

vladyslav-iosdev (approved):

PR #2901, opened 2021-06-15
Read token instance permalink from ERC721 token for those source directly from smart contract + JSON metadata (i.e not from OpenSea)

hboon:
This was already read from the OpenSea API results. But if the ERC721 token isn't found in OpenSea, we would fallback to the `tokenURI` for the token metadata.

This PR makes it so that we look for the token's permalink in that metadata when we read `tokenURI`directly.


hboon:
Bump. In case you missed it.

vladyslav-iosdev (approved):

PR #3006, opened 2021-08-02
Update the design of Add Custom Token #2995

### LINKED ISSUE #2995: Update the design of Add Custom Token
- The Title should be smaller, only in the Navigation Bar (make it consistent with Add Custom RPC Network style)
- Updated input fields (make it consistent with Add Custom RPC Network style)

https://zpl.io/2GqwqBJ

vladyslav-iosdev:
Closes #2995


hboon (approved):
I'll just trust you for this. But would be good to have a before and after screenshot for such PRs :)

PR #3016, opened 2021-08-03
Move `deleteAllTransactions, resetFetchingState` functions from immutable definition #3015

### LINKED ISSUE #3015: Functional or immutable?
Related to #2918

When we do this:

```
extension SomeClass {
    class functional {}
}

extension SomeClass.functional {
    static func foo() { //... }
}
```

The original intention was so it was easier to tell if a function call modifies state. And also:

> We don't have to make it really be functional programming all the way [1]. But as long as a subset of those functions/methods don't mutate state, we get the same benefits. We then wrap around those with our usual OO code. Consider this an experiment, so you can try writing new code around this style too and see if it works for us. Also feel free to voice out if there's a better way syntatically.

But this example https://github.com/AlphaWallet/alpha-wallet-ios/blob/dc4b94c999fd01a56a11194c8b707e01bd8393c5/AlphaWallet/AppCoordinator.swift#L469 seems to violate that because `deleteAllTransactions()` obviously changes state (in the database). Similarly for `resetFetchingState()`?

```
TransactionsStorage.functional.deleteAllTransactions(realm: Wallet.functional.realm(forAccount: account))
TransactionsTracker.functional.resetFetchingState(account: account, config: config)
```

1. So I suggest we don't group this under the immutable/function class/code so the intention (just code that don't mutate state) is preserved.
2. Also, would it be better if we call the class `immutable` instead of `functional`? Does it convey the intention better?
3.  From (1), I *think* we'll ultimately only call other immutable/functional functions, eg.

```
extension SomeClass.functional {
    static func foo(v1: SomeThing) {
        functional.bar()
        SomeThing.functional.bar(something: v1)
        //Only functional.xxx here otherwise it'd definitely mutating state
    }
}
```

It might not look like that right away since we aren't going to re-organize the entire code base.

@vladyslav-iosdev what do you think? Has it been helpful organising code like that so far? If so, does the 3 points above make sense?

vladyslav-iosdev:
Closes #3015

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -466,8 +466,8 @@ extension AppCoordinator: AccountsCoordinatorDelegate {
     } 
 
     func didDeleteAccount(account: Wallet, in coordinator: AccountsCoordinator) {
-        TransactionsStorage.functional.deleteAllTransactions(realm: Wallet.functional.realm(forAccount: account))
-        TransactionsTracker.functional.resetFetchingState(account: account, config: config)
+        TransactionsStorage.deleteAllTransactions(realm: Wallet.functional.realm(forAccount: account))
+        TransactionsTracker.resetFetchingState(account: account, config: config)
     }
 
     func didCancel(in coordinator: AccountsCoordinator) {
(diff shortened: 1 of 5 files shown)


hboon (approved):
Thanks!

hboon:
Merge when you are ready

Issue #2987, opened 2021-07-26
If no active WalletConnect session, immediately run the camera


colourfreak:
Currently we are displaying a blank screen when you don't have any active session. I think we can immediately trigger the camera to scan.

![2021-07-26 11 18 31](https://user-images.githubusercontent.com/51817359/126965708-e1cbe7e8-5af3-41b5-b378-f9a48af00f77.png)
![2021-07-26 11 19 32](https://user-images.githubusercontent.com/51817359/126965723-d6344744-5263-44f0-9c19-eda55d0abb02.png)

hboon:
>Sure, would you help give #2987 a go? It probably involves a slight change in flow.

@Mac3n this.

Mac3n:
Sure, I will check it

PR #3017, opened 2021-08-03
run the camera on no active walletConnect session #2987

Mac3n:
Fixes https://github.com/AlphaWallet/alpha-wallet-ios/issues/2987#event-5102425399

### DIFF
--- a/AlphaWallet/WalletConnect/ViewController/WalletConnectSessionsViewController.swift
+++ b/AlphaWallet/WalletConnect/ViewController/WalletConnectSessionsViewController.swift
@@ -54,6 +54,11 @@ class WalletConnectSessionsViewController: UIViewController {
         view.addSubview(spinner)
 
         sessionsToURLServersMap.subscribe { [weak self] _ in
+            if self!.sessionsValue.isEmpty {
+                DispatchQueue.main.async {
+                    self?.qrCodeButtonSelected(nil)
+                }
+            }
             self?.tableView.reloadData()
         }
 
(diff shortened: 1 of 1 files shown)


Mac3n:
@hboon Please check if this way is a standard solution in the project.

hboon (changes requested):
Thanks for the PR!

> Please check if this way is a standard solution in the project.

Not quite :)

2 things about the code:

1. `self!` isn't good because when the block fires and self is no longer around (because it's weak), `self!` will crash the app. Sometimes we *know* this wouldn't happen, but it's quite risky to write the code that way because when the code/logic changes, that previously valid assumption might be invalidated.

```
sessionsToURLServersMap.subscribe { [weak self] _ in
    if self!.sessionsValue.isEmpty {
```

2. We generally don't want to use `DispatchQueue.main.async()` and `DispatchQueue.main.asyncAfter(deadline:)` to hack around runloops and animation. (I know I do it too and the codebase has too many of these. We should try to avoid them).

And about the flow:

There are 2 ways to create a WalletConnect session. Camera in the Wallet tab and the `Wallet Connect` row in Settings. This PR breaks the one in the Wallet tab: after scanning a WalletConnect QR code and the user has accepted the connection, the camera screen isn't closed.

I think the way to go about it is to check prior to displaying the view controller with list of sessions: if it is empty, open a connection instead. i.e. the caller for `WalletConnectCoordinator.showSessions()` should check first (please verify this, I only took a quick look).

For WalletConnect testing, I find it quite useful to open a private browser session to https://app.uniswap.org (or any dapp supporting it) and just connecting and then closing the browser.

Issue #3011, opened 2021-08-03
Crash when switching wallet in Settings > Change / Add Wallet


hboon:
Complains about Realm accessed from incorrect thread. Will post if I reproduce again. This is with the tip at 5532de05543c5d7d6d047c7538322504f5ac4042

hboon:
@vladyslav-iosdev would you help check if this is resolved on `master`?

vladyslav-iosdev:
Closing, out of date. Reopen if needed

PR #3028, opened 2021-08-05
Fix: Realm crash because: "Realm accessed from incorrect thread."

hboon:
It doesn't happen immediately since this particular instance is only triggered after event fetching is done (while could take a while) might be the cause of #3011


vladyslav-iosdev (approved):

hboon:
@vladyslav-iosdev a reminder we should move away from thread-dependent code (Realm). So painful.

PR #3050, opened 2021-08-11
Fix: "Add token failure" error when tap to add popular tokens in Add/Hide token screen

### LINKED ISSUE #3045: "Add token failure" error message
name: Bug Report
about: "Add token failure" error message on adding any Ethereum token (not other chains though). On compiled master branch only - works fine on app in App Store

---

1. Tap Wallet tab
2. Tap Add / Hide Tokens
3. Tap an ERC20 token, such as TrueUSD
4. Wait ...
5. There error dialog appears

**Expect**
The token to be added to the token list that shows on app startup

latest master branch

IMG_8C97466C0B1E-1

hboon:
Fixes #3045 

Problem is the popular token list needs to be filtered for enabled chains.

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/AddHideTokensCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/AddHideTokensCoordinator.swift
@@ -52,7 +52,9 @@ class AddHideTokensCoordinator: Coordinator {
         navigationController.pushViewController(viewController, animated: true)
 
         popularTokensCollection.fetchTokens().done { [weak self] tokens in
-            self?.viewController.set(popularTokens: tokens)
+            guard let strongSelf = self else { return }
+            let tokensForEnabledChains = tokens.filter { each in strongSelf.config.enabledServers.contains(each.server) }
+            self?.viewController.set(popularTokens: tokensForEnabledChains)
         }.cauterize()
     }
 


vladyslav-iosdev (approved):

benslaney (commented):
This does not appear to fix it for me. The same error is generated at the same time

hboon:

hboon:
@benslaney That looks like a different problem (it is a different error object). If you trace it, it looks like it might be triggered at https://github.com/AlphaWallet/alpha-wallet-ios/blob/284901974c8a8bbf23856ab406f6413213687baf/AlphaWallet/Tokens/Coordinators/SingleChainTokenCoordinator.swift#L325

With these 2 cases which invokes the `completion` block with a nil. The error reporting should be improved, but is there any chance there is a loss of connectivity when you tap to add the popular token you tapped or the Infura call is rate-limited?

```
case .delegateTokenComplete:
    strongSelf.storage.add(delegateContracts: [DelegateContract(contractAddress: contract, server: strongSelf.server)])
    completion(.none)
case .failed(let networkReachable):
    if let networkReachable = networkReachable, networkReachable {
        strongSelf.storage.add(deadContracts: [DeletedContract(contractAddress: contract, server: strongSelf.server)])
    }
    completion(.none)
```

If you still experience this, would you help create a separate issue, mentioning the contents of https://github.com/AlphaWallet/alpha-wallet-ios/pull/3050#pullrequestreview-727690579 and we'll hunt this down?

benslaney:
This is the same as the error I was originally trying to report. When I tried to trace it in the first instance it was indeed tracing to some Coordinator.swift file

And yes, it does look like an API rate limit / timeout issue. However, it consistently does not work for me. I figured you might just need to increase the limit on your API or change one of the parameters so it allows more IP addresses, etc

benslaney:
I just tested this from scratch to be sure and it's there

To reproduce

clone master
change the bundle identifier, groups, etc to that of a third party
run the app on a phone (it's not running in the simulator for me)
click add / hide token
select one that is ERC20, such as Gemini Dollar
the error message consistently appears, it never succeeds

vladyslav-iosdev:
> I just tested this from scratch to be sure and it's there
> 
> To reproduce
> 
> clone master
> change the bundle identifier, groups, etc to that of a third party
> run the app on a phone (it's not running in the simulator for me)
> click add / hide token
> select one that is ERC20, such as Gemini Dollar
> the error message consistently appears, it never succeeds

@benslaney For resolving issue of `Infura calls limitation` you can use your own infura key. From our side we can try to tweak displaying error message to display more informative error message

benslaney:
sure, okay. it is indeed working with my own key. thank you

PR #3055, opened 2021-08-11
Mainnet and Testnets should be mutually exclusive, remove duplicated tokens filtering  #2827

### LINKED ISSUE #2827: Mainnet and Testnets should be mutually exclusive
Noticed this yesterday while working on something else. I don't think the code I was working on caused this: but mainnet and rinkeby native crypto showed up at the same time. Can't reproduce.

vladyslav-iosdev:
Fixes #2827


hboon:
Closed for now, refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/2827#issuecomment-897383077

PR #3063, opened 2021-08-13
Refactor `EventSourceCoordinator` and `EventSourceCoordinatorForActivities` with `.functional` extension, fix crash: converting `Int` from `BigUInt`, fixed with replacing `Int` with `String`  #3011

### LINKED ISSUE #3011: Crash when switching wallet in Settings > Change / Add Wallet

vladyslav-iosdev:
Closes #3011


hboon (changes requested):
Build is broken

hboon (approved):

PR #3065, opened 2021-08-13
Fixes crash of `null pointer` when switch bettwen wallets  #3011

### LINKED ISSUE #3011: Crash when switching wallet in Settings > Change / Add Wallet

vladyslav-iosdev:
Closes #3011


hboon:
Added a TODO

hboon (approved):

PR #3122, opened 2021-09-03
Fix typo in comment

hboon:
@vladyslav-iosdev is that true though? Applies for Heco too?


vladyslav-iosdev:
yep, the same for `heco` too

vladyslav-iosdev (approved):

hboon:
Thanks!

PR #3130, opened 2021-09-07
Update BalanceFetcher with updating tokens using .main queue #3125

### LINKED ISSUE #3125: Reduce background queues count
Fix rapid Realm files grow up, move from background accessing to realm to using main thread

vladyslav-iosdev:
Closes #3125

### DIFF
--- a/AlphaWallet/Core/Coordinators/WalletBalance/PrivateBalanceFetcher.swift
+++ b/AlphaWallet/Core/Coordinators/WalletBalance/PrivateBalanceFetcher.swift
@@ -24,6 +24,7 @@ protocol PrivateBalanceFetcherType: AnyObject {
     func refreshBalance(updatePolicy: PrivateBalanceFetcher.RefreshBalancePolicy, force: Bool)
 }
 
+// swiftlint:disable type_body_length
 class PrivateBalanceFetcher: PrivateBalanceFetcherType {
     typealias TokenIdMetaData = (contract: AlphaWallet.Address, tokenId: BigUInt, json: String, value: BigInt)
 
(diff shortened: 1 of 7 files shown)


vladyslav-iosdev:
Wait until erc1155 will done.

vladyslav-iosdev:
@hboon u can check this pr, i suppose its need to rename it ;)

AlphaWallet/Core/Coordinators/WalletBalance/PrivateBalanceFetcher.swift
```diff
@@ -89,29 +90,32 @@ class PrivateBalanceFetcher: PrivateBalanceFetcherType {
...
-                }
-
-                if let value = balanceValueHasChange, value {
-                    delegate.didUpdate(in: strongSelf)
-                }
-
-                group.leave()
-            }
+            let promise = refreshBalanceAction(forToken: tokenObject)
+            promises.append(promise)
```
hboon, line 111: We can use a `map()`?

hboon (approved):
Looks better. Too many `DispatchGroup`s are harder to track :O

AlphaWallet/Core/Coordinators/WalletBalance/PrivateBalanceFetcher.swift
vladyslav-iosdev: yes, but this removes them at all.

vladyslav-iosdev:
yep, with these changes the app works a little bit slower, but the issues we got with threads should disappear, as well as crushes related to them

vladyslav-iosdev:
but it seems like it stucks on app launch only.

vladyslav-iosdev:
@hboon i am not pretty sure but in `TokensDataStore` we got `updateTokenUnsafe(primary Key` method, that checks whethed value should be changed or not, what do u think, do we need to do this check? it shouldn't brake any logic, its just for safety to avoid notifying that object has changed,  when its value hasn't changed

hboon:
@vladyslav-iosdev I haven’t checked. But sounds like it’ll be fine. So you want to merge this into master? We can run with it and observe like you said

PR #3161, opened 2021-09-08
WIP support for ERC1155 backend

hboon:
This is WIP for ERC1155 backend. It is meant to be merged in master in this state. The code to pull ERC1155 tokens are not executed yet, so it is expected to not break the app.

Want to merge now because the merge conflicts are killing me :)

(I noticed the app is much slower, but not sure if it is because of this PR,  but let's test it out after merging)


hboon:
(And i know this is a big PR, so the usual calls for tests, modularisation applies :))

vladyslav-iosdev (approved):

Issue #2894, opened 2021-06-14
Showing token unit price instead of the total amount changed


zhangzhongnan928:
![image](https://user-images.githubusercontent.com/33795543/121839692-fbf7d180-cd1d-11eb-921a-72f33bb59f01.jpeg)

I find this number 👆 is useless.
And the UI is missing the very important token unit price. @colourfreak

colourfreak:
In the #2273, I explained what this should be. Maybe not clear enough.

I conducted 3 tests using usabilityhub:
1) https://app.usabilityhub.com/tests/86ec06814d57/results/1f78709fe6f1

<img width="423" alt="Screenshot 2021-07-08 at 11 23 29" src="https://user-images.githubusercontent.com/51817359/124897902-03d62900-dfdf-11eb-82d3-5efc3b1ab081.png">

2) https://app.usabilityhub.com/tests/6076f465a67c/results/590f397ceee6

<img width="427" alt="Screenshot 2021-07-08 at 11 23 39" src="https://user-images.githubusercontent.com/51817359/124897929-09337380-dfdf-11eb-9351-ec1e1bd7a566.png">

3) https://app.usabilityhub.com/tests/bcdef3f68e73/results/420f2f243df1

<img width="347" alt="Screenshot 2021-07-08 at 11 23 49" src="https://user-images.githubusercontent.com/51817359/124897981-151f3580-dfdf-11eb-8eaa-851b94b9a4cf.png">

and the result is that they understand if the number you marked is the ‘24H Price change’ of the ‘how much your 
tokens are worth in USD’ (5% x 1500 USD). This is also the same as Delta app, Zerion App, Blockfolio. So the winner was 2)

@zhangzhongnan928 @hboon 
The trigger to open AlphaWallet and check wallet tab should be: **tell me how much money I gained/lost in last 24h**. There are other tools to check the market prices. And if you are looking for that data, open a token and you will find numbers like ETH price. 

So if **1 ETH = 1000 USD** and the ** ETH price raised by 5%** in last 24H then it should like this:

<img width="1665" alt="Screenshot 2021-07-08 at 11 16 29" src="https://user-images.githubusercontent.com/51817359/124896684-f2d8e800-dfdd-11eb-9b34-11034e9ea2de.png">

zhangzhongnan928:
make sense

colourfreak:
[...]

PR #3174, opened 2021-09-09
Improve the total amount change color in the Wallet tab #3165

### LINKED ISSUE #3165: Improve the total amount change color in the Wallet tab
In the #2273, I explained what this should be. Maybe not clear enough.

I conducted 3 tests using usabilityhub:
1) https://app.usabilityhub.com/tests/86ec06814d57/results/1f78709fe6f1

Screenshot 2021-07-08 at 11 23 29

2) https://app.usabilityhub.com/tests/6076f465a67c/results/590f397ceee6

Screenshot 2021-07-08 at 11 23 39

3) https://app.usabilityhub.com/tests/bcdef3f68e73/results/420f2f243df1

Screenshot 2021-07-08 at 11 23 49

and the result is that they understand if the number you marked is the ‘24H Price change’ of the ‘how much your 
tokens are worth in USD’ (5% x 1500 USD). This is also the same as Delta app, Zerion App, Blockfolio. So the winner was 2)

@zhangzhongnan928 @hboon 
The trigger to open AlphaWallet and check wallet tab should be: **tell me how much money I gained/lost in last 24h**. There are other tools to check the market prices. And if you are looking for that data, open a token and you will find numbers like ETH price. 

So if **1 ETH = 1000 USD** and the ** ETH price raised by 5%** in last 24H then it should like this:

Screenshot 2021-07-08 at 11 16 29

vladyslav-iosdev:
Closes #3165


hboon (approved):
Didn't check :)

vladyslav-iosdev:
a few things has left to done

vladyslav-iosdev:
btw i a little bit confised with what should be here?
Screenshot 2021-09-09 at 09 29 56
currently there is `value change` but out of here 
https://github.com/AlphaWallet/alpha-wallet-ios/issues/2894#issue-919990829 there should be market price,

hboon:
Assume

1 ETH = $1000 (from CoinGecko) API
5% 24h change (from CoinGecko) API
Assume 1.5 ETH in wallet

So we display "1.500 ETH"
and "5%"
and hence total fiat value = 1.5 * 1000 =  1500 -> display "$1500.00"
and hence the fiat gain = 1500 * 5% -> display "$75.00"

Does that answer your question?

vladyslav-iosdev:
> Does that answer your question?

thanks for explanation ;)

hboon (approved):

hboon:
@vladyslav-iosdev would you help fix the tests (if it's relevant to this PR) and then merge?

PR #3178, opened 2021-09-09
Make the space between "Wallet Value Number" and the "Portfolio change in last 24h" bigger #3168

### LINKED ISSUE #3168: Make the space between "Wallet Value Number" and the "Portfolio change in last 24h" bigger
Screenshot 2021-09-08 at 07 47 13

These are too close to each other. Make it bigger, 18pt. On the right you have the current build. On the left you have an aimed design. Also, make sure that the red is #ff3b30 if the price change in 24h is down.

vladyslav-iosdev:
Closes #3168


hboon (approved):

hboon:
@vlad would you help resolve the conflict and then merge?

PR #3195, opened 2021-09-17
Add Price Alerts tab [Samoa] #2632

### LINKED ISSUE #2632: Add Price Alerts tab [Samoa]
https://zpl.io/aNN9JDn

Screenshot 2021-03-10 at 15 46 31

vladyslav-iosdev:
Closes  #2632


AlphaWallet/Core/Coordinators/Alerts/AlertDataStoreType.swift
```diff
@@ -0,0 +1,144 @@
...
+            return defaultValue
+        }
+
+        return result
+    }
+
+    func add(alert: Alert) -> Promise<Void> {
+        return Promise { seal in
+            var alerts = self.alerts
+            alerts.append(alert)
```
hboon, line 77: I think just `self.alerts.append(alert)` might work as expected because of how value types (the array) and observer works with value-type properties. i.e.

```
return Promise { seal in
    alerts.append(alert)
    seal.fulfill(())
}
```

Can you verify?

AlphaWallet/Core/Coordinators/Alerts/AlertDataStoreType.swift
```diff
@@ -0,0 +1,144 @@
...
+
+            self.alerts = alerts
+
+            seal.fulfill(())
+        }
+    }
+
+    func update(alert alertToSeach: Alert, update: AlertUpdates) -> Promise<Void> {
```
hboon, line 85: Typo: "alertToSeach"

hboon (commented):
Seems ok otherwise. Would you hide it behind a feature flag? Then we can follow up a PR and experiment if we can get it to use local notifications first?

AlphaWallet/Core/Coordinators/Alerts/AlertDataStoreType.swift
```diff
@@ -0,0 +1,144 @@
...
+
+import UIKit
+import PromiseKit
+
+enum AlertUpdates {
+    case enabled(Bool)
+    case value(value: Double, marketPrice: Double)
+}
+
+protocol AlertDataStoreType: class {
```
hboon, line 16: BTW, do you think it helps to create a protocol when there's only 1 class implementing it? Why?

vladyslav-iosdev: I found it helpfull for testing, it allows us to sapp mock objects  and so on. As well as extract interface from implementation.

AlphaWallet/Core/Coordinators/Alerts/AlertDataStoreType.swift
vladyslav-iosdev: checking

AlphaWallet/Alerts/Coordinators/EditAlertCoordinator.swift
```diff
@@ -0,0 +1,56 @@
...
+        viewController.hidesBottomBarWhenPushed = true
+        viewController.navigationItem.leftBarButtonItem = .backBarButton(self, selector: #selector(backSelected))
+
+        navigationController.pushViewController(viewController, animated: true)
+    }
+
+    @objc private func backSelected(_ sender: UIBarButtonItem) {
+        navigationController.popViewController(animated: true)
+
+        delegate.flatMap { $0.didClose(in: self) }
```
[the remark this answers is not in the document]
vladyslav-iosdev: But without `navigationController.popViewController(animated: true)` view controller will not popped out from navigation stack? Class implemented protocol `EditAlertCoordinatorDelegate` just removes coordinator, and doesn't handle presentation.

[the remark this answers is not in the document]
hboon: No, keep the pop.

I mean, can we replace:

`delegate.flatMap { $0.didClose(in: self) }`

with:

`delegate?.didClose(in: self)`

?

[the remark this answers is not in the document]
hboon: There's a bunch of flatMaps in this PR which seems like they can be replaced by optional chaining `?` like this.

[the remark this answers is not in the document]
vladyslav-iosdev: oh, ok, got it

hboon (commented):
Doesn't build for me:

> Build target AlphaWallet
> Compile Swift source files (x86_64)
> Compile TokenViewController.swift (x86_64)
> (70, 55) Cannot convert value of type '[UIView]' to expected argument type '[PageViewType]'
> (70, 55) Arguments to generic parameter 'Element' ('UIView' and 'PageViewType') are expected to be equal
> Build failed with 1 error and 18 warnings in 2 min, 43 sec

hboon (changes requested):
Looks good, I'm sorry, I wrote these when I looked previously but forgot to post them:

- [ ]  Maybe it's WIP, but the alerts seems to be not tied to particular tokens, eg.

1. Tap on ETH, add an alert for $3000
2. Tap on KICK, the alert for $3000 appears too

- [ ] Could you rename `Alert` to `PriceAlert` and also related types like `EditAlertCoordinator` etc?

vladyslav-iosdev:
> * PriceAlert

ok, np, i will double check.

hboon (approved):
A few more to rename:

* AlertServiceType
* AlertService
* AlertDataStoreType
* AlertType

Typos:

* var decription: String {
* enum AlertsFolterStrategy {

Can we make it such that if price ticker is not available for a token, the alerts tab is hidden for that token?

You can just merge right away when you are ready.

PR #3208, opened 2021-09-23
Add ability to subscribe on queue for Subscribable #3207

### LINKED ISSUE #3207: Add ability to subscribe on queue for `Subscribable`

vladyslav-iosdev:
Closes #3207


hboon (approved):
Ok, but we ought to be careful. Reduce the use of threads/queues remember? :)

PR #3209, opened 2021-09-23
EIP 3085 doesn't work #2966

### LINKED ISSUE #2966: EIP 3085 doesn't work
I was trying out https://chainlist.org/. None of the chains that I tried to add works on AlphaWallet. However, most of them works on Metamask.

image

One random example that worked in Metamask Thai Chain:
Network Name: ThaiChain
Network URL: https://rpc.dome.cloud
Chain ID: 7
Currency Symbol: TCH
Block Explorer URL: https://thaichain.io

vladyslav-iosdev:
Closes #2966

### DIFF
--- a/AlphaWallet/Browser/Coordinators/DappRequestSwitchCustomChainCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/DappRequestSwitchCustomChainCoordinator.swift
@@ -1,6 +1,7 @@
 // Copyright © 2021 Stormbird PTE. LTD.
 
 import UIKit
+import PromiseKit
 
 protocol DappRequestSwitchCustomChainCoordinatorDelegate: AnyObject {
     func notifySuccessful(withCallbackId callbackId: Int, inCoordinator coordinator: DappRequestSwitchCustomChainCoordinator)
@@ -137,6 +138,11 @@ extension DappRequestSwitchCustomChainCoordinator: EnableChainDelegate {
 }
 
 extension DappRequestSwitchCustomChainCoordinator: AddCustomChainDelegate {
+
+    func notifyAddExplorerApiHostnameFailure(customChain: WalletAddEthereumChainObject, chainId: Int) -> Promise<Bool> {
+        UIAlertController.promptToUseUnresolvedExplorerURL(customChain: customChain, chainId: chainId, viewController: viewController)
+    }
+
     //Don't need to notify browser/dapp since we are restarting UI
     func notifyAddCustomChainQueuedSuccessfully(in addCustomChain: AddCustomChain) {
         analyticsCoordinator.log(action: Analytics.Action.addCustomChain, properties: [Analytics.Properties.addCustomChainType.rawValue: "dapp"])
--- a/AlphaWallet/Extensions/UIAlertController.swift
+++ b/AlphaWallet/Extensions/UIAlertController.swift
@@ -91,11 +91,11 @@ extension UIAlertController {
         title: String? = .none,
         message: String? = .none,
         style: UIAlertController.Style,
-        in navigationController: UINavigationController
+        in viewController: UIViewController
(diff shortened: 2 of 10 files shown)


hboon (changes requested):
So take the example of ThaiChain (chainId = 7), it sends this payload:

```
{"id":51530,"object":{"chainType":"ETH","chainId":"0x7","rpcUrls":["https:\/\/rpc.dome.cloud"],"blockExplorerUrls":["https:\/\/thaichain.io"],"chainName":"ThaiChain","nativeCurrency":{"name":"ThaiChain Ether","decimals":18,"symbol":"TCH"}},"name":"walletAddEthereumChain"}
```

(We can log it by looking at `jsonString` in `DappAction.fromMessage(_:)`).

So we can't figure out the blockchain explorer endpoint URL in the payload for the key `blockExplorerUrls`.

If this is the problem, then I think we should just show the user an error that the blockchain explorer URL is invalid, similar to the error we show with, but with a different error message and not allow them to proceed:

```
return Promise(error: AddCustomChainError.others(R.string.localizable.addCustomChainErrorNoBlockchainExplorerUrl()))
```

I think this would be better than adding adding the URL and then failing silently when fetching transaction history. We probably should, in the future, support cases where there is no blockchain explorer better, than we can improve this by just confirming with the user to add it without a blockchain explorer.

What do you think?

PS: Googling suggests it should be https://exp.tch.in.th/home, but I can't form the correct endpoint from that too. We can't fix this manually anyway.

vladyslav-iosdev:
i guess if case when we can't resolve explorer url, then yes, we could show to user warning message, "Explorer URL wasn't resolved. You willn't see transactions." or something like that. And continue without explorer url. anyway as i understand user can edit it later, right?

hboon:
Makes sense. But can you check if we have implemented edits yet? :) I don't think we have.

vladyslav-iosdev:
> Makes sense. But can you check if we have implemented edits yet? :) I don't think we have.
> 
> We can swipe to remove custom chains though.

ah, seems like we can't edit or delate it for now.

hboon:
Yeah. Also to check it is consistent for chains added manually and via EIP3085. After this, we ought to ping Tomek so we set up edits. It's also not clear in the list of chains which chain can be deleted/edited (i.e. not built-into the app).

vladyslav-iosdev:
in current PR it works in the next way: when we are not able to resolve url from `checkBlockchainExplorerApiHostname` method we return instance without url right in 
`
.recover { _ -> Guarantee<(customChain: WalletAddEthereumChainObject, chainId: Int, rpcUrl: String)> in
            return .value((customChain: customChain, chainId: chainId, rpcUrl: rpcUrl))
        }
` 
do i need to update it with throwing an error? for now

vladyslav-iosdev:
> Yeah. Also to check it is consistent for chains added manually and via EIP3085. After this, we ought to ping Tomek so we set up edits. It's also not clear in the list of chains which chain can be deleted/edited (i.e. not built-into the app).

ok, i see.

hboon:
For now, either:

A. throw an error so a message is displayed and user can't add it and
B. show them the message like in (A) and let them proceed anyway

hboon:
And then create an issue so we figure out editing, etc: https://github.com/AlphaWallet/alpha-wallet-ios/pull/3209#issuecomment-926391018

vladyslav-iosdev:
Maybe tweak warning massage is needed, not sure about it;)

hboon (approved):
Looks great. How about this message:

> Blockchain explorer URL can't be resolved. Would you like to continue without it? Transaction history will not be fetched

Feel free to merge.

vladyslav-iosdev:
> Blockchain explorer URL can't be resolved. Would you like to continue without it? Transaction history will not be fetched

awesome! will update with it.

PR #3214, opened 2021-09-24
Research ENS Avatar #3212

### LINKED ISSUE #3212: Research ENS Avatar
https://gist.github.com/Arachnid/9db60bd75277969ee1689c8742b75182

Experiment-only: replacing blockie image with this if available.

If this works out, we can explore showing it at:

A. Replace blockie
B. Display it alongside ENS names (eg. recipient address fields and the my wallet screen)

vladyslav-iosdev:
Closes #3212


hboon (commented):
Yeah, this ENS name should have the data we want: https://app.ens.domains/name/brantly.eth

At "TEXT RECORD" > avatar, it says:

> eip155:1/erc721:0xb7F7F6C52F2e2fdb1963Eab30438024864c313F6/2430

But when I ran the app and add `0x983110309620d911731ac0932219af06091b6744` as a watch wallet, the printed record is an empty string even if I change the record type to `.avatar` like this:

```
ensTextRecordsCoordinator.getENSRecord(for: input, record: .avatar).done { string in
```

Maybe it should be resolved against the ENS name instead of the Ethereum address? Since the text record is for the ENS, not for the Ethereum address

vladyslav-iosdev:
> Maybe it should be resolved against the ENS name instead of the Ethereum address? Since the text record is for the ENS, not for the Ethereum address

is will double check it

vladyslav-iosdev:
@hboon hah u was right ;) it returns values for ens , thanks ;)

hboon:
@vladyslav-iosdev works great. Thanks!

Let's leave this for now and see what Tomek comes up with https://github.com/AlphaWallet/alpha-wallet-ios/issues/3212#issuecomment-928714903

hboon:
@vladyslav-iosdev remember to continue with this, if you haven't, since Tomek has the designs.

vladyslav-iosdev:
@hboon  Bump

hboon (changes requested):
address -> ENS lookup is broken when sending:

Screenshot 2021-10-05 at 5 09 22 PM

hboon:
And I think we should clip (round) the avatars (e.g in that screenshot above)

vladyslav-iosdev:
> And I think we should clip (round) the avatars (e.g in that screenshot above)

hah, i checked it only for ens;) my fail

hboon (changes requested):
In the Send screen, this happens after entering the ENS name and editing it a few times so the avatar is shown/hidden a few times, entering "brantly.eth" again does this:

1. avatar 0x...0 shows up (as expected)
2. the avatar displays immediately leaving a gap, followed by the address.

One more to try: limes.eth -> 0xA7860E99e3ce0752D1ac53b974E309fFf80277C6

hboon (changes requested):
After each new app launch, the first time I enter an ENS name in the send page, the avatar appears as an empty space followed by the address. After I edit and restore the ENS name, the avatar now appears:

1. Launch
2. Send page
3. Enter brantly.eth
4. Resolving...
5. Empty (placeholder) space appears for the avatar followed by the  address
6. Edit and restore brantly.eth
7. Avatar + address appears now

If I then change it to limes.eth, same problem and resolution occurs

vladyslav-iosdev:
@hboon bump

PR #3221, opened 2021-09-27
Finish off first cut of ERC1155 support

hboon:
Probably too big to review. Sorry!

* Read additional data like collection created date and description from OpenSea
* Modified UI to display actual fields/data and not hardcoded ones
* Clean ups

Not in this PR:

* Most of the fields at the token level or token ID/instance level are missing. Will have to figure those out. These and follow ups in #3224
* Some renaming. Boon will follow up on these after this PR and related PRs are merged.


hboon:
Force pushed to make ERC1155 support enabled.

vladyslav-iosdev (approved):

PR #3240, opened 2021-09-29
Fix: crash if RPC URL in EIP3085 is not a valid URL

hboon:
Reproduce:

1. https://chainlist.org in browser tab
2. Look for "Arbitrum One" and add it

Crashes because first RPC URL is `https://mainnet.infura.io/v3/${INFURA_API_KEY}` and because we don't substitute the Infura key [1], is not a valid URL.

[1] We can't support Infura keys because if we don't already support this chain in the app, then it must not have been enabled for our Infura account so it wouldn't work anyway.

### DIFF
--- a/AlphaWallet/Settings/Models/AddCustomChain.swift
+++ b/AlphaWallet/Settings/Models/AddCustomChain.swift
@@ -118,7 +118,8 @@ extension AddCustomChain.functional {
     }
 
     static func checkAndDetectUrls(_ customChain: WalletAddEthereumChainObject, chainId: Int) -> Promise<(customChain: WalletAddEthereumChainObject, chainId: Int, rpcUrl: String)> {
-        guard let rpcUrl = customChain.rpcUrls?.first else {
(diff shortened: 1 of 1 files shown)


hboon:
@vladyslav-iosdev Sadly, #3235 didn't cause your name to appear automatically. Did my name get assigned as the reviewer when you create PRs these 2 days?

vladyslav-iosdev (approved):

vladyslav-iosdev:
> @vladyslav-iosdev Sadly, #3235 didn't cause your name to appear automatically. Did my name get assigned as the reviewer when you create PRs these 2 days?
> 
> There's a little icon that says you are the code owner though. Strange.

when u create a new PR, it tells that code i mine, but u files owner  and vice versa
Screenshot 2021-09-29 at 13 52 57
Screenshot 2021-09-29 at 13 52 48

hboon:
Yeah. Was hoping and thought it would also each other as code reviewer. A few less clicks :)

PR #3249, opened 2021-09-30
Prepare build for iOS 15 #3248

### LINKED ISSUE #3248: Prepare build for iOS 15
Check layout if its still works with iOS 15

vladyslav-iosdev:
Closes  #3248


hboon:
Are these changes needed because you set the minimum deployment target to iOS 15? I have been running the app on iOS 15

vladyslav-iosdev:
> Are these changes needed because you set the minimum deployment target to iOS 15? I have been running the app on iOS 15

no, i just opened the app on simulator with ios 15, (preparing table with changes)

hboon:
Oh. Because you are using Xcode 13. Yeah. Looks like we need it then.

hboon:
Thanks

hboon (commented):
I didn't check in depth, but seems Ok. Does it affect building with Xcode 12.x though? Or should we switch totally to Xcode 13?

vladyslav-iosdev:
i haven't checi it with xCode 12.x but i will.

vladyslav-iosdev:
| After PR | Before PR |
| ------------- | ------------- |
|  |  |
| | |
| ||
| |-|
| ||
|||
|||

vladyslav-iosdev:
@hboon Bump

hboon:
@vladyslav-iosdev Does it affect building with Xcode 12.x though? Or do we have to switch totally to Xcode 13?

vladyslav-iosdev:
@hboon do u know what it could be? i am not able to run the proj with xCode 11.
Screenshot 2021-10-05 at 11 58 55

hboon:
@vladyslav-iosdev try upgrade to a new MailChimp SDK version?

vladyslav-iosdev:
the problem is that i can't build the proj, that uses any spm.

vladyslav-iosdev:
> We can also switch to Xcode 13 to build if it works better for you if it doesn't break things for us.

downloading xCode 12.5, want to check it there first.

vladyslav-iosdev:
@hboon checked with iOS 14.4 works in the same way, seems like everything is ok.

hboon:
As in build with Xcode 12.5 or Xcode 13? and running fine on both 14.4 and 15?

vladyslav-iosdev:
> As in build with Xcode 12.5 or Xcode 13? and running fine on both 14.4 and 15?

my fault, it could be just runned with iOS 14.4 on xCode 13

hboon:
That's ok. So just to be sure since it's hard to go back. We think this is true?

> This runs fine with Xcode 13 + iOS 14.4 and iOS 15.

PR #3261, opened 2021-10-04
Add Arbitrum One RPCServer #3241

vladyslav-iosdev:
Part of #3244, Closes #3241, Closes #3243, Closes #3242, Closes #3245, Closes #3246

(edit: doesn't cover the last item of #3244)

PR #3277, opened 2021-10-06
Chain selection doesn't work properly in Settings > Select Active Networks if user has added custom chain before app officially supports the same chain #3262

### LINKED ISSUE #3262: Chain selection doesn't work properly in Settings > Select Active Networks if user has added custom chain before app officially supports the same chain
Eg.

1. Add Arbitrum One as custom chain
2. Build with Arbitrum One support, i.e with #3261
3. Visit Settings > Select Active Networks

Can't seem to select them. Probably filter off custom chains if there's a matching chainID in the support chains? (`RPCServer.customRpcs` maybe?)

vladyslav-iosdev:
Closes #3262

### DIFF
--- a/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
@@ -376,7 +376,7 @@ final class DappBrowserCoordinator: NSObject, Coordinator {
     }
 
     func isMagicLink(_ url: URL) -> Bool {
-        return RPCServer.allCases.contains { $0.magicLinkHost == url.host }
+        return RPCServer.availableServers.contains { $0.magicLinkHost == url.host }
     }
 
     func `switch`(toServer server: RPCServer, url: URL? = nil) {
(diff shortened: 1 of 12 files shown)


vladyslav-iosdev:
> I didn't explain it properly. Let me try again :)
> 
> 1. Checkout master
> 2. Build and run
> 3. Visit https://chainlist.org in the dapp browser and add Arbitrum One
> 4. Checkout [59dfc04](https://github.com/AlphaWallet/alpha-wallet-ios/commit/59dfc04cb63ecc9e26f00c05a04d588467b91387) (this is the current tip of the PR #3261) #this adds official Arbitrum One support which causes the issue because we did step (3) before this
> 5. git cherry-pick ea7a3493e6ba8ae5da6263c5395bc2ef2412478e (which is the tip of this PR) #this PR is supposed to fix the issue caused by running (3) and (4) in that order
> 6. Run app again and we see the screenshot (also related problem in the Enable/disable network screen)
> 
> Screenshot 2021-10-08 at 10 16 38 AM

thx, yes i was able to reproduce it. Checking.

PR #3282, opened 2021-10-07
iOS 15 Safari extension to intercept ethereum: payment links #3206

### LINKED ISSUE #3206: Explore iOS 15 Safari extension to intercept ethereum: payment links

vladyslav-iosdev:
Closes #3206


hboon:
A simple way to verify if it'not working like we want it to is by removing the custom scheme handler, by changing `ethereum` here:

```
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Ethereum Links</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>ethereum1</string>
			</array>
		</dict>
```

Tapping on the URL in mobile Safari no longer switches to AlphaWallet. We want it to still switch.

vladyslav-iosdev:
@hboon for me it works without removing `ethereum` from `CFBundleURLSchemes`, but anyway.  Want to update it with WebPack to use better way to resolve eip681 url

hboon:
Oh. Maybe it didn’t work for me because it was on the simulator. Let me check again

vladyslav-iosdev:
Screenshot 2021-10-11 at 10 14 03
Screenshot 2021-10-11 at 10 14 26
Screenshot 2021-10-11 at 10 14 31

hboon:
@vladyslav-iosdev can you describe to Tomek what are the images file sizes, that they are grayscale or not etc so he can do it with 1 sweep? Maybe point to https://github.com/AlphaWallet/alpha-wallet-ios/pull/3282#issuecomment-939753920 ?

hboon:
@vladyslav-iosdev do we still need anything from Tomek? Good for review again?

vladyslav-iosdev:
> @vladyslav-iosdev do we still need anything from Tomek? Good for review again?

we are still waiting for images. but i guess we can update it later. once its done, the rest should be ok.

hboon:
@vladyslav-iosdev Ok. Let's get it merged first then. Tests seems to be failing, would you help to fix it and buzz me then let's merge.

PR #3320, opened 2021-10-13
Extract AlphaWallet.Address into a CocoaPod #3268

### LINKED ISSUE #3268: Extract AlphaWallet.Address into a CocoaPod
Let's experiment with this.

* Create a new Cocoapod containing AlphaWallet.Address
* The Cocoapod lives in the same repo, maybe `modules/` (or if you can come up with a better name)?
* Pod will have some 3rd party Cocoapods dependencies
* Pod contains as little else as possible (for example `ThreadSafeDictionary` is needed, let's just include it in the Pod for now)
* We don't have to update the main project (yet)

Let's see how small the pod can be and if there's any gotchas?

---

If this works and we want to proceed, we can update the project after the experiment and start moving other things?

vladyslav-iosdev:
Closes #3268


hboon:
~Might be a bit slow getting to this since we need to make sure master works correctly with iOS 15~

Scratch that :) I did a quick check. Seems ok

@vladyslav-iosdev did you encounter or foresee any problems, especially if we have many of these pods, say 50 (hah!)?

vladyslav-iosdev:
so it is going to be located in the same repo, just in sub directory, right? i don't really know how its going to be resolved, do u? it uses `source` like `https://github.com/vladyslav-iosdev/AlphaWalletAddress.git` with sub directory how it will look like?

hboon:
> so it is going to be located in the same repo, just in sub directory, right

Right.

hboon:
@vladyslav-iosdev regarding `Auth0`, probably pasted into the repo/issue :) But yes, let's merge it in. But tag the current tip before the merge, just in case?

hboon:
Regarding the build problems, I have added a commit based on what I build against. git cloning it locally (separate from my usual working directory), then building works for me. Can you see if it works for you?

PR #3356, opened 2021-10-21
Support SVGs for ENS avatar text records #3342

### LINKED ISSUE #3342: Support SVGs for ENS avatar text records
Test with Victor's ENS

vladyslav-iosdev:
Closes #3342

### DIFF
--- a/AlphaWallet/Accounts/Views/AccountViewCell.swift
+++ b/AlphaWallet/Accounts/Views/AccountViewCell.swift
@@ -6,7 +6,7 @@ class AccountViewCell: UITableViewCell {
     private let addressLabel = UILabel()
     let apprecation24hourLabel = UILabel()
     let balanceLabel = UILabel()
-    private let blockieImageView = BlockieImageView()
+    private let blockieImageView = BlockieImageView(size: .init(width: 40, height: 40))
 
     var viewModel: AccountViewModel?
     var account: Wallet?
@@ -30,9 +30,6 @@ class AccountViewCell: UITableViewCell {
         contentView.addSubview(stackView)
 
         NSLayoutConstraint.activate([
-            blockieImageView.heightAnchor.constraint(equalToConstant: 40),
-            blockieImageView.widthAnchor.constraint(equalToConstant: 40),
-
             stackView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 20),
             stackView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 20),
             stackView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -20),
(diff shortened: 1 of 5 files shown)


vladyslav-iosdev:
> * Chopped off in the list of wallets screen

now it should be ok

vladyslav-iosdev:
need to improve svg image caching, in progress.

hboon:
> > * Chopped off in the list of wallets screen
> 
> now it should be ok

Remember to push :)

hboon:
@vladyslav-iosdev I pushed a commit with some logs, giving the output:

> xxx init with size: (100.0, 100.0)
> xxx set blockie main: true image: Optional(AlphaWallet.BlockiesImage.url(https://storage.opensea.io/files/2ed62dcec7fb8ed6b45eff58c4c104a4.svg))

But the blockie is not shown in the Wallets tab. It only shows up if I tap on another tap (eg. Activity) and then tap on Wallet again, and it appears with an additional print:

> xxx set blockie main: true image: Optional(AlphaWallet.BlockiesImage.url(https://storage.opensea.io/files/2ed62dcec7fb8ed6b45eff58c4c104a4.svg))

It doesn't show up in the send screen (recipient address) too.

Screenshot 2021-11-01 at 4 36 00 PM

Maybe the size is wrong somehow?

vladyslav-iosdev:
> Maybe the size is wrong somehow?
> 


i will double check it

vladyslav-iosdev:
yes i get the same from time to time, i checked logs for this images, and here what it returns 
Screenshot 2021-11-01 at 13 00 52

vladyslav-iosdev:
suppose it specific only to this image

hboon:
Strange indeed. I thought maybe there's a way to hack it since the SVG shows up correctly the 2nd time. So 2 ideas came to mind:

A. Load the HTML wrapping the SVG URL twice. Doesn't work
B. Force a re-render of the `` by dynamically changing the `width` of the `` DOM element. This seems to work.

I pushed a commit with A disabled and B enabled. Can you see if that is acceptable? If so, do you think you can change it to use a document onLoad or something better than a timeout? :)

vladyslav-iosdev:
@hboon for me it works in the same way. i am facing white image background for this image.

hboon:
@vladyslav-iosdev do you mean (B) doesn't work for you too? Try setting the timeout to be longer. Otherwise open the webview in Safari web inspector and tweak the style.width manually to see if it appears after the initial load? That would verify if the hack should work.

vladyslav-iosdev:
yep, b doesn't work for me. checking

vladyslav-iosdev:
seems like (B) works. Removed comments and //hhh.

hboon:
Can you also improve over the setTimeout? Maybe with document onLoad or something better? Best if we can detect the image being loaded, if that’s possible. Otherwise we’ll just leave with it.

Also the NSLog(“xxx :)

PR #3394, opened 2021-10-28
Make WalletConnect signature request be clearer about where it's coming from #3289

### LINKED ISSUE #3289: Make WalletConnect signature request be clearer about where it's coming from
Since users can connect to multiple WalletConnect sessions, including multiple sessions with the same dapp, it looks like we ought to add information to this actionsheet, e.g.

* Be clear this is coming from WalletConnect
* the (optional) dapp icon in #3287
* the network/chain (not editable)
* the site's name
* session ID — unfortunately, if the user connects to the same dapp on the same chain, we can't really distinguish it unless we show the cryptic looking "wc:xxx" string. But seems unlikely a user will be able to use this effectively

This is also important to have good multi-chain support.

Screenshot 2021-10-08 at 10 34 29 PM

vladyslav-iosdev:
Closes #3289

### DIFF
--- a/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
@@ -215,7 +215,7 @@ final class DappBrowserCoordinator: NSObject, Coordinator {
 
     private func signMessage(with type: SignMessageType, account: AlphaWallet.Address, callbackID: Int) {
         firstly {
-            SignMessageCoordinator.promise(analyticsCoordinator: analyticsCoordinator, navigationController: navigationController, keystore: keystore, coordinator: self, signType: type, account: account, source: .dappBrowser)
+            SignMessageCoordinator.promise(analyticsCoordinator: analyticsCoordinator, navigationController: navigationController, keystore: keystore, coordinator: self, signType: type, account: account, source: .dappBrowser, walletConnectSession: nil)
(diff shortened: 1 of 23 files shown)


hboon (commented):
Looks good. A few more:

- [ ] "Confirm Transaction" actionsheet should show Network, Name and Website too if prompted by WalletConnect
- [ ] Signature request in dapp should not show WalletConnect icon (refer to screenshot)
    * Try with https://danfinlay.github.io/js-eth-personal-sign-examples/
    * 
Screenshot 2021-11-01 at 11 07 17 AM

vladyslav-iosdev:
> * https://danfinlay.github.io/js-eth-personal-sign-examples/

the problem here is that we actually not using `WalletConnectCoordinator`, all the calls out of browser and there is no way to even simulate wallet connect session. @hboon any suggestions on it? how it can be reesolved

hboon:
> > * https://danfinlay.github.io/js-eth-personal-sign-examples/
> 
> the problem here is that we actually not using `WalletConnectCoordinator`, all the calls out of browser and there is no way to even simulate wallet connect session. @hboon any suggestions on it? how it can be reesolved

@vladyslav-iosdev I guess you figured it out :)

Looks good. There's still a WIP tag. Is it good to merge?

hboon (approved):

vladyslav-iosdev:
Yep, just resolve conflict is needed

PR #3406, opened 2021-11-02
Empty WalletConnect cell on iOS 15 device right after connecting #3385

### LINKED ISSUE #3385: Empty WalletConnect cell on iOS 15 device right after connecting
Maybe related to: #3341 but this happen right after attempting to connect (and maybe the connection was successful)

vladyslav-iosdev:
Closes #3385


vladyslav-iosdev:
Not quite sure in the implementation, as can't reproduce this issue

hboon (approved):
Ok. Not sure if it will react correctly if the session is connected properly (maybe it can't be)

Let's try it anyway.

PR #3407, opened 2021-11-02
Indicate which chain for each WalletConnect session more prominently by showing an icon for the chain for each session #3403

### LINKED ISSUE #3403: Indicate which chain for each WalletConnect session more prominently by showing an icon for the chain for each session
In the list of WalletConnect sessions screen, we current show the connected chain in brackets, but it's not very readable and may not be visible if the site name is long. This comes in handy when we connect multiple sessions to the same dapp (eg. Swap). Can we improve it?

Current|Design
-|-
Screenshot 2021-11-02 at 3 05 21 PM|

vladyslav-iosdev:
Closes #3403


hboon (changes requested):
The small chain icons are different from the token icons it seems. Do you want to set up a new var + `switch` for these icons? Compare the Polygon (new) chain and (existing) token icon for example.

hboon:
Otherwise working good.

hboon:
Can you also generate a fallback if the chain icon is not available? Maybe like how we programmatically generate them for missing token icons now?

hboon (approved):

PR #3411, opened 2021-11-04
Followup tasks for ERC1155 support #3224

### LINKED ISSUE #3224: Followup tasks for ERC1155 support
There are pending tasks for ERC1155 support after #3221:

- [ ] Classify token IDs in an ERC1155 contract into collections. i.e. 1 contract, N collections
    * James seems to have a solution
- [ ] Can we take decimals > 0 as meaning the token ID is fungible? (But can it be a fractional NFT?)
 Support batch transfers (and Select at top right hand corner of Assets sub-tab needs to be enabled)
- [x] Looks like a bug: `let view = SingleTokenCardAmountSelectionToolbarView(viewModel: .init(availableAmount: 20, selectedAmount: 1))`
    * Affects ERC721 ? Beause it's not used  for ERC1155
- [ ] Missing fields in token collection. Not sure we can get all of them (refer to UI design)
    * Issuer
    * Assets
    * Melts
    * Hodlers/Holders
    * Transfers
- [x] Missing fields in token ID screen (refer to UI design)
    * Issuer
    * Circulating Supply
    * Total Supply
    * Reserve Value
    * Type
    * Transferable
    * Melt Fee
    * Transfer Fee
    * Created (i.e. date)
- [ ] Some token URI JSONs have more metadata (e.g. properties?) eg. https://lostrelics.io/item-metadata/18000000000005e1000000000000000000000000000000000000000000000000
- [ ] Check add custom token from UI handles ERC1155 correctly. Especially balance. Also name/symbol is not part of ERC1155 (e.g. Enjin), and symbol is not applicable
- [ ] Check if relevant, about how tokenURI is interpreted: https://github.com/AlphaWallet/AlchemyNFT/issues/513
- [x] Tap token row in send 1155 should expand to show each tokenId + value
- [x] The big image after tapping on a token (tokenId) should almost fill the screen horizontally
- [x] Show default/fallback token ID icon when they are not available #3223

### LINKED ISSUE #3504: TokensCardCollectionCoordinator instance's delegate not set
https://github.com/AlphaWallet/alpha-wallet-ios/blob/79ae11b8b97cac88f013b54c9e0f9b0b313c04d5/AlphaWallet/Tokens/ViewControllers/Collectibles/Coordinators/TokensCardCollectionCoordinator.swift#L35

Looks like a bug?

vladyslav-iosdev:
Closes #3224, Closes #3504


hboon:
@vladyslav-iosdev could you help give a summary what to look for? :)

vladyslav-iosdev:
sure :)
First of all check out `EnjinProvider` its implemented in same way as `OpenSea`, the key difference that its uses Graphql for retrieving data from Enjin. Here we fetching all tokens from `enjin` for users addressm and compare then with tokens that user already fatched from `OpenSea`.

vladyslav-iosdev:
There are some not good solution for token comparison, trying to figure it out.
after mapping tokens we add then to same json as additional fields, and fetch them from `tokenHolder` values

vladyslav-iosdev:
Currently there are could be issues with authorization, (i am not pretty sure in it). also we are fetching not all fields, there missing info like:
Views:6
Hodlers:3
Transfers:2
Melts:0

AlphaWallet/Core/Enjin/EnjinProvider.swift
```diff
@@ -0,0 +1,448 @@
...
+                }
+            }.then({ balances -> Promise<[AlphaWallet.Address: [GetEnjinTokenQuery.Data.EnjinToken]]> in
+                let ids = (balances[owner] ?? []).compactMap { $0.token?.id }
+                return EnjinProvider.functional.getTokens(graphqlClient: self.graphqlClient, ids: ids, owner: owner)
+            })
+        }
+        return fetch
+    }
+
+    private func fetchPage(forOwner owner: AlphaWallet.Address, offset: Int, sum: [AlphaWallet.Address: [GetEnjinBalancesQuery.Data.EnjinBalance]] = [:], completion: @escaping (Swift.Result<[AlphaWallet.Address: [GetEnjinBalancesQuery.Data.EnjinBalance]], OpenSeaError>) -> Void) {
```
hboon, line 115: Think this is `functional` too, right? (I didn't check every function though)

AlphaWallet/Core/Enjin/EnjinProvider.swift
```diff
@@ -0,0 +1,448 @@
...
+
+        graphqlClient.fetch(query: GetEnjinBalancesQuery(ethAddress: owner.eip55String, page: offset, limit: 50)) { [weak self] response in
+            guard let strongSelf = self else { return }
+
+            switch response {
+            case .failure(let error):
+                completion(.failure(OpenSeaError(localizedDescription: "Error calling Engin API: \(String(describing: error))")))
```
hboon, line 122: Might need a different error type, not `OpenSeaError`?

AlphaWallet/TokenScriptClient/Models/AssetAttributeSyntaxValue.swift
```diff
@@ -273,4 +321,48 @@ extension Dictionary where Key == AttributeId, Value == AssetAttributeSyntaxValu
...
+
+    mutating func setBlockHeight(int: Int?) {
+        self["blockHeight"] = int.flatMap { .init(int: BigInt($0)) }
+    }
+
+    mutating func setMintableSupply(bigInt: BigInt?) {
+        self["mintableSupply"] = bigInt.flatMap { .init(int: $0) }
+    }
+
+    mutating func setTransferable(string: String?) {
```
hboon, line 361: Could you help to move the "setter" functions to below each matching property? Refer to the comment near the top of the extension:

> //"setter" functions are intentionally grouped with their complementary "getter" functions

Easier to find :) But if the arg is optional, we should just use the property instead. The other mutating setter functions were created because they accept a non-optional arg.

AlphaWallet/Core/Enjin/EnjinProvider.swift
```diff
@@ -0,0 +1,448 @@
...
+                    strongSelf.cachePromise(withTokenIdCount: tokenIdCount, forOwner: owner)
+
+                    completion(.success(excludingUefa))
+                }
+            }
+        }
+    }
+
+    private func cachePromise(withTokenIdCount tokenIdCount: Int, forOwner wallet: AlphaWallet.Address) {
+        guard tokenIdCount >= EnjinProvider.numberOfTokenIdsBeforeRateLimitingRequests else { return }
```
hboon, line 157: This just blocks subsequent fetches, but doesn't rate limit, right?

AlphaWallet/TokenScriptClient/Models/AssetAttributeSyntaxValue.swift
```diff
@@ -258,6 +262,50 @@ extension Dictionary where Key == AttributeId, Value == AssetAttributeSyntaxValu
...
+
+    var meltFeeRatio: BigInt? {
+        self["meltFeeRatio"]?.intValue
+    }
+
+    var meltFeeMaxRatio: BigInt? {
+        self["meltFeeMaxRatio"]?.intValue
+    }
+
+    var totalSupply: String? {
```
hboon, line 277: Can you call it `totalSupplyStringValue` instead, if it returns `String?`? Just guarding against name clashes and also when we need to return different types for the same attribute.

AlphaWallet/Tokens/Types/NonFungibleFromJson.swift
```diff
@@ -22,6 +22,17 @@ protocol NonFungibleFromJson: Codable {
     var generationTrait: OpenSeaNonFungibleTrait? { get }
     var collectionCreatedDate: Date? { get }
     var collectionDescription: String? { get }
+    var meltValue: String? { get }
+    var meltFeeRatio: Int? { get }
+    var meltFeeMaxRatio: Int? { get }
+    var totalSupply: String? { get }
```
hboon, line 28: I suppose all the new properties are optional, so it doesn't affect JSON strings that we stored in Realm earlier, right?

AlphaWallet/Core/Enjin/API.swift
```diff
@@ -0,0 +1,798 @@
+// @generated
```
hboon, line 1: Can you write notes for how to generate this file and can we generate it with a different file name? (mention Enjin)

hboon (commented):
I didn't check if we are doing it, but the display for `totalSupply` needs to factor in `decimals`, if it is available (otherwise treated as `decimals = 0`)

Can you see if my understanding is correct:

* This PR fetches additional details about ERC1155 tokens, but only tokens for the Enjin ERC1155 contract? If so, how about other contracts?
* We make a group of calls against the Enjin API (there are multiple calls because of pagination, but logically it's just a single call for each wallet)

hboon:
Switching wallets seems to crash the app.

AlphaWallet/Core/Enjin/API.swift
vladyslav-iosdev: Models generations comes from script `Apollo` running before building proj:
it conrains several commented lines:

```
#NOTE: make sure next lines commented, the GetEnjinToken request has bug in scheme it declares available to mint value as string but returns double instead. Make sure property changed with right type. Or ignore it at all.
#Generate models out fo graphql requests
#"${SCRIPT_PATH}"/run-bundled-codegen.sh codegen:generate --target=swift --includes=./**/*.graphql --localSchemaFile="schema.json" API.swift
#Download graphql scheme 
#"${SCRIPT_PATH}"/run-bundled-codegen.sh schema:download --endpoint="https://cloud.enjin.io/graphql/default" API.swift
```
there some inconsistencies between what api returns and what is returns in scheme, so we got types missmatch during parsing models. that is why we commened these lines

AlphaWallet/Tokens/Types/NonFungibleFromJson.swift
vladyslav-iosdev: not sure. it does, because we update existing json with additional information contained in these fields. (ps if i understood u right)

AlphaWallet/Core/Enjin/EnjinProvider.swift
vladyslav-iosdev: not quite sure, that was already presented in OpenSea, haven't check what is really does

AlphaWallet/Core/Coordinators/WalletBalance/PrivateBalanceFetcher.swift
```diff
@@ -221,31 +226,45 @@ class PrivateBalanceFetcher: PrivateBalanceFetcherType {
...
+            return .value([])
+        }
+
         return firstly {
-            getTokensFromOpenSea()
-        }.then(on: queue, { [weak self] contractToOpenSeaNonFungibles -> Guarantee<[PrivateBalanceFetcher.TokenBatchOperation]> in
+            enjinTokensPromise
+        }.then { enjinTokens in
+            return tokensFromOpenSeaPromise.map { tokensFromOpenSea -> (enjinTokens: EnjinSemiFungibleTokens, tokensFromOpenSea: OpenSeaNonFungiblesToAddress) in
+                return (enjinTokens: enjinTokens, tokensFromOpenSea: tokensFromOpenSea)
```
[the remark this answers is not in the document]
vladyslav-iosdev: unfortunately we can't, as it supposed to be the same types for promises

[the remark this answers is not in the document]
vladyslav-iosdev: sorry seems like we can :)

vladyslav-iosdev:
> * This PR fetches additional details about ERC1155 tokens, but only tokens for the Enjin ERC1155 contract? If so, how about other contracts?

we are fetching all balance values from enjin for users wallet, not sure if this is only for ERC1155, if it supports ERC721 we also should get it.

vladyslav-iosdev:
> * We make a group of calls against the Enjin API (there are multiple calls because of pagination, but logically it's just a single call for each wallet)

yes, we do it in same way as for OpenSea. We can increase amount per page, for reducing calls to API

hboon:
@vladyslav-iosdev this waiting for me?

vladyslav-iosdev:
> @vladyslav-iosdev this waiting for me?

You can recheck it again

hboon:
Just a bit of cosmetic. The text don't seem vertically aligned in these cells, see "Available to Mint" for example. It's below the centerline.

Screenshot 2021-11-22 at 11 15 25 AM

I don't have much to add functionally. If you are comfortable with this can you do a sweep and remove the fields in the above screen that are placeholders (if any).

Are there equivalent data for OpenSea? There's some collection-level metadata. If you like, we can merge this in first before adding OpenSea data

Can you make it possible to tap to copy the "Token ID" field value? Pop up that UI like we do with "Address copied" in the wallet QR code screen

vladyslav-iosdev:
@hboon bump

hboon (approved):
Ok, let's get it merged in first. There's a few TODOs in the PR, check if they need to be resolved before merge?

Issue #3409, opened 2021-11-03
Support editing custom chains


hboon:
We ship with many built-in chains (See Settings > Select Active Networks), but users can add custom chains.

Custom chains can be added by going to the Browser tab and visiting https://chainlist.org. "Expanse Network" is one example you can add (since it's not bundled with the app). Follow through the flow and when you visit Settings > Select Active Networks again, the chain ("Expanse") should be selected.

You can also add a custom chain by tapping the + button in that "Enabled Networks" screen.

Custom chains are shown in the same screen (Settings > Select Active Networks), but they are:

* Currently not indicated as a custom chain, so users can't tell (out of scope here)
* Can be swiped to delete
* Can't be swiped to edit (this is what this issue is about)

In scope
---
* Swipe a custom chain to reveal an Edit and Delete button
* Implement the Edit functionality
* Duplicate/Modify the "Add Custom RPC Network" screen (accessible by tapping +) screen to make it support editing

Out of scope
---
* Icons for Edit and Delete button after swiping custom chain

Because our Infura key (see `Constants+Credentials.swift`) is included in the repo, it's often rate limited. If adding the custom chain doesn't work and you notice `429`s in the console, try a different Infura key.

Refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/3216#issuecomment-953641735 for the swipe menu (but icons aren't necessarily)

eviltofu:
The scope describes setting up the Delete button but not the delete function. Is the delete function to be hooked up to the button?

hboon:
Sorry, I just have edited that part out. Delete is already supported when you swipe; so you only need to implement editing.

eviltofu:
Is there a localised string for the edit button title or should I use `R.string.localizable.editButtonTitle()`? The delete button which was already implemented by someone else has a X using `R.string.localizable.delete()`. It doesn't look consistent.

hboon:
`R.string.localizable.editButtonTitle()` is fine.

eviltofu:
How do I validate a CustomRPC?

eviltofu:
Are duplicate chainName, symbol, rpcEndpoint, explorerEndpoint allowed when editing a CustomRPC?

eviltofu:
What sort of activity need to be logged to the Analytics?

eviltofu:
How do I indicate there are new entries to the localisation strings? Who do I inform so new translations can be added?

hboon:
[...]

PR #3415, opened 2021-11-09
Edit custom rpc

eviltofu:
There appears to be some problem with the Podfile.lock. I appear to be using a newer version:  1.11.2. I started on the master branch before making a new branch and merging the master branch locally. I did not push any changes to the master branch. Hopefully, this won't cause a problem.

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		02D0F18E27379E1300774DDA /* EditCustomRPCCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D0F18D27379E1300774DDA /* EditCustomRPCCoordinator.swift */; };
+		02D0F1902737A14900774DDA /* EditCustomRPCViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D0F18F2737A14900774DDA /* EditCustomRPCViewModel.swift */; };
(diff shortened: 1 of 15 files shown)


hboon (commented):
Not specific to the logic, but logistics for the PR:

> There appears to be some problem with the Podfile.lock. I appear to be using a newer version: 1.11.2.

You can run `bundle` first to install the correct gem versions. Then `pod install`. That should get you the matching versions.

> I started on the master branch before making a new branch and merging the master branch locally.

Best to start a new branch (you'll have to eventually before pushing anyway):

```
git checkout master
git checkout -b <name>
```

> Merge branch 'master' into edit_customRPC

We wouldn't want any commits like this that is not part of GitHub merging a PR (after the PR is done). It breaks the history. Can you help to squash the code for the PR into a single commit unless they are really for different purposes (then it probably belongs to a separate PR). So that when you look at `git log`, it should show these commits:

```
the-one-new-commit-created-for-this-pr
master
1-older-than-master
2-older-than-master
```

This makes it much easier to review and for anyone working on the codebase concurrently to create their PRs. eg. I can't look at the PR and figure out easily what has changed.

eg. `Added entires to other four language files.` should be in the same commit as the actual changes because they are part of the one thing (Add support for editing custom RPC servers)

Give the commit a good 1-liner to explain what is done, in this case:

```
Add support for editing custom RPC servers
```

Then generally if there are details, we can add to it, but it is important what we writee for the first line so we can refer to it.

Can you include this in the first comment for the PR (in GitHub): "Closes \#3409" ? The closes keyword (and a few others) tells GitHub to close the issue automatically when the PR is merged.

PR #3424, opened 2021-11-12
Fix: fiat value displayed for ERC20 tokens in single token view is wrong

hboon:
Fix: fiat value displayed for ERC20 tokens in single token view is wrong

Polygon WETH is using Matic's fiat value:

Before PR|After PR
-|-
Screenshot 2021-11-12 at 3 05 45 PM|Screenshot 2021-11-12 at 3 07 15 PM


hboon:
@vladyslav-iosdev bump. In case you missed this

vladyslav-iosdev (approved):

PR #3425, opened 2021-11-13
WalletConnect v2 #3413

### LINKED ISSUE #3413: WalletConnect v2
* Docs: https://docs.walletconnect.com
* SDK
    * Swift: https://github.com/WalletConnect-Labs/WalletConnectSwiftV2
        * Doc mentions requires minimum deployment iOS 13 (but doesn't seem so in practice?)
    * Kotlin: https://github.com/WalletConnect-Labs/WalletConnectKotlinV2
* Test sites
    Test wallet site
       https://react-wallet.walletconnect.com
    Test dapp site
        https://react-app.walletconnect.com

Observations
---
* Different version number in the wc URL
    * v2 vs v1 example (@2 vs @1)
    * wc:ea6410bc3ea20864f4331095335e0d3a23f5e70820f00f8edae9ec14eeb33917@2?controller=false&publicKey=35af052875ac113c6efab51934c8fa7967990345b6dfeb2a0a8b964693350c32&relay=%7B%22protocol%22%3A%22waku%22%7D and wc:2fa9b5c0-37f9-498f-b975-7874a6cf74b5@1?bridge=https%3A%2F%2Fs.bridge.walletconnect.org&key=6d9e4064d39e574dcdb0218ab31b2d2bb2a60ca15ebf42d0ead9d2a327de396f
* Allow requesting a set of chains
* Wallet no longer controls the chain ID of the chain, but session tracks multiple chain IDs
* Dapp requests permission to use a set of JSON-RPC methods (and whether they are supported)
* Session expiry, default to 7 days
* What are "pairings"? Decouples session from peer to peer pair
* Each session gets (after requesting and granted access) access to multiple wallets/accounts and chains

Wallet
---
* List all active sessions
* User can expand each session to check which account(s) and chain(s)
* Disconnect sessions
* Grant permissions for requested (and supported) JSON-RPC methods dapp needs

> Dapps should request upfront as many chains as possible from the wallet to expose in order to allow the user to control which chain and/or account is using directly from the dapp.
    https://docs.walletconnect.com/protocol/ux-considerations
    Doesn't sound like a good design choice

> Account decoupling
    This more relevant for Wallets where some multi-account wallets have the concept of an "active" account on their interface. This should be completely decoupled from the session state on WalletConnect 2.0. A user should be able to expose and sign methods targetted at an account that is different than the active one in the wallet.
        https://docs.walletconnect.com/protocol/ux-considerations
        We can be limited to 1 account at a time (because we have the concept of an active wallet), for now?

vladyslav-iosdev:
Closes #3413


vladyslav-iosdev:
@hboon is this ok for displaying methods from wallet connect v2 proposal?
IMG_5320
IMG_5321

hboon:
@vladyslav-iosdev yes. As long as it doesn’t affect how v1 is displayed. We can ask Tomek to help with the design once we know all the screens to modify to cater for both versions.

vladyslav-iosdev:
@hboon u can check this pr. it may cause creashes because of WalletConnectV2 isn't perfect yes.

hboon (commented):
Probably like you said because of the protocol or the SDK itself, but crash when the first time I scanned a QR code.

- [ ] Does it need a minimum deployment target of iOS 13.0? Wonder if it will be a problem for compatibility for BC.

- [ ] What happens if we the dapp mixes mainnet and testnets in the chain request?


- [ ] After making a connection by browsing for a QR code from the photo library and connecting successfully, the QR scanner was not closed.

- [ ] Tap on a session and the networks shown are "Kovan, Optimistic Kovan, Ropsten". But should be "Kovan, Optimistic Kovan". Ropsten is just enabled in AlphaWallet, but not involved in the connection.

Screenshot 2021-11-18 at 1 09 19 PM

- [ ] Will the "Switch Network" still do anything for WalletConnect v2? If not, let's remember to note down for Tomek and us to remove it.

- [ ] Can we handle (and reject it) when there are multiple wallets and the dapp requests for it in a single session?

- [ ] How does pairings work? How do we handle it?

- [ ] WalletConnect v1 might be broken. I can't connect to https://example.walletconnect.org

- [ ] Do we handle if the dapp requests for more JSON-RPC request permissions after the connection has been made?

vladyslav-iosdev:
> * [ ]  Does it need a minimum deployment target of iOS 13.0? Wonder if it will be a problem for compatibility for BC.
> *

the library sets minimum deployment target of iOS 13.0 as its using combine features.

vladyslav-iosdev:
> * [ ]  What happens if we the dapp mixes mainnet and testnets in the chain request?
> *

missed to handle this case, but i will update it in way to accept only mainets or only testnets

vladyslav-iosdev:
> * [ ]  Tap on a session and the networks shown are "Kovan, Optimistic Kovan, Ropsten". But should be "Kovan, Optimistic Kovan". Ropsten is just enabled in AlphaWallet, but not involved in the connection.

yep, there is one issue for now, is that we actually don't know that servers are connected to. as Session doesn't return this info. (working on it)

vladyslav-iosdev:
> * [ ]  WalletConnect v1 might be broken. I can't connect to https://example.walletconnect.org
> *

have to recheck, it was working for me fine.

vladyslav-iosdev:
> * [ ]  Do we handle if the dapp requests for more JSON-RPC request permissions after the connection has been made?

yep, we updating session when receiving delegated callback 
`func didUpgrade(sessionTopic: String, permissions: SessionType.Permissions)`

vladyslav-iosdev:
> * Will the "Switch Network" still do anything for WalletConnect v2? If not, let's remember to note down for Tomek and us to remove it.

updated code base with ability of switching rpc servers for v2

vladyslav-iosdev:
@hboon this pr seems more stable

hboon:
@vladyslav-iosdev are you able to build this on *cough* M1? :)

vladyslav-iosdev:
> @vladyslav-iosdev are you able to build this on _cough_ M1? :)

nope, the spm stucks on validation

vladyslav-iosdev:
For me it doesn't work due to this issue

eviltofu:
> For me it doesn't work due to this issue https://developer.apple.com/forums/thread/696504

@vladyslav-iosdev If you download the Xcode from developer resources, it might work. https://developer.apple.com/download/release/

vladyslav-iosdev:
> If you download the Xcode from developer resources, it might work.

yep, it works for me now.

vladyslav-iosdev:
> are you able to build this on _cough_ M1? :)

yep, it works for me

vladyslav-iosdev:
@hboon its ready for review

PR #3427, opened 2021-11-13
Fix price for ERC20 tokens matching same CoinGecko ID not appearing correctly

hboon:
Includes commit from #3424.

Only 1 of the tokens will show the price.

E.g WETH on mainnet and Polygon are based on CoinGecko's "weth", only one of them (random pick) will show the price

Before PR|After PR
-|-
Screenshot 2021-11-14 at 2 13 03 AM|Screenshot 2021-11-14 at 1 53 59 AM


vladyslav-iosdev (approved):

hboon:

PR #3454, opened 2021-11-22
Rearranged Run SwiftLint phase before Compile sources phase

### LINKED ISSUE #3451: Fix SwiftLint results not appearing in build results

eviltofu:
Closes #3451

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -4861,13 +4861,13 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 2912CD1F1F6A830700C6CBE3 /* Build configuration list for PBXNativeTarget "AlphaWallet" */;
 			buildPhases = (
+				29282B511F7629DB0067F88D /* Run SwiftLint */,
 				7C3903F7FBE1023CAA5AED0C /* [CP] Check Pods Manifest.lock */,
 				C83F787024A1C1DB00783B67 /* Check correct credentials used for API keys */,
 				29336FE81F6B24FF005E3BFC /* R.Swift */,
 				2912CCF11F6A830700C6CBE3 /* Sources */,
 				2912CCF21F6A830700C6CBE3 /* Frameworks */,
 				2912CCF31F6A830700C6CBE3 /* Resources */,
-				29282B511F7629DB0067F88D /* Run SwiftLint */,
 				8A3C752ACC5CACBD729A8F25 /* [CP] Embed Pods Frameworks */,
 				F6006D07F1DCADBC1150AD02 /* [CP] Copy Pods Resources */,
 				878EE8F6255ADA49000210DE /* Embed App Extensions */,


hboon (commented):
Does this work properly if:

- [x] it's a clean build
- [ ] if an incremental build is triggered from a change that affects R.swift
- [ ] if an incremental build is triggered from a change that doesn't affects R.swift
- [ ] Podfile changes?
- [ ] Podfile.lock changes?

hboon:
Another thought: Is it actually better to run SwiftLint before compilation? Seems more efficient to catch compilation errors first and stop the build (if it does) rather than wait to solve mostly stylistic issues?

eviltofu:
> * [ ]  it's a clean build

Yes. I cleaned the Xcode project and modified the .swiftlint.yml file to show warnings with files greater than 100 lines. Building the project showed the warnings.

> * [ ]  if an incremental build is triggered from a change that affects R.swift

The R.generated.swift file is excluded from SwiftLint. Any changes that trigger a change in the R.generated.swift file will only affect the swift files that use the strings in the source code when compiling. I added an entry `"appTestingMonkey5" = "2";` in all the strings file and a line `print(R.string.localizable.testErrorAppMonkeys5)` in one of the source files and compiled. There was no reported build error but I did see an intermittent error where `R.string.localizable.testErrorAppMonkeys5` is marked as non-existent after a successful build. This has happened before in my previous pull request and restarting Xcode fixed it. 

> * [ ]  if an incremental build is triggered from a change that doesn't affects R.swift

Yes, I added a new line of code and recompiled. The warnings remained consistent.

> * [ ]  Podfile changes?
> * [ ]  Podfile.lock changes?

The Podfile, Podfile.lock, and Pod directory are excluded from SwiftLint. I added a new pod and recompiled. The warnings remain consistent. 

> Another thought: Is it actually better to run SwiftLint before compilation? Seems more efficient to catch compilation errors first and stop the build (if it does) rather than wait to solve mostly stylistic issues?

It should be before because if the compile works and then you have to change the code to satisfy a stylistic requirement, you will have to compile the file again. 

I also noticed that SwiftLint is not applied to the targets `AlphaWalletShare` and `AlphaWalletSafariExtension`.

hboon:
> It should be before because if the compile works and then you have to change the code to satisfy a stylistic requirement, you will have to compile the file again.

Yes. What I mean was, it seems more common to build and get a compilation error rather than a SwiftLint warning as we code. If SwiftLint goes after compilation and there's a compilation error, SwiftLint will (might?) be skipped, so we can fix the compilation error and then repeatedly re-compile until we finally get to fixing any SwiftLint warnings.

If SwiftLint goes first, we'll always run it even if we end up with a compilation error.

Either way, we'll end up with fixing SwiftLint warnings at the end (unless we choose to skip it in the PR and do a follow up for various reasons). I don't how big is the difference in build time when we skip SwiftLint, but it seems like doing it first doesn't have any advantage. What do you think?


> The Podfile, Podfile.lock, and Pod directory are excluded from SwiftLint. I added a new pod and recompiled. The warnings remain consistent.
> The R.generated.swift file is excluded from SwiftLint.

SwiftLint don't run on those, but they do run on code that use those, eg. importing and using new code pods, calling `R.string.localizable.xxx()` which might be missing etc, so I thought they could be affected. But maybe it just goes back to the point of build speeds above?

> I also noticed that SwiftLint is not applied to the targets AlphaWalletShare and AlphaWalletSafariExtension.

Good catch. See if you can fix it separately .

eviltofu:
If it boils down to speed then just do what you think is best. I don’t know enough about the Xcode compilation process to guess which is the most optimal way. The only reason I raised the point was that it was recommended in the SwiftLint home page. I will file an issue regarding the two other targets (unless you want them as two issues?).

hboon:
> If it boils down to speed then just do what you think is best. I don’t know enough about the Xcode compilation process to guess which is the most optimal way. The only reason I raised the point was that it was recommended in the SwiftLint home page.

Ok. Let's close this PR then. No change is the  best PR :)

> I will file an issue regarding the two other targets (unless you want them as two issues?).

One sounds fine unless there are problems. Thanks.

PR #3461, opened 2021-11-22
Added Run SwiftLint phase to AlphaWalletShare / AlphaWalletSafariExtensions

### LINKED ISSUE #3459: Add SwiftLint phase in Build Phase for Targets AlphaWalletShare and AlphaWalletSafariExtension
Currently, in both targets, the SwiftLint phase is not installed.

eviltofu:
Closes #3459.

I am not able to test this other than by compiling the project under the AlphaWalletShare and AlphaWalletSafariExtension schemes. When I run Project > Test under the AlphaWallet scheme, I get `Cannot test target “AlphaWalletUITests” on “iPhone 12”: UITargetAppPath should be provided`. I am able to compile and run the AlphaWallet target on the simulator. 

In Xcode:
1) Select AlphaWalletShare scheme.
2) Build the extension.

**Expects**
3) Successful build.

In Xcode:
1) Select AlphaWalletSafariExtension scheme.
2) Build the extension.

**Expects**
3) Successful build.

In Xcode:
1) Select the AlphaWallet scheme.
2) Run the app.

**Expects**
3) Successful build.
4) App runs on simulator/iPhone.

The SwiftLint command is invoked via `${SRCROOT}/Pods/SwiftLint/swiftlint` instead of using `${PODS_ROOT}` because `${PODS_ROOT}` does not appear to be defined for the extensions. Adding the extensions as targets to the Podfile results in a `linking against a dylib which is not safe for use in application extensions:` warning.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,9 +24,13 @@ opt_in_rules:
 included:
   - AlphaWallet
   - AlphaWalletTests
+  - AlphaWalletSafariExtension
+  - AlphaWalletShare
+  - modules
 excluded:
   - AlphaWallet/Vendors
   - AlphaWallet/R.generated.swift
+  - AlphaWalletSafariExtension/Resources
 function_parameter_count:
   warning: 20
   error: 25


hboon (changes requested):
But it seems like this configures SwiftLint to run on the same set of source files (used by the `AlphaWallet` target) 3 times, with each target triggering it once. i.e. we want each target to trigger a run on their individual files instead.

Alternatively, is it possible to configure SwiftLint on the main target (`AlphaWallet`) to run on all the files instead of adding to the Build Phase for each target?

Also note that we need to include `modules/` which is built as Cocoapods. This is probably not included prior to the PR. We only moved it out from the main target recently.

eviltofu:
> Alternatively, is it possible to configure SwiftLint on the main target (`AlphaWallet`) to run on all the files instead of adding to the Build Phase for each target?
> 
> Also note that we need to include `modules/` which is built as Cocoapods. This is probably not included prior to the PR. We only moved it out from the main target recently.

Yes, I will modify the .swiftlint.yml file to include these three directories.

I tested this by adding additional spaces before { in the code of swift files to trigger the `Opening Brace Spacing Violation`. This showed up in Xcode for AlphaWalletShare but not for AlphaWalletSafariExtension. Running SwiftLint from the command line does reveal these two files.

I also tested this by copying Config.swift into all the sub directories of modules, AlphaWalletShare, and AlphaWalletSafariExtensions and running SwiftLint from the command line. All the files with warning did show up in the results.

hboon:
Ok. Remember to buzz the reviewer you want to target in the PR comments after you push something. Because it might be a WIP push.

eviltofu:
@hboon pull request is ready for review. There was no way to have the warnings for AlphaWalletSafariExtension show up in Xcode warnings. The warnings can be found running the command line check or opening the file in Xcode.

hboon:
> There was no way to have the warnings for AlphaWalletSafariExtension show up in Xcode warnings. The warnings can be found running the command line check or opening the file in Xcode.

But they do show up when building the `AlphaWallet` target:

Screenshot 2021-11-26 at 10 53 38 AM

After I intentionally introduced spurious newlines and build the `AlphaWallet` target, the first "Vertical Whitespace Violation" is `AppDelegate.swift` and the second "Vertical Whitespace Violation" is for `ShareContextHandler.swift` which are in different targets.

So it's working fine.

hboon (approved):

PR #3465, opened 2021-11-23
Refactored app to use one set of classes for adding & editing CustomRPC

### LINKED ISSUE #3452: Refactor so `AddRPCServerViewController.validateInputs()` is shared for both adding and editing custom chains

eviltofu:
Closes #3452.

When integrating the UIAlertController extension `promptToUseUnresolvedExplorerURL` into the UIAlertController extension file, there was a clash between `Result` of the `PromiseKit` module and the `Result` of the `Result` module. The solution was to only import the `Promise` class of `PromiseKit` instead of the entire module. This is accomplished by `import class PromiseKit.Promise`. 

Xcode did not turn up any warning regarding this statement. 

[Import Declaration](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#grammar_import-declaration)

# Testing by adding, then deleting.

## Pre-test

1. on the main screen, tap the "Settings" tab option followed by the "Select Active Networks" option.
2. delete "Expanse Network" if it exists.

## Add test

1. while on the "Enabled Networks" screen, tap the + icon on the upper right corner.

### Expects

2. the "add custom rpc network" screen to appear.

## Enter network data

3. type "Expanse Network" into the Network Name TextField.
4. type "https://node.expanse.tech" into the RPC URL TextField.
5. type "2" into the Chain ID TextField.
6. type "EXP" into the Symbol TextField.
7. type "https://expanse.tech" into the Block Explorer URL TextField.
8. tap the "Save Network".

### Expects

9. alert with warning appears.

10a. tap "Continue".
10b. tap "Cancel".

### Expects 

11a.  alert to disappear and main screen to appear.
11b. alert to disappear.

## Edit test

1. on the main screen, tap the "Settings" tab option followed by the "Select Active Networks" option.
2. locate the "Expanse Network" entry. 
3. slide the "Expanse Network" row towards the left.

### Expects 

4. the delete and edit options appear on the right side of the row.

5. tap the edit option.

### Expects

6. the "edit custom RPC network" screen to appear. It should look exactly like the "add custom roc network" screen except with a different title.
7. "Expanse Network" shown in the Network Name TextField.
8. "https://node.expanse.tech" shown in the RPC URL TextField.
9. "2" shown in the Chain ID TextField.
10. "EXP" shown in the Symbol TextField.
11. "https://expanse.tech" shown in the Block Explorer URL TextField.
12.  edit one of the fields.
13. tap "Save Network" button.

### At least one field is invalid

14a. Textfield with error to be highlighted by red rectangle with reason for error.
15a. Correct the field and tap "Save Network" button again.
16a. go to 13.

### All fields are valid
14b. alert with warning appears.
15b. tap continue.
16b. alert to disappear and main screen to appear.


hboon (changes requested):
Looks great.

I added a commit that fixes analytics. Can you verify and then squash it + force push if it's OK?

eviltofu:
@hboon squashed. Verified that unit tests pass and it's able to add/edit.

hboon:
After we squash several commits into one, it'll be good to clean up the commit message too rather than combine them. So in this case,

I modified so it goes from:

> Refactored app to use one set of classes for adding & editing CustomRPCs.
>     
> Added custom navigation title code
> 
> Fix: log analytics correctly for when user adds custom RPC server manually

to: 

> Refactor: use one set of classes for adding & editing CustomRPCs.

Also good to rebase master before pushing or force pushing for a PR:

```
git fetch upstream master:master #update local master so it's the same as upstream's. Do this when master is not currently checked out
git rebase master #modifies current branch so that all the new commits are built-upon (rebased) on master
```

I did these and force pushed.

hboon (approved):

hboon:
Forgot to modify the commit message like I said I would. Force pushed again.

PR #3466, opened 2021-11-24
Log analytics user with ens avatar

hboon:
Have to be careful not to do it for watched wallets as well as places where an address is used but not owned by the wallet, eg. in the send screen.


vladyslav-iosdev (approved):

hboon:
Aargh. I broke the test suite. Will fix.

hboon:
Force pushed after fixing test and rebasing master

PR #3469, opened 2021-11-24
Updated Travis.yml and corrected code to allow tests to pass

### LINKED ISSUE #3456: Make tests pass on iOS 15 simulator
Seems to be `QRCodeValueParserTests.swift` (strange!) and maybe `EtherKeystoreTests.swift`.

Follow up:

- [ ] Make travis use iOS 15 instead of 14.x

eviltofu:
Closes #3456

The error were apparently caused by `LockEnterPasscodeCoordinator` not calling the `.stop()` function before terminating the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: swift
-osx_image: xcode12.5
+osx_image: xcode13.1
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


hboon (changes requested):
aThe strange thing is I just tried running the test suite with `master` on iPhone 11 + iOS 15 simulator it passes now.

So I can't verify if this helps, but let's merge it since the tests you fixed were faulty.

Can you modify the commit to mention it's to get it to pass on iOS 15 simulator?

Wait till Travis completes before you force push, just in case.

eviltofu:
@hboon currently the PR makes Travis check against 14.5 and 15.0 on an iPhone 12. I believe Apple is going to support 14.x for some time in the future instead of removing support for it in 13.x when 14.x was released.

hboon:
> currently the PR makes Travis check against 14.5 and 15.0 on an iPhone 12. I believe Apple is going to support 14.x for some time in the future instead of removing support for it in 13.x when 14.x was released.

Ok. It adds 1-2 minutes more to the Travis run. But it's not so bad. Let's go with it and see.

Is this ready to review + merge? You should comment in the PR that it's ready for review (and if it should only be reviewed and not merged). Otherwise we'll wonder if it's WIP.

eviltofu:
@hboon This pull request is ready for review.

hboon (approved):

PR #3474, opened 2021-11-25
Make visible textfields, hidden by keyboard, when becoming first responder

### LINKED ISSUE #3453: Next button (keyboard) should scroll screen so textfield is visible for EditCustomRPCViewController
Probably the add equivalent too.

eviltofu:
Closes #3453

## Summary
If a textfield in the edit custom rpc screen becomes first responder and is hidden, it will be scrolled into view.

### Prerequisites 
Requires a custom rpc to be added. Refer to "enter network data" section of this [pull request](https://github.com/AlphaWallet/alpha-wallet-ios/pull/3465) for instructions on how to add one.
### Instructions
1. When on the main screen of app, tap the settings tab bar icon.
2. Tap select active networks option on the Settings screen.
3. Slide the custom rpc (normally Expanse Network) to the left to expose the edit button.
4. Tap the edit button and the edit custom rpc network screen should appear.
5. Tap on the Symbol text field. A keyboard will appear at the bottom of the screen and hide the Block Explorer URL text field.
6. Tap the next button on the keyboard.
### Expects
7. The Block Explorer URL text field should slide up from the bottom.
### Tested 
Tested on
1. iPhone 12, iOS 14.5 (simulator)
2. iPhone 13, iOS 15.0 (simulator)
3. iPhone X, iOS 15.0 (hardware)
4. iPad Pro 9 inch, iOS 15.0 (hardware)

### DIFF
--- a/AlphaWallet/Settings/ViewControllers/SaveCustomRpcViewController.swift
+++ b/AlphaWallet/Settings/ViewControllers/SaveCustomRpcViewController.swift
@@ -140,6 +140,12 @@ extension SaveCustomRpcViewController: TextFieldDelegate {
         return true
     }
 
+    func didBeginEditing(in textField: TextField) {
+        DispatchQueue.main.async {
+            self.editView.unobscure(textField: textField)
+        }
+    }
+
     func doneButtonTapped(for textField: TextField) {
         //no-op
     }
--- a/AlphaWallet/Settings/Views/SaveCustomRpcView.swift
+++ b/AlphaWallet/Settings/Views/SaveCustomRpcView.swift
@@ -133,6 +133,8 @@ class SaveCustomRpcView: UIView {
 
             footerBar.leadingAnchor.constraint(equalTo: self.leadingAnchor),
             footerBar.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+            footerBar.leadingAnchor.constraint(equalTo: scrollView.leadingAnchor),
+            footerBar.trailingAnchor.constraint(equalTo: scrollView.trailingAnchor),
             footerBar.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor),
         ] + roundedBackground.createConstraintsWithContainer(view: self))
     }
@@ -152,6 +154,15 @@ class SaveCustomRpcView: UIView {
         for field in allTextFields {
             field.status = .none
         }
+    }
+
+    func unobscure(textField: TextField) {
+        guard textField != allTextFields.first else {
+            self.scrollView.setContentOffset(.zero, animated: true)
+            return
+        }
+        let rect = self.scrollView.convert(textField.bounds, from: textField)
+        self.scrollView.scrollRectToVisible(rect, animated: true)
     }
 
     private func configureInputAccessoryView() {
@@ -205,6 +216,7 @@ fileprivate func defaultTextField(_ type: UIKeyboardType, placeHolder: String, l
     textField.keyboardType = type
     textField.textField.autocorrectionType = .no
     textField.textField.autocapitalizationType = .none
+    textField.textField.spellCheckingType = .no
     textField.returnKeyType = .next
     textField.placeholder = placeHolder
     textField.label.text = label
--- a/AlphaWallet/Tokens/Views/TextField.swift
+++ b/AlphaWallet/Tokens/Views/TextField.swift
@@ -7,11 +7,15 @@ protocol TextFieldDelegate: AnyObject {
     func doneButtonTapped(for textField: TextField)
     func nextButtonTapped(for textField: TextField)
     func shouldChangeCharacters(inRange range: NSRange, replacementString string: String, for textField: TextField) -> Bool
+    func didBeginEditing(in textField: TextField)
 }
 
 extension TextFieldDelegate {
     func shouldChangeCharacters(inRange range: NSRange, replacementString string: String, for textField: TextField) -> Bool {
         return true
+    }
+    func didBeginEditing(in textField: TextField) {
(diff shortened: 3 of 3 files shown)


AlphaWallet/Settings/Views/SaveCustomRpcView.swift
```diff
@@ -137,11 +141,20 @@ class SaveCustomRpcView: UIView {
...
-        let fields: [TextField] = [chainNameTextField, rpcEndPointTextField, chainIDTextField, symbolTextField, explorerEndpointTextField]
-        for field in fields {
+        for field in allTextFields {
             field.status = .none
         }
     }
+
+    func focusOnTextField(textField: TextField) {
+        let offset = textField.frame.maxY - scrollView.bounds.height
+        guard
```
hboon, line 151: Can you make this 1 (long) line? With the idea to let editors soft-wrap so it's still readable and we don't have to fiddle with indentation.

AlphaWallet/Tokens/Views/TextField.swift
```diff
@@ -7,12 +7,16 @@ protocol TextFieldDelegate: AnyObject {
...
     func shouldChangeCharacters(inRange range: NSRange, replacementString string: String, for textField: TextField) -> Bool
+    func didBeginEditing(in textField: TextField)
 }
 
 extension TextFieldDelegate {
     func shouldChangeCharacters(inRange range: NSRange, replacementString string: String, for textField: TextField) -> Bool {
         return true
     }
+    func didBeginEditing(in textField: TextField) {
+        return
```
hboon, line 18: This default implementation for the protocol function is so that we don't edit other files right? Could you help to add a TODO to this function or better still a follow up TODO in this PR's comment so we'll remember to create an issue to follow up to remove it? eg.

- [ ] Do xxx

hboon (changes requested):
If I tap Next here:

Screenshot 2021-11-25 at 11 39 58 AM

It becomes like this:

Screenshot 2021-11-25 at 11 39 38 AM

Do you think it'll be better if we use `scrollRectToVisible(_:animated:)` instead?

eviltofu:
> Do you think it'll be better if we use `scrollRectToVisible(_:animated:)` instead?

```
    let rect = self.scrollView.convert(textField.bounds, from: textField)
    self.scrollView.scrollRectToVisible(rect, animated: true)
```
does not scroll the scrollview at all. What am I doing wrong?

vladyslav-iosdev:
@hboon maybe usage of `pod 'IQKeyboardManagerSwift'` can help to resolve issue?

eviltofu:
> @hboon maybe usage of `pod 'IQKeyboardManagerSwift'` can help to resolve issue?

I fixed it by manually calculating the offset.

AlphaWallet/Tokens/Views/TextField.swift
eviltofu: This default implementation is to avoid editing other files. I don't understand what you want. What are we removing?

eviltofu:
@hboon pending check and merge

hboon:
> @hboon maybe usage of pod 'IQKeyboardManagerSwift' can help to resolve issue?

```
IQKeyboardManager.shared.enable = true
```

Wow. It looks amazing. But a bit wary of global solutions. Let's see if we can work it out for this one first. Less third-party dependency would be good if we can help it.

hboon (commented):
@eviltofu Seems like the problem is `UIScrollView.contentSize.width` is `0`. I added a commit to show how it might work if it isn't. See if you want to work on that more along that lines, otherwise let me know and we'll merge what you already pushed.

hboon:
And do mention the screen affected. Since there's just 1.

AlphaWallet/Settings/ViewControllers/SaveCustomRpcViewController.swift
```diff
@@ -120,6 +120,7 @@ class SaveCustomRpcViewController: UIViewController {
             }
         }
         editView.allTextFields.first(where: { !$0.statusLabel.text.isEmpty })?.becomeFirstResponder()
+        editView.setNeedsLayout()
```
eviltofu, line 123: This is needed because when a `TextField` becomes invalid, the layout is changed but `layoutSubViews()` for the `editView` is not called. The `contentSize.width` of the `scrollView` is set to zero again. This forces the `contentSize.width` to be reset to non-zero value.

AlphaWallet/Tokens/Views/TextField.swift
hboon: A reminder to remove the default implementation

eviltofu: What if I made this an optional method?

eviltofu:
Fixed scrollView.contentSize.width == 0 issue. Removed code that resets the contentSize.

hboon:
@eviltofu this ready?

eviltofu:
> @eviltofu this ready?

Yes.

AlphaWallet/Settings/Views/SaveCustomRpcView.swift
```diff
@@ -154,6 +156,15 @@ class SaveCustomRpcView: UIView {
         }
     }
 
+    func unobscure(textField: TextField) {
+        guard textField != allTextFields.first else {
+            self.scrollView.setContentOffset(.zero, animated: true)
```
hboon, line 161: Will be good to not include `self.` references unless they are necessary. It's a good indicator that it might be a strong reference in a closure.

hboon (approved):

Issue #3482, opened 2021-11-26
Unexpected and very long error message when saving custom rpc


eviltofu:
When the RPC URL field in a CustomRPC is a valid URL but does not point to any actual location, the following error message is given. An example of such a url is https://node.expanse.tect.

![error](https://user-images.githubusercontent.com/1050309/143564712-c9bd214e-dc2c-4aa3-a026-8baeca858b28.png)

A better error message should be used.

hboon:
Seems like it should appear as how errors look like when filing in the form

<img width="300" alt="Screenshot 2021-11-30 at 10 56 47 AM" src="https://user-images.githubusercontent.com/56189/143977736-1ceab5cc-8cc1-4ed2-801a-f8c2c9818954.png">

PR #3490, opened 2021-11-30
Added code to indicate rpc url error

eviltofu:
Fixes: #3482

## Pre testing
1. Delete the "expanse network" if it exists.
2. On the main screen, tap "settings", then "select active networks".

## Testing
1. Tap the + button in the upper right corner.
2. **Expects** to see "Add Custom RPC Network" screen.
3. Type "Expanse network" into the Network Name text field.
4. Type "https://node.expanse.tect" into the RPC URL text field.
5. Type "2" into the Chain ID text field.
6. Type "EXP" into the Symbol text field.
7. Type "https://expanse.tech" into the Block Explorer URL text field.
8. Tap the "Save network" button.
9. **Expects** RPC URL text field to be highlighted in red with "RPC URL is invalid" under the text field.

### DIFF
--- a/AlphaWallet/Settings/Models/AddCustomChain.swift
+++ b/AlphaWallet/Settings/Models/AddCustomChain.swift
@@ -27,6 +27,13 @@ protocol AddCustomChainDelegate: AnyObject {
     func notifyAddCustomChainQueuedSuccessfully(in addCustomChain: AddCustomChain)
     func notifyAddCustomChainFailed(error: AddCustomChainError, in addCustomChain: AddCustomChain)
     func notifyAddExplorerApiHostnameFailure(customChain: WalletAddEthereumChainObject, chainId: Int) -> Promise<Bool>
+    func notifyRpcURlHostnameFailure()
+}
+
+// TODO: Remove when other classes which implement AddCustomChainDelegate protocol add this function.
+extension AddCustomChainDelegate {
+    func notifyRpcURlHostnameFailure() {
+    }
 }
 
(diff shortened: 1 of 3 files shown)


eviltofu:
Would it be better to have the View Controller register for notifications and the AddChain object send out notifications when the RPC URL is faulty? Currently routing the message from the AddChain object to the delegate to the coordinator to the view controller.

hboon:
As in using notifications to replace delegates? No. Harder to follow. We use subscribers for 1:N value changes. But really hard to manage.

AlphaWallet/Settings/Models/AddCustomChain.swift
```diff
@@ -65,11 +72,14 @@ class AddCustomChain {
...
-            } else {
+            switch $0 {
+            case SessionTaskError.connectionError:
+                self.delegate?.notifyRpcURlHostnameFailure()
+            case is AddCustomChainError:
+                self.delegate?.notifyAddCustomChainFailed(error: ($0 as! AddCustomChainError), in: self)
+            case PMKError.cancelled:
+                return
+            default:
                 self.delegate?.notifyAddCustomChainFailed(error: .others("\(R.string.localizable.addCustomChainErrorUnknown()) — \($0)"), in: self)
```
hboon, line 83: We can leave it, but generally, I'd prefer if we reserve `switch-default:` for enums and almost never use it (so that we get compilation errors when we introduce new enum values). It's much easier to discover them if we limit them to enums.

eviltofu:
@hboon request review after makings changes

hboon (approved):

PR #3498, opened 2021-12-02
Added Export JSON Keystore screens

### LINKED ISSUE #2719: Allow Private Key and JSON Keystore export for wallets created in AlphaWallet
Metamask has removed the option to import wallet using seed phrase. Now only allows private key or JSON import.
https://metamask.zendesk.com/hc/en-us/articles/360015489331-How-to-import-an-Account

Private key export had been a feature requested several times by users, now without this feature, it is a massive loss of control for users, new and experienced users alike. 
- I was talking to a new user in crypto (5 days experience) today, who wanted to import our wallet to Metamask and was not able to do it.
- Some of the experienced users have mentioned that the ability to export private key is a make or break decision while choosing a wallet. 

@colourfreak - do we have designs for this?

eviltofu:
Closes #2719

# Testing

## Preparation
1. Tap "Settings", "Advanced".
2. **Expects** to see "Advanced" screen with "Export JSON Keystore" option.

## Test Password screen
3. Tap "Export JSON Keystore".
4. **Expects** to see "Set Password for JSON Keystore" screen, with the password text field as first responder, the "Export JSON Keystore" button is disabled.

### Test Enable/Disable "Export JSON Keystore" button
5. Type "abc" into the password text field.
6. **Expects** "Export JSON Keystore" button to be enabled, password text field does not show password.
7. Tap the backspace key to delete a character.
8. **Expects** "Export JSON Keystore" button to be disabled.

### Test character filter (right now only alphanumerics are accepted)
9. Type "+++" into the password text field.
10. **Expects** nothing to happen.

### Test show/hide password
11. Tap the "Eye" button.
12. **Expects** the password to be shown in the password text field.
13. Type "1234567890" into the password text field.
14. **Expects** to see "ab1234567890" in the password field.

## Test Info screen
15. Tap "Export JSON Keystore" button.
16. **Expects** "Export JSON Keystore Info screen" with some random text displayed.

### Test back button
17. Tap the back button.
18. **Expects** to see the "Set Password for JSON Keystore" screen.

### Test complete export
19. Tap "Export JSON Keystore" button.
20. Tap the "Export Button".
21. **Expects** to return to the "Advanced" screen.

# Start Mode
In the `ExportJsonKeystoreCoordinator` class, you can specify which screen to start first via the `start` function.

# Validation rules
The validation rules are a list of rules which can specify the validity of the password. The rules are:
1. lengthLessThan
2. lengthLessThanOrEqualTo
3. lengthMoreThan
4. lengthMoreThanOrEqualTo
5. doesNotContain
6. onlyContain
Rules 1-4 expect an integer. The code does not check for logic. Rules 5-6 expect `CharacterSet` data.

# Test show export page 
### Preparation
You will need an Etherium wallet address. You can get one from [Etherscan](https://etherscan.io/). The one I choose was `0xCD458d7F11023556cC9058F729831a038Cb8Df9c`.

1. Go to the main screen and tap "Wallet".
2. Create a watch only wallet by tapping on the upper left icon in the "Wallet" screen.
3. **Expects** to see "Wallets" screen with list of wallets.
4. Tap the add button in the upper right hand corner and select Watch Wallet.
5.  Enter the wallet address.
6. **Expects** to see the "Wallet" screen.
7. Tap on settings and advanced.
8. **Expects** not to see the Export functions.

## If a Real Wallet

1. Go to the main screen and tap "Wallet".
2. Create a wallet by tapping on the upper left icon in the "Wallet" screen.
3. **Expects to see "Wallets" screen with list of wallets.
4. Tap the add button in the upper right hand corner and select Create/Import.
5. Type in the key phase if you are importing a wallet.
6. **Expects** to see the "Wallet" screen.
7. Tap on settings and advanced.
8. **Expects** to see the Export functions.


eviltofu:
Simulator Screen Shot - iPhone 13 - 2021-12-02 at 20 55 45
Simulator Screen Shot - iPhone 13 - 2021-12-02 at 20 55 51
Simulator Screen Shot - iPhone 13 - 2021-12-02 at 20 56 00
Simulator Screen Shot - iPhone 13 - 2021-12-02 at 20 56 05

AlphaWallet/Settings/Coordinators/ExportJsonKeystoreCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private func startInfoViewController(buttonTitle: String) {
+        let controller = ExportJsonKeystoreInfoViewController(viewModel: ExportJsonKeystoreInfoViewModel(keystore: keystore), buttonTitle: buttonTitle)
+        controller.infoDelegate = self
+        navigationController.pushViewController(controller, animated: true)
+    }
+
+    private func startPasswordViewController(buttonTitle: String) {
+        let validator: StringValidator = StringValidator(rules: [
```
hboon, line 62: I think the definition of the validation rules should be in the view controller (or view model, or model), not in the coordinator.

AlphaWallet/Settings/ViewControllers/ExportJsonKeystorePasswordViewController.swift
```diff
@@ -0,0 +1,88 @@
+//
+//  ExportJsonKeystorePasswordViewController.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 1/12/21.
+//
+
+import UIKit
+
+class ExportJsonKeystorePasswordViewController: UIViewController {
```
hboon, line 10: This should have a Done button instead of Return and tapping the button should hide the keyboard

hboon (changes requested):
There's a glitch when the 2 new view controllers appear, maybe the background color isn't set? (still `.clear`?)

https://user-images.githubusercontent.com/56189/144562392-479539e0-244b-4253-97cf-2a04b7345d3b.mp4


Remember to remove the spurious comments in the git commit message after squashing

hboon:
@eviltofu after the changes, do you want to work on those 2 TODOs too?

So we support 2 types of wallets:

A. Private key
B. HD wallets (so there is a seed and seed phrases). This is used to (internally) generate (A)

We don't allow users to generate A anymore, but allow them to import them. Users can import both types of wallets.

The Keystore/JSON file is generated from A.

Look for this to see how to export it from (A):

```
keystore.exportRawPrivateKeyForNonHdWalletForBackup(forAccount: account, newPassword: newPassword) { [weak self] result in
```

The lines following it also shows how to export it with the sharing sheet.

For B: To export the Keystore/JSON file from B, we'll have to go from B to A first. But I'll work on this later when you are done. The keystore part is a bit more sensitive and it doesn't support this operation yet.

You can check with something like this:

```
if keystore.isHdWallet(account: account) {
    //This is B. Leave it. Leave a TODO
} else {
    //This is A (try implementing this)
}
```

You can verify if it works by importing the JSON file along with the password as a wallet. It should tell you the wallet has already been imported (or exist).

It is also important to check that this is not a watched wallet (no key, no seed phrase) with `Wallet.type`. So only show this new item in Settings > Advanced if this is `.real`.

AlphaWallet/Localization/en.lproj/Localizable.strings
```diff
@@ -97,6 +97,15 @@
 "settings.language.useSystem.title" = "Use System Setting";
 "settings.version.label.title" = "Version";
 "settings.tokenScriptStandard.title" = "TokenScript Compatibility";
+"settings.advanced.exportJSONKeystore.title" = "Export JSON Keystore";
+"settings.advanced.exportJSONKeystore.info.title" = "Export JSON Keystore";
+"settings.advanced.exportJSONKeystore.info.label" = "Your Keystore JSON";
```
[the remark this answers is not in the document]
eviltofu: Can you be more specific? All 3?

AlphaWallet/Settings/Coordinators/ExportJsonKeystoreCoordinator.swift
eviltofu: It should be in the model but we don't have any models. It's passed to the ViewModel. Do you want the ViewModel to automatically have these rules?

AlphaWallet/Settings/Views/ExportJsonKeystoreInfoView.swift
```diff
@@ -0,0 +1,96 @@
...
+//  Created by Jerome Chan on 1/12/21.
+//
+
+import UIKit
+
+class ExportJsonKeystoreInfoView: UIView {
+    private var label: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.font = R.font.sourceSansProRegular(size: 13.0)
```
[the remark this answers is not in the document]
eviltofu: Ok this style usage has to be specified someplace. I copied the styles from the Zeplin mock ups. Same for the other forms I built. Do you want me to start the UIKit Factory class in this PR?

AlphaWallet/Settings/ViewControllers/ExportJsonKeystoreInfoViewController.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  ExportJsonKeystoreInfoViewController.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 1/12/21.
+//
+
+import UIKit
+
+class ExportJsonKeystoreInfoViewController: UIViewController {
```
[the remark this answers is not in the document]
eviltofu: We're not really dealing with a file. `ExportJsonKeystoreDisplayViewController`?

AlphaWallet/Core/Helpers/StringValidator.swift
```diff
@@ -0,0 +1,76 @@
...
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 2/12/21.
+//
+
+import Foundation
+
+typealias StringValidatorResult = Result<Void, StringValidator.Errors>
+
+class StringValidator {
```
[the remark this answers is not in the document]
eviltofu: It's ... in the UnitTests?

AlphaWallet/Localization/en.lproj/Localizable.strings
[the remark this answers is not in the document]
hboon: The button's title which is set to: "settings.advanced.exportJSONKeystore.info.label". That button's title should be "Set password" instead.

AlphaWallet/Settings/Coordinators/ExportJsonKeystoreCoordinator.swift
hboon: Yes. Sounds a bit better than stuffing it in the coordinator (we already do too much of that in existing code)

eviltofu:
https://user-images.githubusercontent.com/1050309/144608075-09530c54-3a79-4657-8213-0f0bc6286479.mp4

hboon:
@eviltofu this isn't blocked by me, right?

eviltofu:
> @eviltofu this isn't blocked by me, right?

This is not blocked by you. It's being blocked by the M1 MacBook Air. Once I finish creating the second issue where I report what fails I'll revert back to this.

eviltofu:
Screen Shot 2021-12-07 at 10 15 06 PM
Screen Shot 2021-12-07 at 10 15 14 PM
Screen Shot 2021-12-07 at 10 15 17 PM

AlphaWallet/Settings/ViewControllers/ExportJsonKeystoreInfoViewController.swift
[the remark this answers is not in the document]
hboon: It is a file because it is known to exist as a "Keystore/JSON file" to cryptocurrency users.

hboon:
Can you not mix in changes related to Dark mode into this PR? Also, I'd rather we not work on Dark mode yet.

eviltofu:
> Can you not mix in changes related to Dark mode into this PR? Also, I'd rather we not work on Dark mode yet.

@hboon The colours are set up with Dark and Light mode in the semantic way we discussed. We have the colours defined in the .xcassets file. Then a struct with static constants that returns the colour based on the user style (dark or light mode). Was I not suppose to build this?

hboon:
Ah, right. You'd wanted to experiment how to build an abstraction for the UI building. The number of modified files had me worried for a minute. Would it be feasible to do this — build this PR without that with 1 commit; it seems like it is almost done. Then follow up with another PR to introduce changes you think is necessary for that abstraction and to support Dark Mode etc? 

Then it's easier to explore, tweak that second PR, drop it, or expand the approach if it's good?

eviltofu:
> Ah, right. You'd wanted to experiment how to build an abstraction for the UI building. The number of modified files had me worried for a minute. Would it be feasible to do this — build this PR without that with 1 commit; it seems like it is almost done. Then follow up with another PR to introduce changes you think is necessary for that abstraction and to support Dark Mode etc?
> 
> Then it's easier to explore, tweak that second PR, drop it, or expand the approach if it's good?

@hboon It's completed right now. Do you still want me to remove the UIFactory code? It's pretty straightforward.

hboon:
> @hboon It's completed right now. Do you still want me to remove the UIFactory code? It's pretty straightforward.

@eviltofu if it's simple, yes, would be good to have a PR without UIFactory etc and focused just on the issue, another introducing UIFactory so we can look at it 1 at a time.

eviltofu:
@hboon removed UIFactory except for the color xcasset.

hboon:
@eviltofu can you help to exclude the UIFactory files too? Seems to be excluded from the project, but they are still in the PR.

eviltofu:
@hboon removed requested files, merged the current upstream/master, and force pushed.

AlphaWallet/Settings/ViewControllers/ExportJsonKeystorePasswordViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+        case .success:
+            computeExportJsonData(password: password)
+        case .failure:
+            break
+        }
+    }
+
+    private func computeExportJsonData(password: String) {
+        view.endEditing(true)
+        navigationController?.displayLoading(text: "Exporting", animated: true)
```
hboon, line 83: Can you help add this to localisation strings?

AlphaWallet/Settings/ViewModels/ExportJsonKeystorePasswordViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+        return validator.validate(string: password)
+    }
+
+    func containsIllegalCharacters(password: String) -> Bool {
+        return validator.containsIllegalCharacters(string: password)
+    }
+
+    func computeHash(password: String, completion: @escaping ExportJsonKeystorePasswordViewModelCompletion) {
+        if keystore.isHdWallet(wallet: keystore.currentWallet) {
+            // TODO: Convert from HDWallet into Non-HDWallet then export
```
hboon, line 39: Do remember to point out this in future PRs, too easy to miss during review

AlphaWallet/Settings/ViewModels/ExportJsonKeystorePasswordViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func validate(password: String) -> StringValidatorResult {
+        return validator.validate(string: password)
+    }
+
+    func containsIllegalCharacters(password: String) -> Bool {
+        return validator.containsIllegalCharacters(string: password)
+    }
+
+    func computeHash(password: String, completion: @escaping ExportJsonKeystorePasswordViewModelCompletion) {
```
hboon, line 37: Do you want to try returning a Promise here instead? We still have lots of completion blocks all over the place, but generally for those that only evaluate once, we should try to use `PromiseKit`. PromiseKit wrapped code is a little more troublesome to write than completion-based ones, but the client code is much clearer.

AlphaWallet/Settings/ViewControllers/AdvancedSettingsViewController.swift
```diff
@@ -35,12 +36,18 @@ class AdvancedSettingsViewController: UIViewController {
...
+    private let keystore: Keystore
     weak var delegate: AdvancedSettingsViewControllerDelegate?
 
-    init(config: Config) {
+    init(keystore: Keystore, config: Config) {
         self.config = config
+        self.keystore = keystore
         super.init(nibName: nil, bundle: nil)
 
+        if keystore.currentWallet.isReal() {
```
hboon, line 47: Can you help add a flag to `Features.swift` like `Features.isExportKeystoreJson`, set it to false, and check it here so the functionality is hidden?

I'll follow up with a PR so it works with HD wallets and enable it.

AlphaWallet/Settings/ViewControllers/ExportJsonKeystorePasswordViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+        DispatchQueue.main.async {
+            self.queue.async {
+                self.viewModel.computeHash(password: password) { result in
```
hboon, line 86: The func `computeHash(password:)`'s name doesn't look intention-revealing

hboon: I think it is slightly better to just let this view controller capture the password, validate it and then pass it on to the next screen which generates the JSON result, rather than generate the JSON and pass it around to reduce places where we capture sensitive data.

BTW, I presume you tested that the async/sync flow works properly with Face ID/Touch ID? Because this line would trigger it:

`keystore.exportRawPrivateKeyForNonHdWalletForBackup(`

and Face ID and Touch ID can be sensitive to which thread you call them from.

hboon (changes requested):

AlphaWallet/Settings/ViewModels/ExportJsonKeystorePasswordViewModel.swift
eviltofu: I'll post an issue once this pull request is merged.

hboon: Ok great. Thanks. But in such a case, would be good to hide the new functionality behind a feature flag in the same PR (described somewhere else in this PR). We release quite aggressively/randomly, so have to make sure `master` is somewhat ready anytime.

eviltofu:
@hboon updated Features.

hboon:
@eviltofu do you mean this is good for review + merge?

eviltofu:
> @eviltofu do you mean this is good for review + merge?

Yes

hboon (approved):

PR #3507, opened 2021-12-07
Added workaround for Makefile problems on M1 Macs

### LINKED ISSUE #3506: Makefile crash when install cocoapods on M1 Macbook Air with `pod install --repo-update`
When installing via `make bootstrap`, cocoapods crashes with the following:

`/Users/jerome/Documents/alpha-wallet-ios/vendor/bundle/ruby/2.6.0/gems/ethon-0.12.0/lib/ethon/curls/classes.rb:35: [BUG] Illegal instruction at 0x0000000100870000
ruby 2.6.8p205 (2021-07-07 revision 67951) [universal.arm64e-darwin21]`

eviltofu:
Fixes #3506

## Testing
1. On an M1 Mac, type `make clean && make bootstrap` in the root directory of the project. 
2. **Expects** The gems and pods to be installed. 
3. Open the AlphaWallet.xcworkspace file and set the target to your physical device. Select run.
4. **Expects** the app to be compiled, installed, and running on your physical device.

## Workaround
The problem appears to occur when pod update or any variation of it is called. The work around simply removes or modifies all makefile targets which invokes it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ target:
 	@echo "install_all  : install gems then pods."
 	@echo "check_gems   : check to see if all the gems in the Gemfile are installed in the vendor directory."
 	@echo "bootstrap    : same as make install_all"
-	@echo "update_pods  : update all the cocoapods."
 
 check_bundle:
 	@$(bundle_cmd) --version 1>/dev/null 2>/dev/null; \
@@ -38,20 +37,11 @@ install_gems: check_bundle
 	fi
 
 install_pods: check_gems
-	@$(bundle_cmd) exec pod install --repo-update; \
+	@$(bundle_cmd) exec pod install; \
 	if [ $$? -eq 0 ]; then \
 		echo "All pods installed."; \
 	else \
 		echo "Error installing."; \
-		exit 1; \
-	fi
-
-update_pods: check_gems
-	@$(bundle_cmd) exec pod update; \
-	if [ $$? -eq 0 ]; then \
-		echo "All pods updated."; \
-	else \
-		echo "Error updating."; \
 		exit 1; \
 	fi
 
--- a/README.md
+++ b/README.md
@@ -70,7 +70,11 @@ This makefile has been tested to run on "Monterey"-12.0.1. It will not work on "
 
 After the Gemfile is updated, run `make install_gems` to update the gems in the vendor/bundle directory.
 
-After the Podfile is updated, run `make install_pods` or `make update_pods` to update the pods in the Pods directory.
+After the Podfile is updated, run `make install_pods` to update the pods in the Pods directory.
+
+### M1 Macs
+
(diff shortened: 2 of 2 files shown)


hboon:
@eviltofu do you want to try if a newer Cocoapods version helps?

eviltofu:
> @eviltofu do you want to try if a newer Cocoapods version helps?

It appears to be some cocoapods setting where making an arm64 version of the code for debugging is flagged as an error. I will post the errors in the next issue.

eviltofu:
@hboon How will you test this without an M1 Mac?

hboon:
> @hboon How will you test this without an M1 Mac?

By trusting you! At least verify it doesn't break non-M1s. I don't believe we should review every PR in detail anyway.

(On the other hand, I do have an M1. It's much delayed, but coming in a few days)

eviltofu:
@hboon needs review on intel Mac

Makefile
```diff
@@ -38,23 +37,14 @@ install_gems: check_bundle
...
+	@$(bundle_cmd) exec pod install; \
 	if [ $$? -eq 0 ]; then \
 		echo "All pods installed."; \
 	else \
 		echo "Error installing."; \
 		exit 1; \
 	fi
 
-update_pods: check_gems
-	@$(bundle_cmd) exec pod update; \
```
hboon, line 50: I know this was added in a separate PR, but I don't think we'll ever want to run `pod update`. It'll almost always be `pod install`, and rarely `pod update <podname>`

hboon (approved):
Thanks. Would you help modify the PR title and, at least the commit message to be more specific about what is the problem, (as a matter of habit)? I've modified the title of #3506 if it helps.

hboon:
@eviltofu And generally run `git rebase master` as a habit when we push/force push? It's not so important for this PR as of writing, but it would generally help reduce conflict when merging and keep the new commit(s) up to date

Makefile
eviltofu: Do you want me to remove it?

eviltofu:
> @eviltofu And generally run `git rebase master` as a habit when we push/force push? It's not so important for this PR as of writing, but it would generally help reduce conflict when merging and keep the new commit(s) up to date

I think I did that when I created the PR last week but you just kept on merging more things? Do I have to keep updating PR daily if they are not merged? Also should it not be `git  rebase upstream/master`?

hboon:
Yes. Keep rebasing master so it’s updated. But don’t have to keep each PR updated daily. More like make sure it’s rebased when you push or force push again (again, it’s a single command and if it’s hard for that particular PR, it’s just going to be more important to do it that time rather than later to fix the conflicts) and once more just prior to merging (whoever does the merge should prompt you if necessary).

In addition, if it’s not rebased and behind master, that means it’s harder to see the change against that is created by the PR once it’s merged.

> git rebased upstream/master

Works too and is more complete. I had presume you would keep your local master updated :)

Issue #3460, opened 2021-11-22
Make Buy Crypto option more available and visible


colourfreak:
There are 3 things we can do. I would definitely go for 1., and keep 2. and 3. optional.

1. Add an additional item in the hidden menu called "Buy Crypto"
2. Add a button to buy crypto at the end of the token list
3. Add one more icon in the Tab Bar.

<img width="1503" alt="Screenshot 2021-11-22 at 15 08 38" src="https://user-images.githubusercontent.com/51817359/142876312-bf70609d-a506-40fb-8832-12bfa762bf85.png">

hboon:
@colourfreak thanks! Yeah, 2 and 3 probably would be later like you said. But I think we ought to expand the feature discovery a bit more beyond (1). We already do that when the user tries to send a transaction but don't have enough for gas (will have to check it opens up the Ramp page easily, I forgot how we do it). For example, maybe we can do a one-time prompt similar to prompting for newsletter for the users if they have very low ETH (we'll just hardcode a value to check against) and some other tokens. I'll think about this a bit more.

PR #3513, opened 2021-12-08
Add fiat on ramp to Wallet tab

hboon:
Part of: #3460


hboon:

vladyslav-iosdev (approved):

PR #3521, opened 2021-12-09
Corrected code that dismisses WhatsNewListingViewController but prevents other views from receiving user interaction

### LINKED ISSUE #3520: Bug: When starting app for the first time, WhatsNewList causes screen freeze.
# Bug report
## Delete app from device ###
1. Install new copy of app.
2. Start app.
3. Create new wallet.
4. **What's New** sheet appears. **Back up your wallet** appears behind the sheet.
5. Press X on the **What's New** sheet. **What's New** sheet disappears.
6. Tapping on the green CONTINUE button no longer works dismisses **Back up your wallet**..

However, if you slide the **What's New** sheet down, you can tap on the green continue button and **Back up your wallet** disappears.

Tested on iOS 13 (iPhone 11 simulator), iOS 14 (iPhone 6S Plus), iOS 15 (iPhone X).

eviltofu:
Closes #3520

## Testing
1. Delete app.
2. Install new copy of app.
3. Start app.
4. Create new wallet.
5. Tap X on "WhatsNewListing".
6. **Expects** "WhatsNewListing" to be dismissed.
7. Tap CONTINUE.
8. **Expects** "Back up" to be dismissed.

### DIFF
--- a/AlphaWallet/WhatsNew/ViewControllers/WhatsNewListingViewController.swift
+++ b/AlphaWallet/WhatsNew/ViewControllers/WhatsNewListingViewController.swift
@@ -41,12 +41,18 @@ class WhatsNewListingViewController: ModalViewController {
 
 extension WhatsNewListingViewController: ModalViewControllerDelegate {
     func didDismiss(_ controller: ModalViewController) {
-        whatsNewListingDelegate?.didDismiss(controller: self)
+        controller.dismissViewAnimated {
+            controller.dismiss(animated: false) {
+                self.whatsNewListingDelegate?.didDismiss(controller: self)
+            }
+        }
     }
 
     func didClose(_ controller: ModalViewController) {
         controller.dismissViewAnimated {
-            self.whatsNewListingDelegate?.didDismiss(controller: self)
+            controller.dismiss(animated: false) {
+                self.whatsNewListingDelegate?.didDismiss(controller: self)
+            }
         }
     }
 }


AlphaWallet/WhatsNew/ViewControllers/WhatsNewListingViewController.swift
```diff
@@ -46,7 +46,9 @@ extension WhatsNewListingViewController: ModalViewControllerDelegate {
 
     func didClose(_ controller: ModalViewController) {
         controller.dismissViewAnimated {
```
hboon, line 48: The logic seems a bit strange. From reading at this client code, it appears that `controller` has to be dismissed twice. Is this something that should be fixed in ModalViewController instead?

hboon (changes requested):
Ah, good find.

I should probably change it so WhatsNewListingViewController doesn't gets used right after a wallet is created (or just add a delay). I'll look into it separately.

Regarding the PR:

I think, it's better to include a screenshot which illustrates the problem, a bit easier/faster to grasp it even before running the reproduction steps.

> Corrected code that dismisses WhatsNewListingViewController

I'd also suggest improving the commit message (and PR title) to describe the problem. This might not be specific to when a wallet is created, but it seems like it really is only a problem there right?

eviltofu:
> I think, it's better to include a screenshot which illustrates the problem, a bit easier/faster to grasp it even before running the reproduction steps.

I put two images in #3520 that sorts of explains the problem. Next time I will put in a movie.

AlphaWallet/WhatsNew/ViewControllers/WhatsNewListingViewController.swift
eviltofu: I'm wary of fixing it there because I don't know what code which uses it is doing. I might have to go fix all the instances  where it's used. I'll post another issue and make it an app wide change if you want.

hboon: Ah, good point. See if it's a problem elsewhere then. 

> > I think, it's better to include a screenshot which illustrates the problem, a bit easier/faster to grasp it even before running the reproduction steps.
> 
> I put two images in #3520 that sorts of explains the problem. Next time I will put in a movie.

Great! But if images display the issue, use that instead of movies. The latter is slower to comprehend.

eviltofu:
> I should probably change it so WhatsNewListingViewController doesn't gets used right after a wallet is created (or just add a delay). I'll look into it separately.

See this [image](https://github.com/AlphaWallet/alpha-wallet-ios/issues/3520#issuecomment-989774594). Leftovers from the View Controller are placed on top of the existing views so taps never reach them.

hboon:
> > I think, it's better to include a screenshot which illustrates the problem, a bit easier/faster to grasp it even before running the reproduction steps.
> 
> I put two images in #3520 that sorts of explains the problem. Next time I will put in a movie.

Great! But if images display the issue, use that instead of movies. The latter is slower to comprehend.

AlphaWallet/WhatsNew/ViewControllers/WhatsNewListingViewController.swift
eviltofu: Four instances use this. 
1. WalletConnectConnectionTimeoutViewController
2. SwitchChainRequestViewController
3. PromptViewController
4. CollectUsersEmailViewController
5. WhatsNewListingViewController

All four execute a variation of the following:
```
        dismissViewAnimated(with: {
            self._delegate?.didClose(in: self)
            self.dismiss(animated: false)
        })
```

hboon: Ah, I see. Maybe create a separate issue/PR to move the call for `elf.dismiss(animated:)` into `dismissViewAnimated()` and modify the name if necessary?

But for this PR, just fix this particular case. But can you double check it this: the 4 examples you listed do not access `controller`, they refer to `self` instead.

eviltofu: @hboon Here `self` is the same as `controller` since the other classes are delegates for themselves and this delegate is set in the `ModalViewController`'s `init` function. If you did change the delegate in the other classes, the code won't work.

hboon:
@eviltofu is this ready?

eviltofu:
> @eviltofu is this ready?

Once the Travis CI tests pass.

hboon (approved):

PR #3522, opened 2021-12-09
Update to latest WalletConnect v1.x SDK #3489

### LINKED ISSUE #3489: Update to latest WalletConnect v1.x SDK
Check diffs against `1.3.1` which we are currently using https://github.com/WalletConnect/WalletConnectSwift/releases. Some changes mentioned might break compatibility?

> Add new Server and Client delegate methods

> v1.4.0 update CryptoSwift version

If it is too much breakage, we should consider not upgrading at all.

vladyslav-iosdev:
Closes #3489


AlphaWallet/WalletConnect/WalletConnectServer.swift
```diff
@@ -382,6 +344,53 @@ extension WalletConnectServer: ServerDelegate {
...
+
+    private func removeSession(for url: WalletConnectURL) {
+        guard var sessions = _sessions.value else { return }
+
+        if let index = sessions.firstIndex(where: { $0.url.absoluteString == url.absoluteString }) {
+            set(server: nil, for: sessions[index].url)
+            sessions.remove(at: index)
+        }
+
+        UserDefaults.standard.walletConnectSessions = sessions
```
hboon, line 375: Better to store a Config instance and read its `defaults` to have the correct `UserDefaults` instance?

hboon (approved):
Seems ok. Do you foresee any problem with upgrading `CryptoSwift`? It was quite painful the last time it was upgraded, right?

hboon:
@vladyslav-iosdev BTW, can you verify that `Podfile.lock` was committed correct? I get a different checksum after I do `pod install`.

vladyslav-iosdev:
> Do you foresee any problem with upgrading CryptoSwift?

 it was need to update because of `Wallet Connnect` requires  `CryptoSwift` greater then `~> 1.4`

hboon (approved):

PR #3526, opened 2021-12-10
Reduce image reloading, if it hasn't changed #3525

### LINKED ISSUE #3525: improve collectibles displaying

vladyslav-iosdev:
Closes #3525

### DIFF
--- a/AlphaWallet/Tokens/Views/OpenSea/OpenSeaNonFungibleTokenViewCell.swift
+++ b/AlphaWallet/Tokens/Views/OpenSea/OpenSeaNonFungibleTokenViewCell.swift
@@ -136,21 +136,26 @@ class OpenSeaNonFungibleTokenPairTableCell: UITableViewCell {
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
 
-        background.translatesAutoresizingMaskIntoConstraints = false
-        contentView.addSubview(background)
         selectionStyle = .none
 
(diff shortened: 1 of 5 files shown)


AlphaWallet/Tokens/Views/OpenSea/OpenSeaNonFungibleTokenViewCell.swift
```diff
@@ -67,6 +68,7 @@ class OpenSeaNonFungibleTokenView: UIView {
         imageHolder.clipsToBounds = true
 
         if let url = viewModel.imageUrl {
+            guard self.viewModel?.imageUrl != url else { return }
```
hboon, line 71: Keeping in the `viewModel` around as a property isn't very clear (in the future). Do you think these 2 alternatives would be better:

A. instead of storing `viewModel`, store a `URL` instead with a name that includes it's currently loaded
B. store the URL in a backing property in `WebImageView` and perform a check in the `WebImageView.url` getter?

hboon (changes requested):
There seems to be some unrelevant changes, can you help split it up into separate commits, at least?

hboon:
@vladyslav-iosdev Good for review?

hboon:
@vladyslav-iosdev Sorry, think I missed the window earlier. Is this good for review?

vladyslav-iosdev:
> @vladyslav-iosdev Sorry, think I missed the window earlier. Is this good for review?

yep

hboon (approved):

hboon (approved):

PR #3533, opened 2021-12-12
Workaround to build on Apple M1 + run on simulator (device is fine)

hboon:
By doing this for simulators:

* Forcing build to be for x86_64
* Excluding MailChimp SDK

A simpler alternative is to run Xcode (didn't try Xcode) in Rosetta (in Finder, right click `Xcode.app` and check "Open using Rosetta"), but maybe this would make development and build speed take a hit.

Tried a few other approaches which didn't work:

* Including MailChimp SDK as a pod (on hindsight, it's probably because of the ffi problem)

---

There also appears to be a crash when running `pod install`. This is probably due to the `ffi` gem. This is probably what #3506 is about

@vladyslav-iosdev help check `pod install` and running on simulator and device works for you on non-M1 (I have verified that this works for both M1 and non-M1 on my side)


hboon:
cc @eviltofu. You mentioned this issue with M1.

vladyslav-iosdev (approved):

PR #3536, opened 2021-12-13
Do not show What's New after wallet creation as it crashes with the back up prompt

hboon:
i.e. fix #3502 so it doesn't clash with the following:

Screenshot 2021-12-13 at 1 54 11 PM


hboon:
Good for review (had forgot to remove unwanted commits)

vladyslav-iosdev (approved):

PR #3539, opened 2021-12-13
Sort tokens Value: Descending in the Wallet tab by default #3528

### LINKED ISSUE #3528: Sort tokens Value: Descending in the Wallet tab by default
Native coin should go first for a network, then the rest with the value descending. Usually, you are the most interested in tokens that have the highest value in your wallet.

- Ethereum
- Token with the highest value in USD
- Token with the second highest value in USD
- Token with the third highest value in USD
- ...

**Now:**

**Should be:**
Screenshot 2021-12-10 at 12 24 48

vladyslav-iosdev:
Closes #3528

### DIFF
--- a/AlphaWallet/Core/Coordinators/FilterTokensCoordinator.swift
+++ b/AlphaWallet/Core/Coordinators/FilterTokensCoordinator.swift
@@ -108,6 +108,21 @@ class FilterTokensCoordinator {
     }
 
     func sortDisplayedTokens(tokens: [TokenObject]) -> [TokenObject] {
+
+        func sortTokensByFiatValues(_ token1: TokenObject, _ token2: TokenObject) -> Bool {
+            let value1 = coinTickersFetcher.ticker(for: token1.addressAndRPCServer).flatMap({ ticker in
+                EthCurrencyHelper(ticker: ticker)
+                    .fiatValue(value: token1.optionalDecimalValue)
+            }) ?? -1
+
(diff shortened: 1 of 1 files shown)


hboon (changes requested):
Descending by fiat value, not by name :)

So those with fiat value first, most $ first and then followed by those without fiat value, by name ascending.

Now that we put it that way, let's see it in action if it makes sense..

hboon:
@vladyslav-iosdev Good for review?

vladyslav-iosdev:
@hboon bump

hboon:
@vladyslav-iosdev by the fiat value (eg $10k of ETH) not the ticket price.

vladyslav-iosdev:
@hboon not it should be ok

hboon (changes requested):
Seems wrong (eg. $1.4 and $0.7 should be before $0):

Screenshot 2021-12-16 at 10 59 21 AM

vladyslav-iosdev:
@hboon if i get it right, we don't sort eth by fiat value, and it's going to be first all the time, right?

hboon:
> @hboon if i get it right, we don't sort eth by fiat value, and it's going to be first all the time, right?

Yes, you are right. I missed that. The native token for each chain goes first, no matter its value.

vladyslav-iosdev:
> Yes, you are right. I missed that. The native token for each chain goes first, no matter its value.

ok, now should be fine

hboon (approved):

PR #3543, opened 2021-12-14
Installed Ruby 3 via rbenv to fix makefile cocoapods crashes when --update-repo option is used.

### LINKED ISSUE #3506: Makefile crash when install cocoapods on M1 Macbook Air with `pod install --repo-update`

eviltofu:
Fixes #3506
Closes #3507

# Setup instructions for rbenv
## Install `rbenv`
Follow the [rbenv and ruby-build installtion instructions](https://github.com/rbenv/rbenv#installation) on the `rbenv` website.
## Go to directory of alpha-wallet-ios directory
On the command line, go to the alpha-wallet-ios directory.
## Install `Ruby 3.0.3`
Type `rbenv install 3.0.3`. This will install a copy of Ruby 3.0.3.
## Set up local ruby.
Type `rbenv local 3.0.3`. This will make Ruby 3.0.3 the version of Ruby which runs in this directory.
Type `rbenv local` to verify it.
## Install bundler
Type `gem install bundler`. This will install a copy of bundler for Ruby 3.0.3.
## Install gems and cocoapods
Type `make install_all`. This will load all the gems and then install all the cocoapods.

# Install instructions if you do not want to install Ruby 3.0.3
Type `make install_gems && bundle exec pod install`.


hboon:
Pending https://github.com/AlphaWallet/alpha-wallet-ios/issues/3506#issuecomment-993610114

eviltofu:
@hboon update on this?

Gemfile.lock
```diff
@@ -29,192 +48,229 @@ GEM
...
     xcpretty (0.3.0)
       rouge (~> 2.0.7)
-    xcpretty-travis-formatter (1.0.0)
+    xcpretty-travis-formatter (1.0.1)
       xcpretty (~> 0.2, >= 0.0.7)
+    zeitwerk (2.5.1)
 
 PLATFORMS
-  ruby
+  arm64-darwin-21
```
hboon, line 273: Seems like this should still include both `ruby` and `arm64-darwin-21`

.gitignore
```diff
@@ -97,3 +97,4 @@ AlphaWalletTests/Snapshots/FailureDiffs
 .idea/
 /.bundle
 /vendor
+/.ruby-version
```
hboon, line 100: Seems like `.ruby-version` should be committed, not ignored.

hboon (changes requested):
Works great. Just minor tweaks.

.gitignore
eviltofu: This will prevent people from having different versions of Ruby. If someone updates their version, every one will have to update to this version if they are using `rbenv` to manage this directory.

eviltofu:
Force push so that I can test this on the intel platform.

.gitignore
hboon: `rbenv` lets developers install multiple versions of ruby and activate a specific version in each project. `3.0.3` is the known version that works. Take a counter example, if they use ruby 2.x instead, `make` will fail. So we do want them to install and use `3.0.3`. If we update to a newer version that works and commits it, then yes, everyone should use that version. That's the point of `rbenv`; each projects pins to a specific version (ditto for `Gemfile.lock`, `Podfile.lock`).

eviltofu:
Tested on Intel and M1 by running `make install_all` at the command line and running all the unit tests in the workspace for 15.2 and 14.5.

eviltofu:
@hboon 

Ok, so the current problem now is that we have to use the same version of `Bundler` or there would be a conflict whenever someone who uses `Ruby 3` commits followed by someone who uses `Ruby 2`. The current `Gemfile.lock` file ends with 
```
BUNDLED WITH
   1.17.2
```
`Ruby 3` uses `2.2.32`. `Ruby 2` uses `1.17.2`. So in order to be compatible, everyone has to use the same version 1.17.2.
We can install `Bundler 1.17.2` in `Ruby 3` and the system will use that automatically. But this defeats the purpose of upgrading to `Ruby 3`. We might as well modify the `Makefile` to compile with the system provided `Ruby 2`.

What should I do?

hboon:
With the changes in this PR, especially this:

```
BUNDLED WITH
   2.2.33
```

Add instructions (if it's not already there) and modify the `Makefile` to run `gem install bundler` (without version number) before `bundle`.

A. If the developer uses Ruby 3.0.3
---
If rbenv is installed, it'll read `.ruby-version`, which is why it should be checked in), it'll work

B. If the developer uses Ruby 2.x
---
Say they have bundler version 2.2.22, it'll still work with a warning like this:

> Warning: the running version of Bundler (2.2.22) is older than the version that created the lockfile (2.2.33). We suggest you to upgrade to the version that created the lockfile by 
> running `gem install bundler:2.2.33`.

The "BUNDLED WITH" value wouldn't be modified.

Maybe it'll fail with much older, incompatible versions, but the error message should quite clear which bundler version is needed

eviltofu:
> Add instructions (if it's not already there) and modify the `Makefile` to run `gem install bundler` (without version number) before `bundle`.

We should specify the version of bundler to match the Gemfile.lock or the makefile will just fail when if Apple or someone updates the system version of Ruby or the version of bundler is updated to a higher version number.

eviltofu:
`bundle exec pod install --repo-update` still causes a crash when running under `Ruby 2.6`. I am going to remove `--repo-update`.

hboon:
Up to you.

eviltofu:
@hboon The .ruby-version file is interfering with the travis build.

hboon:
@eviltofu can't get it to work?

eviltofu:
Travis builds are failing. I'm trying to figure out a solution.

eviltofu:
@hboon review required

hboon (approved):

PR #3547, opened 2021-12-14
Fixed code on Wallet main screen that crashed when token row was swiped to delete

### LINKED ISSUE #3537: Swiping to delete a wallet on main screen Wallet Tab causes app to crash
Commit hash: e4581507a0718abbeabb55cb52814497b9610cfd

On master, on the Wallet screen, when I swipe to delete on both my iPhone X (15.1) and iPhone 6S Plus (14.8.1), I get the following error message and the app crashes. Built on M1 Air.
```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid update: invalid number of rows in section 3. The number of rows contained in an existing section after the update (2) must be equal to the number of rows contained in that section before the update (4), plus or minus the number of rows inserted or deleted from that section (0 inserted, 1 deleted) and plus or minus the number of rows moved into or out of that section (0 moved in, 0 moved out).
```
It looks like when the entry is deleted, the table view cell above it is not removed.
Screen Shot 2021-12-13 at 5 35 41 PM

eviltofu:
Fixes #3537.

## Prerequisites
You need some items to display on the Wallet main screen. Add some if you have none.
## Testing
1. Swipe left on a token displayed in the Wallet main screen.
2. **Expects** delete icon "X" appears.
3. Tap on "X".
4. **Expects** deleted token disappears from Wallet main screen.

Screen Shot 2021-12-15 at 7 32 28 AM copy

## Side note:
The token rows displayed on the view controller consists of an array of `TokenObjectOrRpcServerPair` objects. `TokenObject` is used to represent a token while `RpcServerPair` is used to represent the header. We need to consider a minimum of 3 rows (previous to the swiped row, the swiped row, next to the swiped row) in order to determine the outcome. If a row does not exist, it is treated as Not removable. If the row is a `RpcServerPair`, the row is treated as Not removable. If the row is a `TokenObject`, it is treated as removable for this case,

| Previous | Current | Next | Outcome | Notes |
| --- | --- | --- | --- | --- |
| Not removable | Not removable | Not removable | Don't remove | Possibly an error condition but we do some defensive programming |
| Not removable | Not removable | Removable | Don't remove | Possibly an error condition but we do some defensive programming |
| Not removable | Removable | Not removable | Remove Previous and Current | The only token row in the section |
| Not removable | Removable | Removable | Remove Current | First row in a section with two or more token rows |
| Removable | Not removable | Not removable | Don't remove | Possibly an error condition but we do some defensive programming |
| Removable | Not removable | Removable | Don't remove | Possibly an error condition but we do some defensive programming|
| Removable | Removable | Not removable | Remove Current | Last row in a section with two or more token rows |
| Removable | Removable | Removable | Remove Current | Not the first or last row in a section with two or more token rows |

We cannot use 2 rows to determine the outcome. A not removable followed by a removable row can indicate one of two outcomes (remove current row or remove previous and current row) as you can see in the table above.

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
@@ -613,11 +613,12 @@ extension TokensViewController: UITableViewDataSource {
             let hideAction = UIContextualAction(style: .destructive, title: title) { [weak self] (_, _, completion) in
                 guard let strongSelf = self else { return }
 
+                let deletedIndexPathArray = strongSelf.viewModel.indexPathArrayForDeletingAt(indexPath: indexPath)
                 strongSelf.delegate?.didHide(token: token, in: strongSelf)
 
                 let didHideToken = strongSelf.viewModel.markTokenHidden(token: token)
                 if didHideToken {
-                    strongSelf.tableView.deleteRows(at: [indexPath], with: .automatic)
+                    strongSelf.tableView.deleteRows(at: deletedIndexPathArray, with: .automatic)
                 } else {
                     strongSelf.reloadTableData()
                 }
--- a/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
+++ b/AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
@@ -234,7 +234,54 @@ class TokensViewModel {
             return nil
         }
     }
+
+    func indexPathArrayForDeletingAt(indexPath current: IndexPath) -> [IndexPath] {
+        let canRemoveCurrentItem: Bool  = item(for: current.row, section: current.section).isRemovable
+        let canRemovePreviousItem: Bool = current.row > 0 ? item(for: current.row - 1, section: current.section).isRemovable : false
+        let canRemoveNextItem: Bool = {
+            guard (current.row + 1) < filteredTokens.count else { return false }
+            return item(for: current.row + 1, section: current.section).isRemovable
+        }()
+        switch (canRemovePreviousItem, canRemoveCurrentItem, canRemoveNextItem) {
+            // Truth table for deletion
+            // previous, current, next
+            // 0, 0, 0
+            // return []
+            // 0, 0, 1
+            // return []
+            // 0, 1, 0
+            // return [current.previous, current]
+            // 0, 1, 1
+            // return [current]
+            // 1, 0, 0
+            // return []
+            // 1, 0, 1
+            // return []
+            // 1, 1, 0
+            // return [current]
+            // 1, 1, 1
+            // return [current]
+        case (_, false, _):
+            return []
+        case (false, true, false):
+            return [current.previous, current]
+        default:
+            return [current]
+        }
+    }
 }
+
+fileprivate extension TokenObjectOrRpcServerPair {
+    var isRemovable: Bool {
+        switch self {
+        case .rpcServer:
+            return false
+        case .tokenObject:
+            return true
+        }
+    }
+}
+
 
 fileprivate extension WalletFilter {
     static var orderedTabs: [WalletFilter] {
@@ -306,5 +353,11 @@ extension TokensViewModel.functional {
                 return true
             }
         }
+    }
+}
+
+fileprivate extension IndexPath {
+    var previous: IndexPath {
+        IndexPath(row: row - 1, section: section)
     }
 }


eviltofu:
@hboon I don't know how to add more items to this screen. Also, I'm trying to add a token from the Add / Hide Token menu and all I seem to get are "Add token failure" results.

eviltofu:
Ok so I'm looking at the next issue and it dawns on me that there could be multiple items in each section. So this might take a bit more time.

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -555,7 +555,7 @@ extension TokensViewController: UITableViewDataSource {
 
                 let didHideToken = strongSelf.viewModel.markTokenHidden(token: token)
                 if didHideToken {
-                    strongSelf.tableView.deleteRows(at: [indexPath], with: .automatic)
+                    strongSelf.tableView.deleteRows(at: indexPath.indexPathArrayForTokenType(), with: .left)
```
hboon, line 558: Why `.left` and not `.automatic`?

hboon:
> Ok so I'm looking at the next issue and it dawns on me that there could be multiple items in each section. So this might take a bit more time.

>  I don't know how to add more items to this screen. When I try to add a token from the Add / Hide Token menu and all I seem to get are "Add token failure" results.

You can watch the wallet 0x983110309620d911731ac0932219af06091b6744 which has some tokens

AlphaWallet/Tokens/ViewControllers/TokensViewController.swift
eviltofu: Automatic looks horrible when animated.

eviltofu: I've switched it back to automatic. It only appears to be horrible only when you have a few entries.

eviltofu:
@hboon ready for review

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -172,8 +172,57 @@ class TokensViewModel {
             return nil
         }
     }
+
+    func indexPathArrayForDeletingAt(indexPath current: IndexPath) -> [IndexPath] {
+        let canRemoveCurrentItem: Bool  = item(for: current.row, section: current.section).isRemovable
+        let canRemovePreviousItem: Bool = {
+            current.row > 0 ? item(for: current.row - 1, section: current.section).isRemovable : false
+        }()
+        let canRemoveNextItem: Bool = {
```
hboon, line 181: When can we remove the next item? Seems like there should only be 3 possibilities when we swipe a token:

A. The token can't be removed because it is a native token. Taken care by a check against `TokenObjectOrRpcServerPair.canDelete` or similar
B. There is more than 1 token in the section (group). The token is removed
C. There is exactly 1 token in the section (group). The token as well as the header is removed

hboon (changes requested):
In the future, would you help `git rebase master` as a habit when you push/force push? It's not so important for this PR as of writing, but it would help reduce conflicts when merging and keep the new commit(s) up to date. Good to make it the rule rather than exception since it's just 1 command (and if it's difficult to resolve for a particular PR, then it's even more important to run it prior attempting to merge).

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
eviltofu: It’s to determine if the item is the last one in its “section” or not. The current code does not break the tokens into sections but into one large section and uses rpcServers to display the header. `canDelete`
 doesn’t work because the token is hidden and not actually deleted.

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -172,8 +172,57 @@ class TokensViewModel {
...
+        case (false, true, false):
+            return [current.previous, current]
+        default:
+            return [current]
+        }
+    }
 }
 
+fileprivate extension TokenObjectOrRpcServerPair {
+    var isRemovable: Bool {
```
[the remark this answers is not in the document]
eviltofu: It’s only used in this particular case.

AlphaWallet/Tokens/ViewModels/TokensViewModel.swift
hboon: > The current code does not break the tokens into sections but into one large section and uses rpcServers to display the header. 

Right. But I still don't understand why we ever consider if we can remove the next token (i.e. 0, 1, 1 in the truth table). Can you give the simplest example of how it would look like in the Wallet tab and which row is being swiped?

Actually, I think I am starting to understand your point after writing this out, do you think you can modify the PR to make it clearer or as a last resort, add a comment to the function `indexPathArrayForDeletingAt(indexPath:)`?

hboon: > canDelete doesn’t work because the token is hidden and not actually deleted. This was the first thing I tried.

Just to be sure: Tokens are only hidden when the user swipes (no matter if the function names calls them hide or delete. It ought to be the same action). We can't delete tokens from the chain nor really delete in the app (we do remove/move the token entries in the Realm database though, so you might be referring to that).

With that in mind, can you help think/describe that again why it doesn't work? And these 2 checks are important:

```
guard !token.isInvalidated else { return false }
if token.contractAddress.sameContract(as: Constants.nativeCryptoAddressInDatabase) {
```

The 1st line is a defensive check against Realm crashing on us, the next line makes sure we don't hide the native token for each chain

eviltofu: `canDelete` doesn't work because it does what it says. The property tells you it can't be deleted. We want to remove the token row from the tableview. `canDelete` doesn't tell me if the token row can be removed since it will always return false for those native tokens for each chain. Swiping on Ethereum will never remove it from the Wallet main screen. 

This is based on the behaviour of the original code which allows Ethereum to be removed. If this is not the case, then a new issue should be filed since this issue deals only with crashing while swiping to delete.

eviltofu:
Based on last nights conversation, it appears that I’m solving the wrong problem. Do not merge this.

eviltofu:
Just an update, I think this problem is due to the code allowing items on the wallet main screen to be deleted. Can @hboon or @colourfreak tell me what should happen on this screen when a user swipes to delete an item?

hboon:
To the user, that token will disappear from the Wallet tab. And in `Add / Hide Tokens` (Wallet tab > … > Add / Hide Tokens), that token will no longer appear under `Displayed Tokens` and will appear in `Hidden Tokens`.

For the Wallet tab, let's keep it as the users being to swipe and hide all tokens except for the native token (i.e. address = `Constants.nativeCryptoAddressInDatabase`). It is still possible to hide those in `Add / Hide Tokens` screen, but not allowing that in Wallet tab guards against users accidentally swiping their native tokens and not being able to get it back.

eviltofu:
> To the user, that token will disappear from the Wallet tab. And in `Add / Hide Tokens` (Wallet tab > … > Add / Hide Tokens), that token will no longer appear under `Displayed Tokens` and will appear in `Hidden Tokens`.
> 

Even if the user swipes on Ethereum, xDai, Polygon Mainnet? Is it possible for this screen to be empty? @hboon

hboon:
No. Ethereum (the token), xDai etc are the native crypto token (their addresses are stores as that constant mentioned above) for their respective chains. Those can’t be swipe to hide in the Wallet tab. (We might change it later). But the user can still choose to hide them in the Add/Hide screen and thus, yes all the entries can end up being hidden in the Wallet tab

eviltofu:
> No. Ethereum (the token), xDai etc are the native crypto token (their addresses are stores as that constant mentioned above) for their respective chains. Those can’t be swipe to hide in the Wallet tab. (We might change it later). But the user can still choose to hide them in the Add/Hide screen and thus, yes all the entries can end up being hidden in the Wallet tab

Ok because currently the master branch does allow them to be swiped and deleted (which is the cause of this bug).

hboon:
@eviltofu this waiting for me?

eviltofu:
> @eviltofu this waiting for me?

No, I just need to figure out what’s wrong.

eviltofu:
The problem is that the current code allows native crypto currencies to be deleted. So the solution to this issue is to prevent that from happening. (Note to myself).

eviltofu:
@hboon ready for review. Merge after #3721.

hboon:
Pushed after rebase master

hboon (approved):

PR #3579, opened 2021-12-20
24 word seed phrase feature

### LINKED ISSUE #3447: Allow users to import a wallet with 24 seed phrase
Refer https://github.com/AlphaWallet/alpha-wallet-android/issues/2204#issuecomment-972776818

- [x] Also need to check for references to "12 words seed phrases" or similar in the app
- [x] Also need to check if export functionality needs to be changed
- [x] Hide functionality behind a `Features` flag.
- [ ] Check verify seed phrase screen works for both 12-words and 24-words. This might not be accessible on the simulator due to #3609. This is accessible via Settings > Back up this Wallet > Back up my Wallet
  
* Let's just not allow user to create them at the moment

eviltofu:
Closes #3447

This branch is for testing the 24 word seed import feature and what other UI it might impact.

- [ ] This pull request needs to be squashed before merging.

### DIFF
--- a/AlphaWallet/KeyManagement/EtherKeystore.swift
+++ b/AlphaWallet/KeyManagement/EtherKeystore.swift
@@ -341,9 +341,9 @@ open class EtherKeystore: NSObject, Keystore {
     private func doesSeedMatchWalletAddress(mnemonic: String) -> Bool {
         guard let wallet = HDWallet(mnemonic: mnemonic, passphrase: emptyPassphrase) else { return false }
         guard wallet.mnemonic == mnemonic else { return false }
-        let seed = HDWallet.computeSeedWithChecksum(fromSeedPhrase: mnemonic)
         guard let walletWhenImported = HDWallet(entropy: wallet.entropy, passphrase: emptyPassphrase) else { return false }
         //If seed phrase has a typo, the typo will be dropped and "abandon" added as the first word, deriving a different mnemonic silently. We don't want that to happen!
+
         guard walletWhenImported.mnemonic == mnemonic else { return false }
         let privateKey = derivePrivateKeyOfAccount0(fromHdWallet: walletWhenImported)
         let address = AlphaWallet.Address(fromPrivateKey: privateKey)
--- a/AlphaWallet/UI/Form/MnemonicLengthRule.swift
+++ b/AlphaWallet/UI/Form/MnemonicLengthRule.swift
@@ -6,7 +6,13 @@ import WalletCore
 
 struct MnemonicLengthRule<T: Equatable>: RuleType {
     public init() {
-        self.validationError = ValidationError(msg: R.string.localizable.importWalletImportInvalidMnemonicCount())
+        var msg: String
+        if Features.is24SeedWordPhraseAllowed {
+            msg = R.string.localizable.importWalletImportInvalidMnemonicCount24()
+        } else {
+            msg = R.string.localizable.importWalletImportInvalidMnemonicCount12()
+        }
+        self.validationError = ValidationError(msg: msg)
     }
 
     public var id: String?
@@ -15,8 +21,15 @@ struct MnemonicLengthRule<T: Equatable>: RuleType {
     public func isValid(value: T?) -> ValidationError? {
         if let str = value as? String {
             let words = str.trimmed.split(separator: " ")
-            return (words.count != HDWallet.mnemonicWordCount) ? validationError : nil
+            return !isValidSeedPhraseLength(count: words.count) ? validationError : nil
         }
         return value != nil ? nil : validationError
+    }
+
+    private func isValidSeedPhraseLength(count: Int) -> Bool {
+        if Features.is24SeedWordPhraseAllowed {
+            return HDWallet.validSeedPhraseCounts.contains(count)
+        }
+        return count == HDWallet.SeedPhraseCount.word12.count
     }
 }
--- a/AlphaWallet/Core/Features.swift
+++ b/AlphaWallet/Core/Features.swift
@@ -26,4 +26,5 @@ enum Features {
     static let isAttachingLogFilesToSupportEmailEnabled = false
     static let isPalmEnabled = true
     static let isExportJsonKeystoreEnabled = true
+    static let is24SeedWordPhraseAllowed = false
 }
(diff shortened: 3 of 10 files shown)


hboon:
@eviltofu leave this with me first. Looks like the thing that's broken is generating the seed from a 24-word seed phrase.

hboon:
OK, got it working, I'll need to test it more, clean it up and need to create a PR before this and then you can rebase this current PR. It touches the key management part of the code base. I'll try to finish this before I go, but definitely wouldn't create the PR until I'm back. Want to be sure.

So stand by.

hboon:
@eviltofu I have created PR #3602 which adds backend support for 24-word seed phrases and rebased a copy of this PR's branch to include it: https://github.com/AlphaWallet/alpha-wallet-ios/tree/24_Word_Seed_Phrase_Feature-with-backend-support

See if it works for you? i.e. you can do something like:

1. Check if the branch `24_Word_Seed_Phrase_Feature-with-backend-support` works for you
2. Rename `24_Word_Seed_Phrase_Feature-with-backend-support` to `24_Word_Seed_Phrase_Feature` for this PR
3. Tweak it
4. Be ready for merge after #3602 is merged

hboon:
Post merge:

- [ ] For @hboon to rename vars etc. It's entropy (from seed phrase), not seed.

eviltofu:
Rebased this into `24_Word_Seed_Phrase_Feature`. Imported a 24 word seed phrase and 12 word seed phrase wallets.

eviltofu:
Merged `24_Word_Seed_Phrase_Feature-with-backend-support` branch into my local `24_Word_Seed_Phrase_Feature` branch. Imported a 12/24 word seed phrase. Not forced pushed. Should Features.is24SeedWordPhraseAllowed be set to true or false for the forced push later?

hboon:
> Should Features.is24SeedWordPhraseAllowed be set to true or false for the forced push later?

`false`

> Merged 24_Word_Seed_Phrase_Feature-with-backend-support branch into my local 24_Word_Seed_Phrase_Feature branch. Imported a 12/24 word seed phrase. Not forced pushed.

You mentioned force pushing, so I think it's just semantics: "Merged" implements `git merge` So `24_Word_Seed_Phrase_Feature` now contains the original 4 commits, as well as the 2 commits from `24_Word_Seed_Phrase_Feature-with-backend-support` and a new merge commit. If this is what you meant, that's not desirable. We don't want any merge commits in feature branches and only the 1 merge commit created by GitHub when merging a PR into `master` (but I've always wondered if we should skip that merge commit too and thus all merge commits).

hboon:
#3602 merged. Feel free to go ahead with this PR after rebasing master.

eviltofu:
I've done as instructed by https://github.com/AlphaWallet/alpha-wallet-ios/pull/3579#issuecomment-1002432308 and force pushed the result. Tested with 12 and 24 word seed phrases as well as their QR code counterparts. @hboon

eviltofu:
```
"wallets.showSeedPhrase.subtitle" = "Making backup is very simple and safe:\n just write down 12 words and\n keep them in a secret place, offline.";
```
Needs to be changed. What should the new phrase be? 
```
"Making backup is very simple and safe:\n just write down these 12/24 words and\n keep them in a secret place, offline."
```
or 
```
"Making backup is very simple and safe:\n just write down the following words and\n keep them in a secret place, offline."
```

hboon:
The former. But careful that it's used in 2 places, the 2nd version doesn't fit both cases.

I thought of having 2 (12 and 24) versions, but there's 1 place where we don't know how many words are in the seed phrase. Let's just keep it simple.

AlphaWallet/KeyManagement/EtherKeystore.swift
```diff
@@ -344,7 +344,10 @@ open class EtherKeystore: NSObject, Keystore {
         let seed = HDWallet.computeSeedWithChecksum(fromSeedPhrase: mnemonic)
         guard let walletWhenImported = HDWallet(entropy: wallet.entropy, passphrase: emptyPassphrase) else { return false }
         //If seed phrase has a typo, the typo will be dropped and "abandon" added as the first word, deriving a different mnemonic silently. We don't want that to happen!
-        guard walletWhenImported.mnemonic == mnemonic else { return false }
+
+        // The wallet created via the above seed is not valid. It generates a mnemonic of "" so the next line of code always fails.
+        // FIXME: - add the line below back when issue above is solved
```
hboon, line 349: Is this still true? If so, we should fix it first.

AlphaWallet/UI/Form/MnemonicLengthRule.swift
```diff
@@ -15,8 +16,12 @@ struct MnemonicLengthRule<T: Equatable>: RuleType {
...
         if let str = value as? String {
             let words = str.trimmed.split(separator: " ")
-            return (words.count != HDWallet.mnemonicWordCount) ? validationError : nil
+            return !isValidSeedPhraseLength(count: words.count) ? validationError : nil
         }
         return value != nil ? nil : validationError
     }
+
+    private func isValidSeedPhraseLength(count: Int) -> Bool {
+        return Features.is24SeedWordPhraseAllowed ? count == 12 || count == 24 : count == 12
```
hboon, line 25: We can use `HDWallet.SeedPhraseCount.word12.count` instead of 12/24 and introduce  a `HDWallet.SeedPhraseCount.count` property. Helps to discover this line if we ever add support for other seed phrase of other lengths.

hboon (changes requested):

AlphaWallet/KeyManagement/EtherKeystore.swift
eviltofu: No, I'll remove in next update

AlphaWallet/UI/Form/MnemonicLengthRule.swift
eviltofu: Can we just add an extension function to `HDWallet.SeedPhrase.Count` to return an array of valid lengths? Then just compare the value to the values in the array?

hboon: static func? Sure. The one I suggested makes it impossible to miss a change and your version keeps it localized. Pros and cons to each. But both works since the type is small.

eviltofu:
@hboon Any more changes?

hboon:
Rebased master and force pushed

hboon (approved):
Looks OK. Will test a bit. Thanks!

PR #3584, opened 2021-12-21
WalletConnect v2 #3413

### LINKED ISSUE #3413: WalletConnect v2

oa-s:
Closes #3413


hboon:
@oa-s Do you think this is working sufficiently well and safe to merge, i.e. it doesn't break WalletConnect v1? If it is, can you help to clean it up, rebase and let's merge it.

oa-s:
@hboon is this ok that deployment target version increased to 13?

hboon:
Ok. Let’s do it.

Issue #2918, opened 2021-06-29
Higher level architecture tasks


hboon:
This is on-going work, to keep in mind.

This will translate to lower level work. Just collecting here to keep them in mind as we go along. General goal is to improve the app quality, our development speed as well as support adding more developers working on the same code base.

Logging system (rather than all temporary logs)
---
* Possible to debug user problems remotely
* Useful during development and bug fixing any way

More tests
---
* Fewer regressions, develop newer/changes features faster
* Higher quality
* Also helps with white label projects, especially after resync

Prettifier
---
* Lower priority
* Less time wasted on styling

SwiftLint
---
* Enforce some good practises
* Less time wasted

Functional core
---
* Easier to reason
* Easier to test
* Easier to optimize
* Hence less bugs

Modules
---
* Clearer boundaries
* Hence better code quality
* Hence scale up team size
* Easier to swap out third party code
* Easier to optimize

Less non-main thread/processes
---
* Easier to reason
* Easier to optimize

A few of these work together when we make a change. I think one immediate one we can proceed with is to break out access to Realm into a Cocoapod. This would remove all dependencies on Realm in other parts of the code base, just passing value objects back and forth and provide an opportunity to improve performance of data access (if/when we need to). E.g. we can also swap Realm for an alternate form of datastore.

@vladyslav-iosdev would you think about this and feel free to add any comments/questions/feedback?

vladyslav-iosdev:
@hboon Logging system (rather than all temporary logs)
it means to use something like Firebase to log all warnings errors and so on?

hboon:
> @hboon Logging system (rather than all temporary logs)
> it means to use something like Firebase to log all warnings errors and so on?

No. I wrote a bit of code to use Papertrail for that currently. Just for discovering what RPC node errors we encounter to map them to more user-friendly versions. But not sure we will want that more generally.

Logging system here refers more to things like OSLog and Lumberjack. So we can enable logging for subsystems or full logging during development and at runtime, maybe offer a switch to capture logs when user reproduce a bug. I use a log of `NSLog("xxx")` and I think you use `print()` but most of the time, I keep adding a deleting the same logs to fix related bugs, so even if we don't use it in production, it should help during development.

vladyslav-iosdev:
ok sound good, yep, it could make our life easier.

vladyslav-iosdev:
i am not clearly understood about:
`Functional core`
 What do u meant here?

[...]

PR #3595, opened 2021-12-24
Allow user to select one or more CustomRPC from a list

### LINKED ISSUE #3464: Let users browse among all possible RPC networks
Problem: users have to enter RPC networks manually, where in most cases it can be automated.

**Changes:**

- We add a second line in the list where we mention ChainID
- We add tabs in "Add Custom RPC Network". One is to Enter it manually, second to browse among all existing networks (we can use chainlist)
- Users should be able to hide any network they want except Ethereum Mainnet, but by default, we keep it as it is

**Prototype:**
https://invis.io/MW1220T36ET3

https://user-images.githubusercontent.com/51817359/143016538-ab62a697-ff92-4d51-8896-46ea2c665f85.mov

eviltofu:
Closes #3464

Allows user to select CustomRPC from a list. The list is read from a zipped json text file. Pushed to allow for feedback.


hboon:
The "Save Network" button is clipped at its bottom

Screenshot 2021-12-28 at 10 54 17 AM

hboon:
The checkmark should be bluish. Refer to the screen before it.

Screenshot 2021-12-28 at 10 56 16 AM

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
```diff
@@ -0,0 +1,115 @@
...
+//  AvailableRpcNetwork.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 20/12/21.
+//
+
+import Foundation
+
+@available(iOS 13.0, *)
+fileprivate let SelectedCompressionAlgorithm = NSData.CompressionAlgorithm.lzma
```
hboon, line 11: Lowercase variable name is more consistent?

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
```diff
@@ -0,0 +1,115 @@
...
+
+fileprivate let ChainsZipFileName = "chains_mini.zip"
```
hboon, line 13: Seems better to use `R.file.chains_miniZip()`?

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
```diff
@@ -0,0 +1,115 @@
...
+    let name: String
+    let chainId: Int
+    let shortName: String
+    let networkId: Int
+    let nativeCurrency: NativeCurrency
+    let rpc: [String]
+    let faucets: [String]
+    let infoURL: String
+
+    var isTestNet: Bool {
```
hboon, line 26: The source `.json` files has a `network` field. Can we use that to tell if it's mainnet/testnet (careful that the value might not be standardize, eg. "mainnet" vs "Mainnet") and then fallback to scanning `name` if we have to?

```
$ ag \"network\" | head
eip155-888.json:4:  "network": "mainnet",
eip155-71393.json:4:  "network": "testnet",
eip155-172.json:4:  "network": "testnet",
eip155-66.json:4:  "network": "mainnet",
eip155-42220.json:6:  "network": "Mainnet",
eip155-31.json:4:  "network": "testnet",
eip155-200625.json:4:  "network": "mainnet",
eip155-365.json:4:  "network": "testnet",
eip155-80001.json:4:    "network": "testnet",
eip155-1313500.json:4:  "network": "mainnet",
```

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
```diff
@@ -0,0 +1,115 @@
...
+        name.lowercased().contains("testnet")
+    }
+    var customRpc: CustomRPC {
+        CustomRPC(chainID: chainId, nativeCryptoTokenName: nativeCurrency.name, chainName: name, symbol: nativeCurrency.symbol, rpcEndpoint: rpc.first ?? "", explorerEndpoint: infoURL, etherscanCompatibleType: .unknown, isTestnet: isTestNet)
+    }
+
+    static func availableServers(fileName: String = ChainsZipFileName) -> [CustomRPC]? {
```
hboon, line 33: For functions like this that don't cause any state changes (basically static func + not change class vars and their function arguments), can you try writing them as like described as "Functional core"  in https://github.com/AlphaWallet/alpha-wallet-ios/issues/2918#issuecomment-870350388 and the few comments after that? It's a little odd and 1 more indentation, but it helps to group them together, filtering out other code that causes state changes.

eviltofu:
> The "Save Network" button is clipped at its bottom
> 
> Screenshot 2021-12-28 at 10 54 17 AM

Can you reproduce this? What steps were taken? I am not seeing this for simulator iPhone 12 and 11 running 15.2, hardware iPhone X (15.2), iPhone 6S Plus (14.8).

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
eviltofu: The mini json file choosen did not have this field. Do you want me to use the other file?

hboon:
> Can you reproduce this? What steps were taken? I am not seeing this for simulator iPhone 12 and 11 running 15.2, hardware iPhone X (15.2), iPhone 6S Plus (14.8).

iPhone 11 + iOS 15 simulator + Xcode 13.0 with dfe68aedb5b7de4ae73016ce197964ac0affc60e. I see it right away by navigating to that screen. Scrolling reveals the "Save Network" button correct, but it seems like there is a white rect below the Save Network button that doesn't scroll with it and hence the initial state. I have "Connect Hardware Keyboard" enabled in the simulator, so if this isn't an issue on the device without a keyboard attached, then it's not too important to fix it.

AlphaWallet/Rpc Network/AvailableRpcNetwork.swift
hboon: Sorry missed this one. Yes

eviltofu:
> > Can you reproduce this? What steps were taken? I am not seeing this for simulator iPhone 12 and 11 running 15.2, hardware iPhone X (15.2), iPhone 6S Plus (14.8).
> 
> iPhone 11 + iOS 15 simulator + Xcode 13.0 with [dfe68ae](https://github.com/AlphaWallet/alpha-wallet-ios/commit/dfe68aedb5b7de4ae73016ce197964ac0affc60e). I see it right away by navigating to that screen. Scrolling reveals the "Save Network" button correct, but it seems like there is a white rect below the Save Network button that doesn't scroll with it and hence the initial state. I have "Connect Hardware Keyboard" enabled in the simulator, so if this isn't an issue on the device without a keyboard attached, then it's not too important to fix it.

I never noticed this because I always have the hardware keyboard disconnected on the simulator. Tested on iOS 14/15 hardware devices. The white rect doesn't appear at the bottom of the screen. I think it's the accessory view that is normally attached on the top of the software keyboard.

eviltofu:
@hboon @colourfreak Any feed back on the current build in this pull request? There is one more bug to fix then I need to change to the other JSON file and it should be ready for testing.

eviltofu:
@hboon Ready for user testing. Currently using the chains.json file in compressed form.

hboon:
@eviltofu Good to review + merge?

eviltofu:
> @eviltofu Good to review + merge?

Yes @hboon Removed the background global dispatch queue.

hboon:
Thanks. Removing the wip tag then.

hboon (changes requested):
- [ ] Can we replace the right navigation `+` button with the `Add Network` button at the bottom for the Browser tab? (refer to the prototype in https://github.com/AlphaWallet/alpha-wallet-ios/issues/3464#issue-1061147324). Looks like the attached below.
- [ ] Filter away the entries that have already been added via "Browse" or bundled in the app. Compare using chain ID
- [x] Trim leading and trailing whitespaces for filtering based on the search box input
- [x] Can you check why these first 2 results show up when searching for "test" on mainnet chains? Attached
- [x] The error message "1 entries failed to add…". How about "Failed to add chains". It's not descriptive enough, but wouldn't be too confusing.

Add Network|Test
-|-
Screenshot 2022-01-05 at 12 51 37 PM|Screenshot 2022-01-05 at 12 37 26 PM

Seems to be working fine otherwise.

eviltofu:
> * [ ]  Can we replace the right navigation `+` button with the `Add Network` button at the bottom for the Browser tab? (refer to the prototype in [#3464 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/issues/3464#issue-1061147324)). Looks like the attached below.

The [prototype](https://projects.invisionapp.com/share/MW1220T36ET3#/screens) has the "Add Network" button appear from the bottom when a valid selection condition is present. Can the button be there but disabled until the user selects a network?

> * [ ]  Can you check why these first 2 results show up when searching for "test" on mainnet chains? Attached
```
{
        "name": "Tao Network",
        "chain": "TAO",
        "network": "core",
        "rpc": [
            "https://rpc.testnet.tao.network",
            "http://rpc.testnet.tao.network:8545",
            "https://rpc.tao.network",
            "wss://rpc.tao.network"
        ],
        "faucets": [],
        "nativeCurrency": {
            "name": "Tao",
            "symbol": "TAO",
            "decimals": 18
        },
        "infoURL": "https://tao.network",
        "shortName": "tao",
        "chainId": 558,
        "networkId": 558
    },
```
The rpc array has test in their urls. 
```
{
        "name": "Moonbase Alpha",
        "chain": "MOON",
        "network": "moonbase",
        "rpc": [
            "https://rpc.api.moonbase.moonbeam.network",
            "wss://wss.api.moonbase.moonbeam.network"
        ],
        "faucets": [],
        "nativeCurrency": {
            "name": "Dev",
            "symbol": "DEV",
            "decimals": 18
        },
        "infoURL": "https://docs.moonbeam.network/networks/testnet/",
        "shortName": "mbase",
        "chainId": 1287,
        "networkId": 1287,
        "explorers": [
            {
                "name": "moonscan",
                "url": "https://moonbase.moonscan.io/",
                "standard": "none"
            }
        ]
    },
```
The infoURL has test in it.

Do you want me to not search all the URLs?

> * [ ]  The error message "1 entries failed to add…". How about "Failed to add chains". It's not descriptive enough, but wouldn't be too confusing.
> 

I am also currently removing entries which failed to be added or are duplicates. This behaviour is ok?

eviltofu:
> * [ ]  Filter away the entries that have already been added via "Browse" or bundled in the app. Compare using chain ID

```swift
    private func computeRpcList() -> [CustomRPC] {
        let presentSet: Set = Set(RPCServer.availableServers.map { $0.chainID })
        guard let available: [CustomRPC] = RpcNetwork.functional.availableServers() else {
            return []
        }
        let remaining = available.drop { presentSet.contains($0.chainID) }
        return Array(remaining)
    }
```
Are the hidden servers stored somewhere else?

hboon:
By hidden, do you mean the servers that have already been added or bundled? Just re-compute them every time?

eviltofu:
Editing entry doesn't seem to work.

eviltofu:
> By hidden, do you mean the servers that have already been added or bundled? Just re-compute them every time?

They are being recomputed every time the user goes into the Add CustomRPC from a list screen. I'll have to go look into how the app hides them on the Wallet screen and on the enabled network screen. 
Screen Shot 2022-01-06 at 8 16 22 AM copy 2 Screen Shot 2022-01-06 at 8 16 38 AM copy 2

eviltofu:
Fixed the visibility bug when editing.

hboon:
@eviltofu

> The prototype has the "Add Network" button appear from the bottom when a valid selection condition is present. Can the button be there but disabled until the user selects a network?

Sure.

> The infoURL has test in it.
> Do you want me to not search all the URLs?

Ok, let's ignore infoURL

> I am also currently removing entries which failed to be added or are duplicates. This behaviour is ok?

Can you elaborate? Especially the part about duplicates

If you mean user choosing to add 5, add 3 were successful and 2 failed, that we would discard the 3 too; then yes, that's what we want.

If you mean duplicates as in the entire list of chains from the `chains` repo has chains with duplicate chain IDs, then yes, just silently remove those. And would be good to create a PR to the `chains` repo to fix it

> They are being recomputed every time the user goes into the Add CustomRPC from a list screen. I'll have to go look into how the app hides them on the Wallet screen and on the enabled network screen.

The app doesn't hide it in that sense. Only enabled (i.e with the tick beside them) chains in the Enabled Networks screen are retrieved and displayed in the Wallet tab. Chains that are in that screen but not enabled doesn't get used by the app except showing up in Enabled Networks screen.

> For entries which fail to be added (error or duplicate), should I quarantine these entries for 24 hours? The chains.json file appear to have entries which might be outdated due to inertia; no one checks on them or our check fails due to something we don't expect. Each time the app restarts, these stale entries get put back into the list.

Just leave it for now. Simpler.

Would you log them with something like this:

```
info("[Custom chains] Failed to add xxx field: \(value) name: \(someValue)")
```

Then we might sweep later to log them remotely or something so we can fix those with PRs upstream to `chains`

eviltofu:
When I add from `chains`, I discard any that results in error as well as those which are successfully added. Duplicates I don't think can exist in the code right now but I added code to check for duplicates just in case. So when I add 5, 3 succeeds and 2 fails, I discard all 5. Leaving the failures behind just makes the user try to add them again. Since I cannot cancel the adding process this will force the user to wait until the network time out is exceeded for each failure if they choose to add again. If the user cancels before the adding process is complete, those remaining are not removed.

hboon:
Ok. Sounds good

hboon:
Oh, sorry, thought it's WIP. Would you buzz me and remove the WIP tag if it's good to go?

eviltofu:
> Oh, sorry, thought it's WIP. Would you buzz me and remove the WIP tag if it's good to go?

Only the info logging for failed entries is left. Is there a way we can send this to analytics?

hboon:
We have analytics (Mixpanel), but let's not send it there. Doesn't quite fit. There's a set up for Papertrail, but not sure if we want to continue with it yet.

eviltofu:
@hboon ok ready for review

eviltofu:
Waiting for #3648 before pushing up version with conflicts resolved.

hboon (approved):
Merge when you are ready

eviltofu:
@hboon ready

hboon:
@eviltofu go ahead and merge it.

eviltofu:
> @eviltofu go ahead and merge it.

@hboon So it's a merge commit, squash and merge, or rebase and merge?

hboon:
Always use "Create a merge commit" (for now), which is the default when you click "Merge pull request".

hboon:
And before you merge your own PRs especially, if `master` has had new commits since, it'll be good to rebase master first especially if the PR includes more than a few lines.

PR #3598, opened 2021-12-27
Update WebImageView with right width scale  #3597

### LINKED ISSUE #3597: Adjust Web ImageView content
Its need properly adjust large and small web images. The current implementation doesn't handle it, large images are displayed mostly ok, but small images stretched on all container that causes quality loosing.

oa-s:
Closes #3597, Closes #3632


hboon (approved):

hboon:
Whoops. Didn't mean to approve.

Looks like both before and after this PR, the vertical placement (or maybe the height?) of bigger web images are not correct. Might not be in the scope of this PR, but seems realated? Examples taken from my wallet:

Screenshot 2021-12-28 at 10 25 43 AM
Screenshot 2021-12-28 at 10 25 45 AM
Screenshot 2021-12-28 at 10 25 48 AM

oa-s:
@hboon  please check out this Debug version, it has updated with resizing image to center it and fit with containers width if its large

hboon:
I broke `TokenImageView` #3631 causing it to show the chain overlay in the Collectibles tab and after tapping an NFT. You'll see it when you rebase master. I'll try to fix it separately.

Seems OK in the few examples I tried because it's fitting by height. Does it work if the token image is wide and short? I suppose it'll just be truncated at the sides (I think that's OK).

Screenshot 2022-01-05 at 10 46 27 AM

Why is this red rectangle not stretch to fill the height or width though?

Screenshot 2022-01-05 at 10 48 36 AM

oa-s:
@hboon `red rectangle` its actual images frame, we can made it to stretch but in this case small images are going to be in bad quality. Wide images also would be fitted to containers width.

hboon:
@vladyslav-iosdev ok, if this is good, let's get it merged first then.

oa-s:
@hboon Updated PR with scaling token images to fill frame, and remove rounding 
| Before PR     | After PR |
| ------------ | -------- |
|  |  |
|  |  |

oa-s:
@hboon  Please confirm these changes, and if its ok, we could merge it

hboon (approved):

PR #3601, opened 2021-12-28
Fix vulnerabilities by GitHub for json-schema CVE-2021-3918

hboon:


hboon:
@eviltofu help review and merge it

hboon:
@eviltofu bump

eviltofu:
@hboon Where is this code used? The web view on the main screen?

hboon:
@eviltofu It's an npm package that is a dependency mentioned in `yarn.lock`. It's not listed in `package.json` under `devDependencies` nor `dependencies`, it is just a third party dependency used by another dependency [^1]

[Danger](https://danger.systems/js/) is the only reason we use these npm packages for now according to https://github.com/AlphaWallet/alpha-wallet-ios/blob/f08a03ed6e9bef2c3c8f779d155f5724071210d4/package.json#L3 so it doesn't affect the app (production). `dangerfile.js` was where I suggested you add the check (for localization strings maybe?) in. Looks like it's broken #3610

The JavaScript we inject comes from a Cocoapod, not this.

[^1] Sad state of dependencies, especially in the JavaScript ecosystem.

yarn.lock
```diff
@@ -1893,10 +1893,10 @@ json-schema-traverse@^0.4.1:
   resolved "https://registry.yarnpkg.com/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz#69f6a87d9513ab8bb8fe63bdb0979c448e684660"
   integrity sha512-xbbCH5dCYU5T8LcEhhuh7HJ88HXuW3qsI3Y0zOZFKfZEHcpWiHU/Jxzk629Brsab/mMiHQti9wMP+845RPe3Vg==
 
-json-schema@0.2.3:
-  version "0.2.3"
-  resolved "https://registry.yarnpkg.com/json-schema/-/json-schema-0.2.3.tgz#b480c892e59a2f05954ce727bd3f2a4e882f9e13"
-  integrity sha1-tIDIkuWaLwWVTOcnvT8qTogvnhM=
+json-schema@0.2.3, json-schema@^0.4.0:
```
eviltofu, line 1896: Why does it say @0.2.3 as well as @^0.4.0?

hboon:
```
json-schema@0.2.3, json-schema@^0.4.0:
version "0.4.0"
```

This means that dependencies `json-schema@0.2.3`, `json-schema@^0.4.0` will be resolved with "0.4.0".

eviltofu (approved):

PR #3603, opened 2021-12-29
Created adaptor for ScrollableSegmentedControl and replaced Segmented…

### LINKED ISSUE #3553: Make the “Hit test surface area” bigger
How to explain...  note that when you tap in here nothing will happen. This is because we set links area only to the text. But in reality, it should be a way beyond the text. 
Screenshot 2021-12-15 at 13 36 36

Now:
Screenshot 2021-12-15 at 13 40 42

Should be
Screenshot 2021-12-15 at 13 40 18

eviltofu:
Closes #3553

ScrollableSegmentedControl was created and an adaptor was used to fit it into the current code.

https://user-images.githubusercontent.com/1050309/147637518-7216a381-2c24-4809-a8ef-b996eef6b021.MP4


AlphaWallet/ScrollableSegmentedControl/ScrollableSegmentedControl.swift
```diff
@@ -0,0 +1,270 @@
...
+        } else {
+            configureEqualWidthCells()
+        }
+        configureHighlightLine()
+    }
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    // MARK: - Life cycle
+    override func layoutSubviews() {
```
hboon, line 73: @eviltofu Quick comment! can you leave a single linebreak before each func and group of class/struct var/let(s)?

AlphaWallet/ScrollableSegmentedControl/ScrollableSegmentedControl.swift
```diff
@@ -0,0 +1,270 @@
...
+        case selected(Int)
+    }
+    // Only if not scrollable (total cells width + spacing < control width).
+    enum Alignment {
+        case leading
+        case trailing
+        case centered
+        case filled
+    }
+    // MARK: - Properties
```
hboon, line 31: I don't think `MARK`s works in practice.. eg. a few lines later, what if a propery is both "UI Elements" and `Public`? (I know it looks better in Xcode)

And what if someone places a function under the wrong group (based on `MARK`)?

You can leave them in if they really help you though.

AlphaWallet/ScrollableSegmentedControl/ScrollableSegmentedControlAdapter.swift
```diff
@@ -0,0 +1,62 @@
...
+            handleSelection(selection)
+        }
+    }
+    override init(cells: [ScrollableSegmentedControlCell], configuration: ScrollableSegmentedControlConfiguration) {
+        super.init(cells: cells, configuration: configuration)
+    }
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    // func to invoke delegate
```
hboon: Such comments are spurious and too easy to be out of sync with the code

AlphaWallet/ScrollableSegmentedControl/ScrollableSegmentedControl.swift
eviltofu: It is used for quick navigation in Xcode.

hboon (commented):
Seems to work as expected. Is this PR going to replace the segment bars used in other parts of the app like when importing a wallet (not sure if there are more)?

What is the reason between `ScrollableSegmentedControlAdapter` and `ScrollableSegmentedControl`. What does it adapt the latter to?

IMO, the video isn't necessary here. It's helpful, but a screenshot would probably cost you less time and still point out to the reviewer which part of the app is affected.

AlphaWallet/ScrollableSegmentedControl/ScrollableSegmentedControlAdapter.swift
```diff
@@ -0,0 +1,62 @@
...
+//  ScrollingSegmentedControlAdapter.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 28/12/21.
+//
+
+import UIKit
+
+class ScrollableSegmentedControlAdapter: ScrollableSegmentedControl, ReusableTableHeaderViewType {
+    static var reusableIdentifier: String {
```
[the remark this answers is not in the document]
eviltofu: > Seems to work as expected. Is this PR going to replace the segment bars used in other parts of the app like when importing a wallet (not sure if there are more)?
> 
> What is the reason between `ScrollableSegmentedControlAdapter` and `ScrollableSegmentedControl`. What does it adapt the latter to?
> 
> IMO, the video isn't necessary here. It's helpful, but a screenshot would probably cost you less time and still point out to the reviewer which part of the app is affected.

No, this PR only replaces this one instance. The adapter just makes this a drop in replacement for this instance. There are some differences in set up. I wanted to see if this would have minimal disruption so that when we do have to replace it everywhere it would be easier. If you check the TokensViewController file, the only thing that is changed is the type.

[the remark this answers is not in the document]
eviltofu: Removed `static var reusableIdentifier: String`

eviltofu:
@hboon review required.

hboon (approved):
Looks good. Thanks!

PR #3612, opened 2021-12-31
Improve implementation of leftPadStringWithZero() so it's more idiomatic and potentially faster

hboon:


eviltofu (approved):

hboon:
Thanks! I will update the commit message though. It’s not meant to say it’s faster.

hboon:
Force pushed after modify commit message and rebase master

PR #3618, opened 2021-12-31
Make SwiftLint ignore autogenerated file API.swift

hboon:


eviltofu:
@hboon Should `API.swift` be placed in `.gitignore` like the `*.generated.swift` files?

hboon:
@eviltofu might work too. But will have to remove it from the index and make sure it’s always autogenerated for each build. Not sure if it’s done already. A separate change though.

eviltofu:
> @eviltofu will have to remove it from the index and make sure it’s always autogenerated for each build. Not sure if it’s done already. A separate change though.

I checked the build phases. There is a new Apollo phase which explicitly generates the API.swift file.

eviltofu (approved):
Submit a new issue to ignore or remove API.swift from the repository.

PR #3619, opened 2021-12-31
When computing nonce while private transactions are enabled, check against the private JSON-RPC endpoint instead #3574

### LINKED ISSUE #3574: When computing nonce while private transactions are enabled, check against the private JSON-RPC endpoint instead
So that pending transactions sent privately are included

oa-s:
Closes #3574


hboon (approved):
What a 1-liner fix :)

PR #3673, opened 2022-01-10
Fix AccountsViewController memory leaks #3672

### LINKED ISSUE #3672: Fix `AccountsViewController` memory leaks
While closing opened AccountsViewController from setting, it doesn't get deleted

oa-s:
Closes #3672


AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceCoordinator.swift
```diff
@@ -186,7 +186,7 @@ class WalletBalanceCoordinator: NSObject, WalletBalanceCoordinatorType {
     }
 
     private func notifyWalletSummary() {
-        walletSummary.done(on: queue, { [weak self] summary in
+        walletSummary.done(on: .main, { [weak self] summary in
```
hboon, line 189: Doesn't need `.main` since that's the default for PromiseKit, right? But we should reduce our usage of custom queues anyway. Would you modify and then merge?

hboon (approved):

PR #3676, opened 2022-01-10
Change xDAI network name to Gnosis #3638

### LINKED ISSUE #3638: Change xDAI network name to Gnosis
They have recently changed the native name
https://www.xdaichain.com/

oa-s:
Closes #3638


hboon (changes requested):
Hah. Sorry, is it possible to just change the network name (if there are no other functionality changes required) without renaming the files, var names etc? 44 files affected. Or do you foresee that it's better we do the renaming once and for all?

oa-s:
@hboon so we don't need to rename tokens name, symbol, icon, just networks name?

hboon:
> @hboon so we don't need to rename tokens name, symbol, icon, just networks name?

We don't need to rename the variable names (unless there's a good reason to). But we should check if the token name, symbol and icon should be updated. i.e. minimal internal/code changes, but I don't know if the token name/symbol/icon shown to the user is now different.

hboon:
@oa-s still good for review and merge?

oa-s:
> @oa-s still good for review and merge?

yep

hboon (changes requested):
Seems like it's not building.

hboon:
Not building because this:

```
return R.image.xdai()
```

Should be:

```
return R.image.xDai()
```

AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
```diff
@@ -289,7 +289,11 @@ fileprivate struct Ticker: Codable {
                 return false
             }
         } else {
-            return symbol.localizedLowercase == tokenObject.symbol.localizedLowercase && name.localizedLowercase == tokenObject.name.localizedLowercase
+            if tokenObject.server == .xDai {
+                return symbol.localizedLowercase == "xDai".localizedLowercase && name.localizedLowercase == "xDai".localizedLowercase
```
hboon, line 293: Do you think it will be better if we change the body of the `if` to also match "gno"/"GNO"? To support the future where CoinGecko changes the xDai/GNO `symbol` from "xDai" to "GNO"/"gno".

AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -128,6 +129,23 @@ class TokensDataStore: NSObject {
         }
     }
 
+    private func migrateXDaiToGnosis() {
```
hboon, line 132: This should be in `MigrationInitializer.swift`

oa-s:
@hboon bump

hboon:
@oa-s checking. BTW tests seem to fail

hboon (approved):
Looks good. Can you help replace this line too `return R.image.iconsNetworkXdai()` (check for image dimensions) and merge?

hboon:
@oa-s sorry, hold merging for a bit.

Are you sure the token name and symbol should be changed to GNO?

hboon:
@oa-s still says "xDAI" on the xDai bridge. Only the chain name and logo/icon is updated there. If we aren't sure. Let's don't rename the token symbol and name. Just the chain name.

Screenshot 2022-01-14 at 1 04 02 PM
 
i.e. in this screenshot:

Screenshot 2022-01-14 at 1 05 33 PM

* Keep "GNOSIS"
* Revert "Gnosis"
* Revert "GNO"

I guess we don't need the migration anymore :)

hboon:
@oa-s Good to merge when you are done.

hboon:
Can you help summarize what needs to be changed and what not to over at https://github.com/AlphaWallet/alpha-wallet-android/issues/2302 ?

oa-s:
> Can you help summarize what needs to be changed and what not to over at AlphaWallet/alpha-wallet-android#2302 ?

in case if we don't need to change token, we change:
- chain name;
- chain icon.

oa-s:
> @oa-s Good to merge when you are done.

@hboon ready for review

hboon (approved):

PR #3680, opened 2022-01-10
Added code to display selected language script

### LINKED ISSUE #3651: Localized strings not appearing when set to Simplified Chinese
Verify if simulator only.

1. Change language in Settings.app to Simplified Chinese
2. Launch AlphaWallet app
3. Activity tab

Dates are in Chinese but "Wallet" tab title is still in English. Our localization effort is somewhat patchy and incomplete so not all English strings are localized, but it seems like none of the localized ones are used.

Let's figure out why this is broken before fixing it. It might well be good to stay this way until we patch up the missing localization strings.

Screenshot 2022-01-07 at 10 32 28 AM

eviltofu:
Closes #3651

### DIFF
--- a/AlphaWallet/Accounts/Coordinators/AccountsCoordinator.swift
+++ b/AlphaWallet/Accounts/Coordinators/AccountsCoordinator.swift
@@ -34,9 +34,9 @@ struct AccountsCoordinatorViewModel {
         var navigationTitle: String {
             switch self {
             case .changeWallets:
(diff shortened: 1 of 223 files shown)


hboon (commented):
Check this first: https://github.com/AlphaWallet/alpha-wallet-ios/issues/3651#issuecomment-1009583771

hboon:
@eviltofu can you help close this if it's not applicable anymore and refer to the PR that supersedes this?

PR #3690, opened 2022-01-11
Empty new features view #3684

### LINKED ISSUE #3684: Empty new features view
name: Bug Report
about: Using AlphaWallet, but it's not working as you expect?

---

**Describe the bug**
Empty features view

**Steps to reproduce (REQUIRED)**
Steps to reproduce the behavior

1. uninstall the AlphaWallet app
2. install the AlphaWallet app on iPhone 8 Plus simulator with iOS 15

**Expect**
6a. List of features is displaying

**Observed**
6b. List of features is empty

**Screenshots**
Screenshot 2022-01-11 at 09 38 07

Version number: 3.34 (417)

oa-s:
Closes  #3684


hboon (approved):
@oa-s would you help split out a commit for removing the `JavaScriptKit` pod?

hboon:
So the problem here is that the height constraint isn't set?

hboon:
Merge when you are ready.

oa-s:
> So the problem here is that the height constraint isn't set?

seems like yes, for some reasons it doesn't get updated with its contents height

PR #3702, opened 2022-01-13
Fixed LiveLocaleSwitcherBundle Localization Bug

### LINKED ISSUE #3651: Localized strings not appearing when set to Simplified Chinese

eviltofu:
Closes: #3651 and closes: #3680

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -79,10 +79,6 @@ class AppCoordinator: NSObject, Coordinator {
     }
 
     func start() {
-        if Features.isLanguageSwitcherDisabled {
-            Config.setLocale(.system)
-        }
-
         if isRunningTests() {
             startImpl()
         } else {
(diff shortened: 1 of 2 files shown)


hboon:
Keep it for now if we proceed with this PR.

But I don't understand the reason for this PR. We don't want the in-app language switcher to be enabled.

eviltofu:
@hboon I’ve changed the code to remove the reason why we only see English. Do you want me to go further into Config.setLocale to see if there is anything wrong with that code?

hboon:
Oh yes. Would be good.

Can you also help to modify the PR title and commit message to reflect that this is fixing the bug and not changing the flag?

hboon:
Thanks. I’ll look at it tomorrow. But if you are still around, can you check if we can bypass LiveLocalSwitcherBundle completely if in-app switching is disabled? It shouldn’t be necessary anymore when it is disabled.

eviltofu:
> Thanks. I’ll look at it tomorrow. But if you are still around, can you check if we can bypass LiveLocalSwitcherBundle completely if in-app switching is disabled? It shouldn’t be necessary anymore when it is disabled.

I think we just don't set the Config locale at startup to skip it.

hboon (approved):
Looks good and thanks for hunting this down!

> I think we just don't set the Config locale at startup to skip it.

Can you help remove this and merge?:

```
if Features.isLanguageSwitcherDisabled {
    Config.setLocale(.system)
}
```

It should be safe. Users who have had the language override a long time ago should have already encountered this code already

PR #3703, opened 2022-01-13
App crashes on launch #3699

### LINKED ISSUE #3699: App crashes on launch
The app is crashing on launch due to failure access
```
if let image = icon.value?.image {
        hasher.combine(image.hashValue)
}
```

oa-s:
Closes #3699

### DIFF
--- a/AlphaWallet/Tokens/Types/TokenObject.swift
+++ b/AlphaWallet/Tokens/Types/TokenObject.swift
@@ -96,9 +96,6 @@ extension Activity {
             hasher.combine(contractAddress)
             hasher.combine(symbol)
             hasher.combine(decimals)
-            if let image = icon.value?.image {
-                hasher.combine(image.hashValue)
-            }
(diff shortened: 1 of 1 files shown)


eviltofu:
How do I test this?

oa-s:
actually can't answer as this issue is not reproducible, but the reason of its appearence was in thread accessing while trying to get image  for token object. When calculation hash value for it `TokenObject`, so we can suppose to remove image from calculation its hash

eviltofu:
Do you remember which part is causing the crash? Is there a way to mitigate it?

hboon:
> The code is suppose to do something?

This particular instance seems ok. It's slow and unnecessary to hash the image (I think it is wrong too. The subscribable's value can change at least between "no image yet" and "have image loaded", but the `TokenObject` is actually the same no matter what that value is)

eviltofu (approved):

PR #3720, opened 2022-01-14
After a crash during unit testing, the app cannot start up #3669

### LINKED ISSUE #3669: After a crash during unit testing, the app cannot start up
On hardware, after the app crashes during unit testing, the app cannot start up.

1. Delete the app.
2. Install a new copy and run unit tests from Xcode.
3. App crashes (see #3662).
4. Restart app.
5. App is frozen at start up screen.

oa-s:
Closes #3669


hboon:
@oa-s bit delicate. Could you describe the cause of the problem and how the PR fixes it?

eviltofu:
Does this create a separate Realm folder for unit tests?

oa-s:
> @oa-s bit delicate. Could you describe the cause of the problem and how the PR fixes it?

The issue apears by next cause: after app launch we don't have created wallet, and we run tests, realm files created during tests stay in the document folder, and don't get deleted. On the next launch we can pass througn check 
```MigrationInitializer.hasRealmDatabasesForWallet && !keystore.hasWallets && !isRunningTests()```
as we got db files and dont have addreses in keychain or user default.

including the next one
https://github.com/AlphaWallet/alpha-wallet-ios/issues/3669#issuecomment-1014208789

**Comment by hboon** [comment]
> The issue apears by next cause: after app launch we don't have created wallet, and we run tests, realm files created during tests stay in the document folder, and don't get deleted.

Does it mean that if the user does this (before this PR):

1. Install app
2. Launch app
3. Kill app
4. Launch app

App will not start up?

**Comment by oa-s** [comment]
@hboon  a bit different:
1. Install app
2. Launch app for tests (tests've passed or app has crashed)
3. Launch app
4. App stucks on launch screen

**Comment by hboon** [comment]
@oa-s so it could only happen if the test suite is run? It will not happen for normal users (app store)

**Comment by oa-s** [comment]
> @oa-s so it could only happen if the test suite is run? It will not happen for normal users (app store)

right

File: AlphaWallet/Core/Types/RealmConfiguration.swift [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        var directory = try createSubDirectoryIfNotExists(name: walletsFolderForTests)
+        directory.appendPathComponent(fileName)
+
+        return directory.path
+    }
+
+    static func removeWalletsFolderForTests(name: String = walletsFolderForTests) throws {
+        let fileManager = FileManager.default
```
hboon, line 46: Can you add as the first line of this function to guard for `isRunningTests()` just to be defensive?

AlphaWallet/Core/Types/RealmConfiguration.swift
```diff
@@ -0,0 +1,60 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import RealmSwift
+
+struct RealmConfiguration {
+    private static let walletsFolderForTests = "wallets"
```
hboon, line 7: Can you change the value of the string to include "test"? Maybe "testSuiteWallets"?

AlphaWallet/Core/Initializers/MigrationInitializer.swift
```diff
@@ -92,8 +92,8 @@ extension MigrationInitializer {
 
     //We use the existence of realm databases as a heuristic to determine if there are wallets (including watched ones)
     static var hasRealmDatabasesForWallet: Bool {
-        let documentsDirectory = URL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0])
```
hboon, line 95: How do we figure out that `RealmConfiguration.defaultRealmFolderUrl` returns the same value as the old one (`URL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0])`) ?

hboon: Especially when not running tests.

hboon: Ok

hboon (approved):
@oa-s merge when you are ready

PR #3721, opened 2022-01-15
Added code to display EmptyTableView when there are no items or the search filter did not turn up anything

### LINKED ISSUE #3566: Empty Assets and Collectibles Tabs on Wallet screen needs to be filled.
There should be a screen to explain that there is nothing here and this is not a bug?

Simulator Screen Shot - iPhone 12 - 2021-12-16 at 16 30 42 copy
Simulator Screen Shot - iPhone 12 - 2021-12-16 at 16 30 48 copy

eviltofu:
Closes #3566

When there are no assets or collectibles, the `EmptyTableView` will be displayed with the corresponding message.
When searching tokens and there are no tokens matching the search criteria, the `EmptyTableView` will be displayed with the corresponding message. The `EmptyTableView` is currently added to the `tableView`.

There is however another `EmptyView` which is part of `StatefulViewController` but I have no idea how to change and trigger it.

### DIFF
--- a/AlphaWallet/Settings/Views/EmptyTableView.swift
+++ b/AlphaWallet/Settings/Views/EmptyTableView.swift
@@ -10,7 +10,6 @@ import UIKit
 class EmptyTableView: UIView {
 
     private var imageView: UIImageView?
-    private let title: String
     private let image: UIImage
     private let heightAdjustment: CGFloat
     private lazy var label: UILabel = {
@@ -22,15 +21,21 @@ class EmptyTableView: UIView {
         return label
     }()
 
+    lazy var title: String = "" {
+        didSet {
+            label.text = title
+        }
+    }
+
     convenience init(title: String, image: UIImage, heightAdjustment: CGFloat) {
         self.init(frame: .zero, title: title, image: image, heightAdjustment: heightAdjustment)
     }
 
     init(frame: CGRect, title: String, image: UIImage, heightAdjustment: CGFloat) {
-        self.title = title
         self.image = image
         self.heightAdjustment = heightAdjustment
         super.init(frame: frame)
+        label.text = title
         configureView()
     }
 
--- a/AlphaWallet/Assets.xcassets/activities_empty_list.imageset/Contents.json
+++ b/AlphaWallet/Assets.xcassets/activities_empty_list.imageset/Contents.json
@@ -2,20 +2,14 @@
   "images" : [
     {
       "filename" : "activities_empty_list.pdf",
-      "idiom" : "universal",
-      "scale" : "1x"
-    },
-    {
-      "idiom" : "universal",
-      "scale" : "2x"
-    },
-    {
-      "idiom" : "universal",
-      "scale" : "3x"
+      "idiom" : "universal"
     }
   ],
   "info" : {
     "author" : "xcode",
     "version" : 1
+  },
+  "properties" : {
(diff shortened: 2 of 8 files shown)


AlphaWallet/Settings/Views/EmptyTableView.swift
```diff
@@ -38,6 +38,11 @@ class EmptyTableView: UIView {
         return nil
     }
 
+    func set(title: String) {
```
hboon, line 41: Maybe more idomatic to expose `title` as non-private for the property setter?

AlphaWallet/Settings/Views/EmptyTableView.swift
```diff
@@ -38,6 +38,11 @@ class EmptyTableView: UIView {
...
+        self.title = title
+        label.text = title
```
hboon, line 43: Do we need to mark the layout as dirty?

hboon (changes requested):
- [x] The illustration looks a bit too high on shorter screens:

(don't worry about the non-zero balance. I hardcoded the display to show no tokens for testing)

Screenshot 2022-01-15 at 6 24 07 PM

- [x] A bit too low when searching:

Screenshot 2022-01-15 at 6 29 15 PM


- [ ] The text doesn't get removed if we go back to the `All` sub-tab

1. Launch the app with 0 tokens
2. No text
3. Tap Currencies
4. Displays "You don't have any Currencies Yet"
5. Tap All

Expected:

6. No text (or whatever 2 was showing)

Observed:

6. Displays "You don't have any Currencies Yet"

- [ ] A little odd when we pull down the table header as the empty state illustration stays in place, but not too important:

Screenshot 2022-01-15 at 6 32 01 PM

Looks good otherwise. Merge when you are done.

eviltofu:
The EmptyTableView is positioned in the middle of the space from the last row in the tableview to the bottom of the screen. If you want it higher, please tell me and I will move it to 1/3 up from the 1/2.

hboon:
> The EmptyTableView is positioned in the middle of the space from the last row in the tableview to the bottom of the screen. If you want it higher, please tell me and I will move it to 1/3 up from the 1/2.

Yes, higher. At least be enough to reveal the text below the graphic when the keyboard is shown. Special case it for shorter screens if necessary.

eviltofu:
> > The EmptyTableView is positioned in the middle of the space from the last row in the tableview to the bottom of the screen. If you want it higher, please tell me and I will move it to 1/3 up from the 1/2.
> 
> Yes, higher. At least be enough to reveal the text below the graphic when the keyboard is shown. Special case it for shorter screens if necessary.

Ok I've added the EmptyTableView to the tableView itself so when the keyboard appears, the tableView is shortened and the EmptyTableView is positioned in the middle of the space from the last row in the tableView to the bottom of the tableView.

AlphaWallet/Settings/Views/EmptyTableView.swift
eviltofu: I think the label will mark itself dirty.

eviltofu:
So now there is a problem when you swipe to delete a token if you hide the Ethereum, xDai, Polygon Mainnet. The app crashes. So I do need to implement the deletion algorithm in #3547.

eviltofu:
@hboon ready for review. Review/merge #3547 after merging this.

hboon (approved):

PR #3732, opened 2022-01-17
Move `tokenActionsService` initialization to `AppCoordinator` #3731

### LINKED ISSUE #3731: Move `tokenActionsService` initialization to `AppCoordinator`
the instance of inCoordinator could be created several times, (each time we change wallet) due to this we make some requests for some services in `tokenActionsService`, we can skip them and perform these calls only once.

oa-s:
Closes #3731


hboon (approved):
Would you help fix the test and merge when you are done?

PR #3737, opened 2022-01-17
Make Travis only run on iOS 15, dropping 14.5 temporarily for speed

hboon:


hboon:
Have switched plans so Travis can run builds concurrently. But will only take effect in 1 Feb. Limiting to not clog the pipeline

PR #3742, opened 2022-01-18
Wallet fiat total don't get refreshed after enabling/disabling chain until app is restarted #3738

### LINKED ISSUE #3738: Wallet fiat total don't get refreshed after enabling/disabling chain until app is restarted
If we have $E amount of crypto in chain Ethereum mainet and $G in Gnosis:

1. Make sure app only has Ethereum mainnet and Gnosis enabled
2. Kill app
3. Launch app
4. Verify fiat total shown in Wallet tab header is E + G
5. Disable Ethereum mainnet so only Gnosis is active

Expect:

6. Total = G

Observed:

6. Total = E + G (until restart app)

eg. "Fiat total" in screenshot is "$2.63":

Screenshot 2022-01-18 at 12 32 54 AM

oa-s:
Closes #3738

### DIFF
--- a/AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
+++ b/AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
@@ -122,6 +122,13 @@ class WalletBalanceFetcher: NSObject, WalletBalanceFetcherType {
         for each in deletedServers {
             services.remove(at: each)
         }
+
+        queue.async { [weak self] in 
+            guard let strongSelf = self else { return }
+
+            strongSelf.notifyUpdateBalance()
+            strongSelf.notifyUpdateTokenBalancesSubscribers()
+        }
     }
 
     private func notifyUpdateTokenBalancesSubscribers() {


AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
```diff
@@ -122,6 +122,13 @@ class WalletBalanceFetcher: NSObject, WalletBalanceFetcherType {
         for each in deletedServers {
             services.remove(at: each)
         }
+
+        queue.async { [weak self] in
```
hboon, line 126: Is the `queue.async` necessary?

oa-s: added it as the rest of calls of this method called async.

oa-s:
> There is a lot of jumping from the custom dispatch queue back to the main queue and back again. It's very hard to follow. Is there a diagram or something to help understand the code?

nope, we don't have such kind of diagram, i will recheck this issue later

hboon (approved):

PR #3758, opened 2022-01-20
Refactor Enjin OpenSea providers #3757

### LINKED ISSUE #3757: Refactor Enjin OpenSea providers

oa-s:
Closes #3757

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -331,7 +331,6 @@ extension AppCoordinator: InitialWalletCreationCoordinatorDelegate {
 extension AppCoordinator: InCoordinatorDelegate {
 
     func didRestart(in coordinator: InCoordinator, reason: RestartReason, wallet: Wallet) {
-        OpenSea.resetInstances()
         disconnectWalletConnectSessionsSelectively(for: reason, walletConnectCoordinator: coordinator.walletConnectCoordinator)
 
         keystore.recentlyUsedWallet = wallet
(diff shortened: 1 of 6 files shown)


AlphaWallet/Core/Enjin/EnjinUserManagementInterceptor.swift
```diff
@@ -213,7 +213,7 @@ class UserManagementInterceptor: ApolloInterceptor {
                 } else if promise.isPending {
```
hboon, line 213: It wasn't modified in the PR, but seems can we just store the promise and use `promise.done` multiple times, once for each "pending" block instead of manually maintaining a `Self.pending`?

AlphaWallet/AppCoordinator.swift
```diff
@@ -331,7 +331,6 @@ extension AppCoordinator: InitialWalletCreationCoordinatorDelegate {
 extension AppCoordinator: InCoordinatorDelegate {
 
     func didRestart(in coordinator: InCoordinator, reason: RestartReason, wallet: Wallet) {
-        OpenSea.resetInstances()
```
hboon, line 334: Why is this call no longer needed?

AlphaWallet/Core/Enjin/EnjinProvider.swift
```diff
@@ -13,25 +13,25 @@ struct EnjinError: Error {
     var localizedDescription: String
 }
 
-class EnjinProvider {
```
hboon, line 16: Can remove the one in `AssetDefinitionStoreCoordinator` too

hboon (commented):
This PR is mainly a rename, right?

AlphaWallet/AppCoordinator.swift
oa-s: on wallets switching it resets instances. it was needed when we had had only one active wallet. but for now its non required.

oa-s:
> This PR is mainly a rename, right?

Yep

AlphaWallet/Core/Enjin/EnjinUserManagementInterceptor.swift
oa-s: not clearly understood this

hboon (approved):

PR #3759, opened 2022-01-21
Replaced SegmentedControl with ScrollableSegmentedControl and SegmentedControl.Selection with ControlSelection everywhere else in app. #3641

### LINKED ISSUE #3641: Replace SegmentedControl with ScrollableSegmentedControl
Currently there are 14 instances (according to GitHub search) where `SegmentedControl` is used. Some of these are just using the `SegmentedControl.Selection` type. We should replace these instances of `SegmentedControl` with `ScrollableSegmentedControl`. I also suggest renaming `SegmentedControl.Selection` into something else, perhaps `Control.Selection` or rewriting it into `ControlSelectable<T>`?

eviltofu:
Closes #3641 

@oa-s @hboon Can help test to see if I missed anything? Nothing looks out of place?


hboon (approved):
Didn't check extensively, seems OK, except when importing wallet, tapping "Watch Wallet" should go to the reveal the "Watch" sub-tab, not "Seed" (which is shown if "Import Wallet" is shown).

Screenshot 2022-01-21 at 1 19 23 PM

Screenshot 2022-01-21 at 1 20 40 PM

Merge when you are ready?

PR #3794, opened 2022-01-24
Remove green sent/received icons from the Activity #3770

### LINKED ISSUE #3770: Remove green sent/received icons from the Activity
At this point, I don't think they make any sense. 
- too similar to each other
- look very bad with a network symbol on the other side of a token symbol
- look cluttered

I know that there are many issues related to Activities in https://github.com/AlphaWallet/alpha-wallet-ios/issues/3278, but this is the easy gain.

Screenshot 2022-01-21 at 10 25 54

@hboon This issue needs your special attention. Might not be trivial.

oa-s:
Closes #3770


eviltofu (approved):

eviltofu:
@oa-s merge when ready

PR #3800, opened 2022-01-24
Tweak bottom insets for `Assets` view #3796

### LINKED ISSUE #3796: Could we make list (collection view too) cover all the way to the bottom of the screen with safe area margins to handle the virtual home button?
- [ ] This might be unrelated to this GitHub issue/PR, but could we make list (collection view too) cover all the way to the bottom of the screen with safe area margins to handle the virtual home button?

Screenshot 2022-01-20 at 12 37 53 PM

- [ ] There seems to be many changes that are unrelated to the original issue. Would you help to separate them out into a separate PR?

oa-s:
Closes #3796

### DIFF
--- a/AlphaWallet/Tokens/Collectibles/ViewModels/AssetsPageViewModel.swift
+++ b/AlphaWallet/Tokens/Collectibles/ViewModels/AssetsPageViewModel.swift
@@ -18,7 +18,7 @@ class AssetsPageViewModel {
     }
 
     var backgroundColor: UIColor {
-        GroupedTable.Color.background
+        Colors.appBackground
     }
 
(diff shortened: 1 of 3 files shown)


hboon (changes requested):
This layout looks right, but can we make the background at the bottom be the same as the cells'? :

Screenshot 2022-01-25 at 12 09 25 PM

This layout doesn't look right. The content should be underneath the virtual home bar instead:

Screenshot 2022-01-25 at 12 09 19 PM

i.e. like this, how the content is underneath the virtual home bar:

Screenshot 2022-01-25 at 12 16 58 PM

hboon (approved):

PR #3840, opened 2022-01-27
Unstoppable Domains name resolution deprecation/changes #3722

### LINKED ISSUE #3722: Unstoppable Domains name resolution deprecation/changes
There are some changes to name resolution or deprecation.

This:

> We wanted to share an update regarding changes we are making to our API integration (https://unstoppabledomains.com/api/v1/). We will be deprecating this API next year in favor of resolution services hosted by our partners: Alchemy and Polygon. We will share details of these resolution services soon.

and this:

> Also, we are dropping support for ENS domain name resolution from our API on Dec 31st.
> If you have been using our APIs for ENS name resolution, we recommend you integrate with ENS directly by following their guide:
> https://docs.ens.domains/dapp-developer-guide/resolving-names

There's also a separate integration doc. Will send you the link privately.

Would you help to check them, summarize what we need to do? And if it's obvious it needs to be done and simple enough, would you make the change?

We'll also need to follow up and add the integration for Android

oa-s:
Closes #3722

### DIFF
--- a/AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
+++ b/AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
@@ -1,0 +1,88 @@
+//
+//  DomainResolutionServiceType.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 27.01.2022.
+//
+
+import Foundation
+import PromiseKit
+
+class DomainResolutionService: DomainResolutionServiceType {
+    let server: RPCServer = .forResolvingEns
+
+    func resolveAddress(string value: String) -> Promise<BlockieAndAddressOrEnsResolution> {
+
+        func resolveBlockieImage(addr: AlphaWallet.Address) -> Promise<BlockieAndAddressOrEnsResolution> {
+            BlockiesGenerator()
+                .promise(address: addr)
+                .map { image -> BlockieAndAddressOrEnsResolution in
+                    return (image, .resolved(.address(addr)))
+                }.recover { _ -> Promise<BlockieAndAddressOrEnsResolution> in
+                    return .value((nil, .resolved(.address(addr))))
+                }
+        }
+
+        let getEnsAddressCoordinator = GetENSAddressCoordinator(server: server)
+        let unstoppableDomainsV1Resolver = UnstoppableDomainsV1Resolver(server: server)
+        let unstoppableDomainsV2Resolver = UnstoppableDomainsV2Resolver(server: server)
+
+        let services: [CachebleAddressResolutionServiceType] = [
+            getEnsAddressCoordinator,
+            unstoppableDomainsV2Resolver,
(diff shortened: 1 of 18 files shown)


AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
```diff
@@ -0,0 +1,88 @@
...
+                }
+        }
+
+        let getEnsAddressCoordinator = GetENSAddressCoordinator(server: server)
+        let unstoppableDomainsV1Resolver = UnstoppableDomainsV1Resolver(server: server)
+        let unstoppableDomainsV2Resolver = UnstoppableDomainsV2Resolver(server: server)
+
+        let services: [CachebleAddressResolutionServiceType] = [
+            getEnsAddressCoordinator,
+            unstoppableDomainsV1Resolver,
```
hboon, line 32: Should v2 be before v1? (I don't know)

hboon (commented):
Any test data we can play with?

Can you help describe what are the v1 and v2 resolvers?
and what's the difference?
Do we need both?

Seems like v1 only supports name -> address resolution and v2 does both ways?

hboon:
- [ ] Create issue for Android

oa-s:
For testing you can use domain: `vladylav.wallet` as well as these two addreses:
- 0xF5FFF32CF83A1A614e15F25Ce55B0c0A6b5F8F2c
- 0x8aad44321a86b170879d7a244c1e8d360c99dda8

oa-s:
> Can you help describe what are the v1 and v2 resolvers?

v2 supports for these domains:
- .zil
- .crypto
- .nft
- .blockchain
- .bitcoin
- .coin
- .wallet
- .888
- .dao
- .x

v1 supports:
- .bit
- BTC
- ETH
- ZIL

oa-s:
> Do we need both?

actually can't answer, but i guess we can leave both

oa-s:
> Seems like v1 only supports name -> address resolution and v2 does both ways?

Right, v2 works with domain and address resolution. That is allowed by api.

AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
oa-s: actually i don't know, we can update the order.

hboon (approved):
Ok. Looks good to merge then. Would you help check if v1 is going away (or soon)? If it is and v2 covers everything v1 does, then we can swap the order around and at some point remove v1 completely. If v2 is a superset, it's at least faster now if we go with v2 first

oa-s:
> Ok. Looks good to merge then. Would you help check if v1 is going away (or soon)? If it is and v2 covers everything v1 does, then we can swap the order around and at some point remove v1 completely. If v2 is a superset, it's at least faster now if we go with v2 first

Updated order of execution v2 then v1.

PR #3849, opened 2022-01-29
Changed input view of SlidableTextField to .numberpad and added coded to allow entry of only numbers or an empty field.

### LINKED ISSUE #3751: Gas price and Gas limit textboxes' keyboard should not allow entering "." in configure transaction screen
`ConfigureTransactionViewController`

Probably should be `.numberPad`? (like `Nonce` in the same screen).

Screenshot 2022-01-20 at 1 34 58 PM

eviltofu:
Fixes #3751


oa-s (approved):

oa-s:
Merge when ready

PR #3857, opened 2022-01-31
Add initial Finnish localisation

pekkanikander:
1-liner summary: Add initial Finnish localisation

Other notes:

As a part of making myself familiar with AlphaWallet iOS code, I went ahead and created initial Finnish localisation. Even though I haven't been able to test all of it, mainly because I haven't yet been able to figure out how to trigger some controllers and their corresponding UI, I think this is probably ready to be merged.  I've reviewed main UI views and they look reasonable.


hboon (changes requested):
Hi @pekkanikander, thank you for your contribution. I don't read Finnish, so I'll trust your translations :)

It all looks good in terms of files and structures. There are 2 lines missing though. Would you help to add those in and force push? Then we should be good to merge. These:

https://github.com/AlphaWallet/alpha-wallet-ios/blob/aaaf0665e853269d04b9e244e3a0e1bc3e9bc7f4/AlphaWallet/Localization/en.lproj/Localizable.strings#L149

https://github.com/AlphaWallet/alpha-wallet-ios/blob/aaaf0665e853269d04b9e244e3a0e1bc3e9bc7f4/AlphaWallet/Localization/en.lproj/Localizable.strings#L787

pekkanikander:
Thank you @hboon   I have now added those two lines, as you requested.

hboon:
Thanks again.

Could you help add this one?

https://github.com/AlphaWallet/alpha-wallet-ios/blob/259f0473ef1d9ae5eff7f7b148abf413dd0c85ec/AlphaWallet/Localization/en.lproj/Localizable.strings#L152

We have both `wallets.showSeedPhrase.subtitle` and `wallets.showSeedPhrase.subtitle2` which is unfortunately confusing.

pekkanikander:
Done.  Sorry for my sloppiness.

hboon (approved):
Looks great. Thank you!

hboon:
@pekkanikander it took a while, but this is now on the app store. Thank you!

PR #3858, opened 2022-01-31
Show more collection-level data from OpenSea #3730

### LINKED ISSUE #3730: Show more collection-level data from OpenSea
We can get this additional data for NFTs available on OpenSea:

```
{
  "stats": {
    "count": 1700,
    "num_owners": 596,
    "floor_price": 4.9
    "total_volume": 9397.148721515618,
    "average_price": 9.159014348455768,

    "total_sales": 1026,
    "total_supply": 1700,
    //…
}
```

Here's the corresponding OpenSea page showing some of these data:

Screenshot 2022-01-17 at 6 04 09 PM

Maybe we can display them in the Info tab:

Screenshot 2022-01-17 at 6 20 47 PM

But perhaps we can highlight the floor price?

I was thinking if we should use the floor price:

* display in the Wallet tab
* include in fiat value of the wallet

What do you think?

---

Notes for implementation:

Doc: https://docs.opensea.io/reference/retrieving-collection-stats

```
//Get tokens owned by user:
curl -H 'X-API-KEY: <key>' 'https://api.opensea.io/api/v1/assets/?owner=0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362&order_by=pk&order_direction=asc&limit=50&offset=0'

//Read slug
"slug" : "blitmap"

//Use slug
curl -H 'X-API-KEY: <key>' 'https://api.opensea.io/api/v1/collection/blitmap/stats'
```

oa-s:
Closes #3730


hboon:
@oa-s anything specific to check? Things like floor price and URLs don't seem to show up yet

oa-s:
it seems like it returns as `zero` for each collection

hboon:
@oa-s see if this helps? https://github.com/AlphaWallet/alpha-wallet-ios/issues/3730#issue-1105671620

oa-s:
> @oa-s see if this helps? [#3730 (comment)](https://github.com/AlphaWallet/alpha-wallet-ios/issues/3730#issue-1105671620)

i will double check, but that is where we are getting it from

oa-s:
The other question: what do we need to show when there is no data for non fungible? It seems like the balance json isn't empty, but it doesn't include right filled fields, seems like they are missing.
Screenshot 2022-02-03 at 08 11 50
Screenshot 2022-02-03 at 08 11 57

hboon:
For now, just hide the headers too. But can you verify with curl if you haven’t that they are really missing? I thought the  first one should have something. Could we be rate limited?

oa-s:
> Could we be rate limited?

yes, it happens from time to time

oa-s:
> But can you verify with curl if you haven’t that they are really missing? I thought the first one should have something. 

i will recheck it

hboon (changes requested):
- [ ] What does the "Metadata" field being "Centralized" mean? What are the possible values?

- [x] Can we display "Counts", "Owners", "Floor price", "Total Volume", "Average Price", "Total Sales", "Total Supply" too?

- [ ] Also the "Links" section — "Website", "Medium", "OpenSea" (Maybe Twitter etc if those are available)

- [x] Rename "Attributes" header to "Properties"

- [ ] For "Attributes", seems the value (middle column) should be truncated instead of wrapped:

- [ ] There's now Finnish localization :)

Wrapped| (should be) Truncated
-|-
Screenshot 2022-02-08 at 11 37 55 AM|Screenshot 2022-02-08 at 11 37 59 AM

oa-s:
Also the "Links" section — "Website", "Medium", "OpenSea" (Maybe Twitter etc if those are available)

We display the next links if they are available:
- Instagram;
- Twitter;
- Discord;
- Telegram;
- OpenSea;
- website.

oa-s:
> What does the "Metadata" field being "Centralized" mean? What are the possible values?

actually i didn't find any matched field in OpenSea response, so hurdcoded it.

hboon:
> > What does the "Metadata" field being "Centralized" mean? What are the possible values?
> 
> actually i didn't find any matched field in OpenSea response, so hurdcoded it.

Haha. Can you help remove it?

hboon:
Oh, I see why some of my comments are invalid. I looked at the token instance of the Info tab. Sorry!

oa-s:
@hboon The PR is ready for review

hboon (approved):
Looks good. Just a bit more:

- [ ] Would you help truncate the Token ID too? (I know it's not introduced in this PR)

- [ ] The "OpenSea Collection" 0x495f947276749ce646f68ac8c248420045cb7b5e should have a floor price of "0.1000 ETH" not "0.0000 ETH" (it shows up correctly in the Info tab, but not in the individual token ID in Assets tab)

Screenshot 2022-02-15 at 10 53 31 AM

JSON:

```
"floor_price" : 0.1,
```

- [ ] The View on Twitter and View on Discord buttons (maybe the rest too? But Visit website seems correct) are pointing (wrongly) to AlphaWallet's :)

- [ ] When properties are truncated, can the ellipsis (…) be displayed? It looks like the width of the UILabel might be wrong here?

Screenshot 2022-02-15 at 11 04 45 AM

It's mostly good, and involves many files, feel free to merge first and fix the above with another PR if it's easier for you.

hboon:
@oa-s and remember [this](https://github.com/AlphaWallet/alpha-wallet-ios/issues/3730#issuecomment-1034803414)

hboon:
@oa-s maybe just make it a 1-liner UILabel and truncate trailing?

oa-s:
@hboon The "floor_price" is specific for collection, and for each token id, it returned as different values for asset and collection. Do we need to update assets `floor_price` with collections `floor_price` value?

hboon:
Yes. It should be pulled from the collection then. It means the current cheapest value (by some definition) for the entire collection.

hboon:
@oa-s don't forget to clean up and merge this

PR #3866, opened 2022-02-04
Tests broken #3865

### LINKED ISSUE #3865: Tests broken

oa-s:
Closes #3865

### DIFF
--- a/AlphaWallet/Foundation/Formatter.swift
+++ b/AlphaWallet/Foundation/Formatter.swift
@@ -73,6 +73,7 @@ struct Formatter {
         let formatter = Formatter.default
         formatter.numberStyle = .decimal
         formatter.usesGroupingSeparator = false
+        formatter.locale = Locale(identifier: "en_US")
         return formatter
     }()
 
--- a/AlphaWalletTests/Models/EtherTests.swift
+++ b/AlphaWalletTests/Models/EtherTests.swift
@@ -6,7 +6,8 @@ import XCTest
 import BigInt
 
 class EtherTests: XCTestCase {
-    func testEtherRepresentation() {
+    func testEtherRepresentationEnglishLocale() {
+        Config.setLocale(AppLocale.english)
         let e = Ether(string: "1.2")!
         XCTAssertEqual("\(e)", "1.2")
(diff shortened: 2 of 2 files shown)


hboon (commented):
Good catch. This is very strange. This test passes for me on `master` / 259f0473ef1d9ae5eff7f7b148abf413dd0c85ec. Maybe it's because we have different locales, but I don't understand why.

But more importantly, can you help to check this first:

```
extension NSDecimalNumber {
    var localizedString: String {
        return self.description(withLocale: Config.locale)
    }
}
```

from: https://github.com/AlphaWallet/alpha-wallet-ios/blob/259f0473ef1d9ae5eff7f7b148abf413dd0c85ec/AlphaWallet/UI/Views/AmountTextField.swift#L617

It seems like we should use one of the formatters instead, especially after the recent refactoring and forcing of decimal and thousands separator to always be `.` and `,`. Then maybe this test can would be modified/removed.

oa-s:
For some reasons even there was set value for `Config.locale`, it clears, and `Config.locale` appears with `<current locale>`. For example `PL` it its selected on device.

AlphaWallet/Foundation/Formatter.swift
```diff
@@ -66,6 +66,7 @@ struct Formatter {
 
     static let `default`: NumberFormatter = {
         let formatter = NumberFormatter()
+        formatter.locale = Locale(identifier: "en_US")
```
hboon, line 69: Maybe we shouldn't set it here and use one of the other formatters instead? `.shortCrypto` maybe?

oa-s: i wasn't sure in applying Locale for `default` NumberFormatter, updated it

hboon: I mean, can we not add this line at all?

> formatter.locale = Locale(identifier: "en_US")

since `basicNumberFormatter()` already add set it.

oa-s:
@hboon updated `scientificAmount` formatter with setting localization manually
`formatter.locale = Locale(identifier: "en_US")`

hboon (approved):

PR #3886, opened 2022-02-08
Update CoinTickersFetcher with removing creating AlphaWallet.Address for Ticker as its expensive operation #3881

### LINKED ISSUE #3881: Update CoinTickersFetcher with removing creating AlphaWallet.Address for `Ticker` as its expensive operation

oa-s:
Closes #3881

### DIFF
--- a/AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
+++ b/AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
@@ -263,7 +263,7 @@ fileprivate struct Ticker: Codable {
     let id: String
     let symbol: String
     let name: String
-    let platforms: [String: AlphaWallet.Address]
+    let platforms: [String: String]
 
     func matches(tokenObject: TokenMappedToTicker) -> Bool {
         //We just filter out those that we don't think are supported by the API. One problem this helps to alleviate is in the API output, certain tickers have a non-empty platform yet the platform list might not be complete, eg. Ether on Ethereum mainnet:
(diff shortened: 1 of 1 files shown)


AlphaWallet/Core/CoinTicker/CoinTickersFetcher.swift
```diff
@@ -277,8 +277,26 @@ fileprivate struct Ticker: Codable {
...
+
         guard isServerSupported(tokenObject.server) else { return false }
-        if let (_, contract) = platforms.first(where: { platformMatches($0.key, server: tokenObject.server) }) {
+        if let (_, maybeContractValue) = platforms.first(where: { platformMatches($0.key, server: tokenObject.server) }) {
+            func maybeAddressValue(from str: String) -> AlphaWallet.Address? {
+                let rawValue = str.trimmed
+                if rawValue.isEmpty {
+                    //CoinGecko returns nullAddress as the value (contract) in `platforms` for tokens is sometimes an empty string: `"platforms" : { "ethereum" : "" }`, so we use the 0x0..0 address to represent them
+                    return Constants.nullAddress
+                } else if let value = AlphaWallet.Address(string: rawValue) {
```
hboon, line 289: Why is it faster if we delay `AlphaWallet.Address(string:)` till we do it here? Wouldn't we potentially be doing it more often for the same `rawValue`?

oa-s: its faster as we filter out a lot of objects when do:
`platformMatches($0.key, server: tokenObject.server)` and we don't create not needed addresses

oa-s:
### Code execution time

| Before PR | After PR  |
| -------- | ---------- |
|Decoding takes: **0.7874749898910522** _seconds_|Decoding takes: **26.22158193588257** _seconds_|

oa-s:
Execution time measured with: `CFAbsoluteTimeGetCurrent()` [CFAbsoluteTimeGetCurrent](https://www.hackingwithswift.com/example-code/system/measuring-execution-speed-using-cfabsolutetimegetcurrent)

hboon:
Ok. But maybe the table is the wrong way around? We want after PR to be faster :)

hboon (approved):

PR #3898, opened 2022-02-10
    Integrate support `wallet_switchEthereumChain` and `wallet_addEthereumChain` calls for WalletConnect

### LINKED ISSUE #3887: WalletConnect doesn't work properly on Uniswap and Anyswap (Polygon Network)
When trying to connect to the Polygon version of Uniswap or Anyswap via WC on iOS I get this.

image

Next, I enable Ethereum just to allow WC to connect, but the app is then locked on Ethereum mainnet. Clicking the 'switch to polygon' in the dapp doesn't do anything (press, no change in the network on Uniswap, Prompt - 'switch Metamask to Polygon' on Anyswap.

This has started after the latest update. 

On Android, everything works fine as it does allow you to pick up which chainID you are connecting using WalletConnect, before confirming.

oa-s:
Closes #3887

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -750,6 +750,7 @@
 		8712A39226F476EF0009C376 /* PriceAlertsPageViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8712A39126F476EF0009C376 /* PriceAlertsPageViewModel.swift */; };
 		8712F337265BCAEF0043493B /* WalletSummaryTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8712F336265BCAEF0043493B /* WalletSummaryTableViewCell.swift */; };
 		8717D37327B2947000FB789D /* WalletConnect in Frameworks */ = {isa = PBXBuildFile; productRef = 8717D37227B2947000FB789D /* WalletConnect */; };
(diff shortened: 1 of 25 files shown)


hboon:
Would check. But can you help rename the commit message and PR title? Better to mention the add and switch Wallet JSON/RPC methods instead.

hboon:
So switching in the dapp to a chain that is already enabled works ✅

But when I switching in the dapp to a chain that is supported by the app, but disabled, this happens:

1. Make sure Optimisim is disabled in app
2. Switch in dapp to Optimism
3. Actionsheet to switch appears in app
4. Agree

Expected
---
5. Inform dapp it's successful (because chain is disabled, but supported), app reloads UI, enable Optimism

Observed
---
5. App reloads UI, enable Optimism, but dapp still loaded with previous chain


Is it possible to do it like this? :

> Inform dapp it's successful (because chain is disabled, but supported), app reloads UI, enable Optimism

---

> When we reject a call with 4902 error code in the Dapp nothing happens, there is no error, and no any action. It is possible that we return invalid response and the Dapp doesn't recognise it as error.

This seems to work for me. I get this:

Screenshot 2022-02-11 at 12 56 16 PM

Don't you?

hboon (approved):
Merge when you are satisfied

oa-s:
> This seems to work for me (I get it for Optimistic because I made some code changes to trigger 4902). I get this:

@hboon  Did u push your changes?

hboon:
> > This seems to work for me (I get it for Optimistic because I made some code changes to trigger 4902). I get this:
> 
> @hboon Did u push your changes?

No, I didn't. Just force this line to trigger:

https://github.com/AlphaWallet/alpha-wallet-ios/blob/1cf2e19a658a02d613c68d109f3b1911c06d293d/AlphaWallet/WalletConnect/Coordinator/WalletConnectCoordinator.swift#L323

PR #3928, opened 2022-02-16
Add development flag `Config.development.shouldPretendIsRealWallet` so we can force watch wallets to proceed to sign (and fail), etc

hboon:
Useful especially when we want to force the send buttons, signing actionsheets to appear for watched wallets


oa-s (approved):

hboon:
@oa-s feel free to help to merge in the future :)

PR #3931, opened 2022-02-16
Remove unused IsReadOnlyViewController.isReadyOnly #3929

### LINKED ISSUE #3929: Are implementors and usage of `IsReadOnlyViewController.isReadyOnly` correct?
```
protocol IsReadOnlyViewController: class {
    var isReadOnly: Bool { get set }
}
```

https://github.com/AlphaWallet/alpha-wallet-ios/blob/1f92c7beef64a01fdef480a844da2f7bec466794/AlphaWallet/Tokens/Collectibles/Coordinators/TokensCardCollectionCoordinator.swift#L335

Implementors seem to only implement didSet observer and never read the value.

oa-s:
Closes #3929

### DIFF
--- a/AlphaWallet/Tokens/Collectibles/Coordinators/TokensCardCollectionCoordinator.swift
+++ b/AlphaWallet/Tokens/Collectibles/Coordinators/TokensCardCollectionCoordinator.swift
@@ -76,26 +76,6 @@ class TokensCardCollectionCoordinator: NSObject, Coordinator {
         refreshUponEthereumEventChanges()
     }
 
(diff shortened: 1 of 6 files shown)


hboon (approved):
Not quite sure about this. `makeCoordinatorReadOnlyIfNotSupportedByOpenSeaERC721()` seems to have served a purpose. But doesn't seem to be that bad even if this causes a bug. So let's do it.

hboon:
@oa-s would you help merge after resolving conflicts?

PR #3958, opened 2022-02-17
Remove Promise from `add(events:)` functions #3957

### LINKED ISSUE #3957: Remove Promise from `add(events:)` functions

oa-s:
Closes #3957

### DIFF
--- a/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinatorForActivities.swift
+++ b/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinatorForActivities.swift
@@ -140,18 +140,11 @@ extension EventSourceCoordinatorForActivities.functional {
                         return when(resolved: promises).map(on: queue, { values -> [EventActivityInstance] in
                             values.compactMap { $0.optionalValue }.compactMap { $0 }
                         })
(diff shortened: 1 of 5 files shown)


AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinatorForActivities.swift
```diff
@@ -139,18 +139,11 @@ extension EventSourceCoordinatorForActivities.functional {
...
-                    if events.isEmpty {
-                        return .value(())
-                    } else {
-                        return eventsDataStore.add(events: events, forTokenContract: tokenContract).then(on: queue, { _ -> Promise<Void> in
-                            return .value(())
-                        })
-                    }
-                }).done { _ in
+                }).done(on: .main, { events in
+                    eventsDataStore.add(events: events, forTokenContract: tokenContract)
```
hboon, line 143: The semantics is slightly different from the original. The original returns a promise and only when that is resolved, this promise resolves (with the `return .value(())`). The replacement doesn't. Does the caller depend on this behavior? eg. insert into database and display it right away?

hboon (approved):
Merge when you are ready

PR #3971, opened 2022-02-17
Reduce amount of map functions for promises #3970

### LINKED ISSUE #3970: Reduce amount of map functions for promises

oa-s:
Closes #3970


hboon (approved):
Why would the new version be better?

Issue #3755, opened 2022-01-20
Migrate from Realm


hboon:

oa-s:
sqlite its a good alternative.
in this case we will need to implement all stuff like, fetching, watching changes, and migration by own

eviltofu:
Use a [facade pattern](https://en.wikipedia.org/wiki/Facade_pattern)? Then we can change the back end later? How many types and entries are we talking about?

oa-s:
> How many types and entries are we talking about?

its about most main of instances:
<img width="353" alt="Screenshot 2022-01-20 at 13 53 44" src="https://user-images.githubusercontent.com/96283950/150333948-b97effb4-6ee4-479f-ac83-3350a96915c1.png">

hboon:
> Use a facade pattern?

It'll be quite an abstraction that way, so let's not overdo it. But still, we can be careful not to be too leaky.

> in this case we will need to implement all stuff like, fetching, watching changes, and migration by own

OK, let's say we go with sqlite, to be concrete, we can try something like:

* Create a Cocoapod to do this
* Pass only dumb Swift data types and instructions to update across app/pod
* Optimize for our usage, we have only ~10 types and we are mostly reading and bulk inserting:
    * reading, fetch the entire list of instances for each "type" [^1]
        * so the bigger ones are tokens and transactions [^1]
        * smaller ones like bookmarks etc are easy (need to iron out if we combine bookmarks into 1 set per app instance. Probably easier to just keep it as 1 set per wallet, to reduce scope)
        * we need to be verify if Swift collections (Array and Dictionary) will work or do we need to use NSArray and NSDictionary
    * little or very rudimentary querying
    * writing, bulk writes of multiple instances at one time, by offering:
        * a beginWrite() and commitWrite() (let's give them better names than Realm did) to match Realm and especially sql(ite)
        * be able to write hundreds and thousands of records at a time (download transaction history)
        * provide single function to update a type
        * these should map quite well to sqlite
        * be able to replace entire existing instances based on primary keys/identifiers
        * maybe bulk delete by primary keys/identifiers (forgot if this is needed)
[...]

PR #3973, opened 2022-02-17
Replace Subscribable<Void> with original Realm observation #3972

### LINKED ISSUE #3972: Replace `Subscribable<Void>` with original `Realm observation`

oa-s:
Closes #3972


hboon (commented):
Would this make it harder to implement #3755? Or will it make it unnecessary?

oa-s:
> Would this make it harder to implement #3755? Or will it make it unnecessary?

nope, it won't be harder to implement it, wrapped in `AnyPublisher` so we are still able to handle all Realm events + observe

hboon (approved):
> nope, it won't be harder to implement it, wrapped in `AnyPublisher` so we are still able to handle all Realm events + observe

I mean, if we do something like (D) mentioned in https://github.com/AlphaWallet/alpha-wallet-ios/issues/3755#issue-1108952873, we aren't going to have access to Realm anymore. But let's go ahead with this, who knows when we'll get to it :)

Merge when you are ready

PR #3977, opened 2022-02-18
Replace transaction filtering with NSPredicate #3975

### LINKED ISSUE #3975: Replace transaction filtering with `NSPredicate`
For improving Realm queries we can replace outgoing filtering with predicates to move filter logic to db queries

oa-s:
Closes #3975


hboon (approved):
Looks good. Merge when ready.

Don't forget the `functional` trick :)

PR #3985, opened 2022-02-18
Make CallForAssetAttributeCoordinator, XMLHandler nonThreadSafe #3984

### LINKED ISSUE #3984: Make CallForAssetAttributeCoordinator, XMLHandler nonThreadSafe
`ThreadSafeDictionary` Subscribable for doesn't work well, causes a lot of bugs (crashes), also everything related to CallForAssetAttributeCoordinator, XMLHandler, need to move to main queue, (check execution time)

oa-s:
Closes #3984

### DIFF
--- a/AlphaWallet/Activities/ActivitiesService.swift
+++ b/AlphaWallet/Activities/ActivitiesService.swift
@@ -97,7 +97,7 @@ class ActivitiesService: NSObject, ActivitiesServiceType {
     private lazy var tokenObjectsCache: CachedTokenObjectResolverType = TokenObjectsCache(tokensCollection: tokenCollection)
     //Cache tokens lookup for performance
     private var tokensCache: ThreadSafeDictionary<AlphaWallet.Address, Activity.AssignedToken> = .init()
-
+    private let activitiesThreadSafeQueue = DispatchQueue(label: "ActivitiesSynchronizedAccessQueue", qos: .background)
     private var cancelable = Set<AnyCancellable>()
 
     init(
@@ -157,7 +157,7 @@ class ActivitiesService: NSObject, ActivitiesServiceType {
         if let promise = contractsAndCardsPromise, promise.isPending {
             return
         }
-
+        let enabledServers = self.config.enabledServers
         let promise = firstly {
             //Hack way to keep all token objects for all datasources
             tokenObjectsCache.updateCache().map(on: .main, { tokenObjects -> [TokenObject] in
@@ -202,7 +202,7 @@ class ActivitiesService: NSObject, ActivitiesServiceType {
                     guard let server = xmlHandler.server else { continue }
                     switch server {
                     case .any:
-                        for each in self.config.enabledServers {
+                        for each in enabledServers {
                             contractAndCard.append((tokenContract: eachContract, server: each, card: card, interpolatedFilter: interpolatedFilter))
                         }
                     case .server(let server):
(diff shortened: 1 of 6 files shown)


AlphaWallet/Activities/ActivitiesService.swift
```diff
@@ -242,8 +242,10 @@ class ActivitiesService: NSObject, ActivitiesServiceType {
 
             strongSelf.reloadViewController(reloadImmediately: reloadImmediately)
 
-            for (activity, tokenObject, tokenHolder) in activitiesAndTokens {
-                strongSelf.refreshActivity(tokenObject: tokenObject, tokenHolder: tokenHolder, activity: activity)
+            DispatchQueue.main.async {
+                for (activity, tokenObject, tokenHolder) in activitiesAndTokens {
+                    strongSelf.refreshActivity(tokenObject: tokenObject, tokenHolder: tokenHolder, activity: activity)
```
hboon, line 247: This might potentionally have many refreshes on the main thread though. Does it block anything?

AlphaWallet/Foundation/Subscribable.swift
```diff
@@ -21,15 +21,15 @@ open class Subscribable<T>: Hashable {
     }
 
     private var _value: T?
-    private var _subscribers: ThreadSafeDictionary<SubscribableKey, Subscription> = .init()
+    private var _subscribers: [SubscribableKey: Subscription] = .init()
```
hboon, line 24: Not specifically for this one or this PR. But I think our approach is wrong; we should look greatly reduce the usage of `Subscribable`. It's just too brittle. What do you think?

hboon (approved):

hboon:
Merge when ready

oa-s:
Reasons of updating the function `func tokenIdsForErc721Token(contract: AlphaWallet.Address, forServer server: RPCServer, inAccount account: AlphaWallet.Address) -> Promise<[String]>`:
### First - 
```
let transactions = self.objects
                    .sorted(byKeyPath: "date", ascending: true)
                let operations: [LocalizedOperationObject] = transactions.flatMap { $0.localizedOperations.filter { $0.contractAddress?.sameContract(as: contract) ?? false } }
```
### Second
```
let transactions2 = self.realm.objects(Transaction.self)
                    .filter(TransactionsStorage.functional.transactionPredicate(server: server, operationContract: contract))
                let operations2: [LocalizedOperationObject] = transactions2.flatMap { $0.localizedOperations.filter { $0.contractAddress?.sameContract(as: contract) ?? false } }

```

[XXX] First Took 0.050396084785461426 seconds
[XXX] Second Took 0.0014380216598510742 seconds, 0 == 0
[XXX] First Took 0.05182504653930664 seconds
[XXX] Second Took 0.0013730525970458984 seconds, 0 == 0
[XXX] First Took 0.05002999305725098 seconds
[XXX] Second Took 0.0013420581817626953 seconds, 0 == 0
[XXX] First Took 0.049034953117370605 seconds
[XXX] Second Took 0.0014369487762451172 seconds, 0 == 0
[XXX] First Took 0.04983508586883545 seconds
[XXX] Second Took 0.0013610124588012695 seconds, 0 == 0
[XXX] First Took 0.04944503307342529 seconds
[XXX] Second Took 0.0013920068740844727 seconds, 0 == 0
[XXX] First Took 0.05259990692138672 seconds
[XXX] Second Took 0.001374959945678711 seconds, 0 == 0
[XXX] First Took 0.05085194110870361 seconds
[XXX] Second Took 0.0014400482177734375 seconds, 0 == 0
[XXX] First Took 0.05003798007965088 seconds
[XXX] Second Took 0.0012660026550292969 seconds, 0 == 0
[XXX] First Took 0.050215959548950195 seconds
[XXX] Second Took 0.001277923583984375 seconds, 0 == 0

AlphaWallet/Foundation/Subscribable.swift
oa-s: yep, 100 percent agree, it causes a lot of issues, for large wallet it causes creating a lot of threads for synchronization, so yes, it's fragile too much. Do you suggest using somethig instead? `PassthroughSubject<T, Failure> ``CurrentValueSubject<T, Failure>`

AlphaWallet/Activities/ActivitiesService.swift
oa-s: updated it, `refreshActivity` can be updated on `.background` queue, just locked access to non atomic `activities` where the app was crashing

AlphaWallet/Foundation/Subscribable.swift
hboon: Maybe just trigger it less granularly so ultimately whole screens refresh. Should be easier when we have more dumb objects rather than smarter ones like native Realms. Then we might even do diffs. Screens like UITableViews and UICollectionViews are only access in 1 (or a few, for pre-fetches) pages anyway. It will be way easier to reason. A good guide is how much we can reason about the code, eg. when we have a bug, can we often figure it out without needing to run the app to reproduce it with breakpoints?

PR #4013, opened 2022-03-02
If there's a significant impact to using GoogleContentSize.s750 for all images hosted by Google #3944

### LINKED ISSUE #3944: If there's a significant impact to using GoogleContentSize.s750 for all images hosted by Google
If there's a significant impact to using GoogleContentSize.s750 for all images hosted by Google

Related to: #3909

This:

> Are we rewriting the URL for Google hosted images for all cases? Should we add a case or somehow be able to choose when to use the original. Eg. we only want to use the rewritten, bigger version when in Wallet > Collectibles, and Info, and in the screen showing a single token ID, but not in screens like the little icons in Wallet tab or Activity Tab.

> Concerned with memory usage and speed.

Probably have to try a wallet with some NFTs

oa-s:
Closes #3944

### DIFF
--- a/AlphaWallet/Tokens/Types/TokenObject.swift
+++ b/AlphaWallet/Tokens/Types/TokenObject.swift
@@ -8,19 +8,14 @@ extension Activity {
 
     struct AssignedToken: Equatable, Hashable {
 
-        struct TokenBalance {
-            var balance = "0"
-            var json: String = "{}"
-        }
-
         enum Balance {
             case value(NSDecimalNumber?)
-            case nftBalance(NSDecimalNumber)
+            case nftBalance([NonFungibleFromJson])
 
             var isEmpty: Bool {
                 switch self {
                 case .nftBalance(let values):
-                    return values == 0
+                    return values.isEmpty
                 case .value(let value):
                     return value == 0
                 }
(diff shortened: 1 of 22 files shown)


AlphaWallet/Tokens/Types/TokenObject.swift
```diff
@@ -8,30 +8,34 @@ extension Activity {
 
     struct AssignedToken: Equatable, Hashable {
 
-        struct TokenBalance {
-            var balance = "0"
-            var json: String = "{}"
-        }
-
         enum Balance {
-            case value(NSDecimalNumber?)
```
hboon, line 17: This seems like an unrelated change to the image size changes. Can you help to separate them?

AlphaWallet/Core/Types/WalletBalance.swift
```diff
@@ -28,7 +28,7 @@ struct WalletBalance: Equatable {
     }
 
     var etherAmountShort: String? {
-        guard let token = etherTokenObject, let value = token.valueDecimal else { return nil }
```
hboon, line 31: Could you help to separate these changes out? Looks like they aren't related to image sizes?

PR #4017, opened 2022-03-02
Move the parts of URLTests.swift that test WebImageURL to pod #3942

### LINKED ISSUE #3942: Help move the parts of `URLTests.swift` that test `WebImageURL`
Help move the parts of `URLTests.swift` that test `WebImageURL` into a different folder `AlphaWalletTests/modules/AlphaWalletCore`. It's more correct to move it into the pod, but since we only routinely run our app's tests, lets just keep it in the app for now.

oa-s:
Closes #3942


hboon (commented):
Do you think it's better to move them into a different folder like `AlphaWalletTests/modules/AlphaWalletCore`, that is part of the app's test suite, not the pod's test suite? It's more correct to move it into the pod, but since we only routinely run our app's tests but not the pods'…

oa-s:
got it, np

oa-s:
@hboon The PR is ready for review

hboon (approved):

PR #4025, opened 2022-03-03
Reduce incoordinator size, move Tabs model to separate file #4024

### LINKED ISSUE #4024: Reduce incoordinator size, move Tabs model to separate file

oa-s:
Closes #4024

### DIFF
--- a/AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
+++ b/AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
@@ -54,7 +54,7 @@ class ActivitiesCoordinator: NSObject, Coordinator {
         let viewModel = ActivitiesViewModel()
         let controller = ActivitiesViewController(analyticsCoordinator: analyticsCoordinator, keystore: keystore, wallet: wallet, viewModel: viewModel, sessions: sessions)
         controller.delegate = self
-
+        
         return controller
     }
 
--- a/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
@@ -109,9 +109,8 @@ final class DappBrowserCoordinator: NSObject, Coordinator {
         self.analyticsCoordinator = analyticsCoordinator
 
         super.init()
-
         //Necessary so that some sites don't bleed into (under) navigation bar after we tweak global styles for navigationBars after adding large title support
-        self.navigationController.navigationBar.isTranslucent = false
+        navigationController.navigationBar.isTranslucent = false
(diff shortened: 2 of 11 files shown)


AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
```diff
@@ -109,9 +111,9 @@ final class DappBrowserCoordinator: NSObject, Coordinator {
         self.analyticsCoordinator = analyticsCoordinator
 
         super.init()
-
+//        navigationController.configureForLargeTitles()
```
hboon, line 114: Remember to delete if they aren't used

AlphaWallet/Extensions/UINavigationController.swift
```diff
@@ -3,6 +3,31 @@
 import UIKit
 
 extension UINavigationController {
+
+    static func withOverridenBarAppearence(appearence: UINavigationBarAppearance = .defaultAppearence) -> UINavigationController {
```
hboon, line 7: Typo: Appearance (It was there before the PR though)

AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
```diff
@@ -45,6 +47,7 @@ class ActivitiesCoordinator: NSObject, Coordinator {
         let viewModel = ActivitiesViewModel()
         let controller = ActivitiesViewController(viewModel: viewModel, sessions: sessions)
         controller.delegate = self
+        controller.tabBarItem = UITabBarController.Tabs.activities.tabBarItem
```
hboon, line 50: But assigning to `.tabBarItem` doesn't seem like it should be the individual coordinators' (or view controllers) responsibility, it should be higher up? `InCoordinator` (where it used to be in) or something that `InCoordinator` delegates to for tab configuration/management seems more appropriate?

AlphaWallet/Activities/Coordinators/ActivitiesCoordinator.swift
```diff
@@ -30,6 +30,8 @@ class ActivitiesCoordinator: NSObject, Coordinator {
         self.navigationController = navigationController
         super.init()
 
+        navigationController.configureForLargeTitles()
```
hboon, line 33: Similar for calling `.configureForLargeTitles()` like this:

> But assigning to .tabBarItem doesn't seem like it should be the individual coordinators' (or view controllers) responsibility, it should be higher up? InCoordinator (where it used to be in) or something that InCoordinator delegates to for tab configuration/management seems more appropriate?

hboon (changes requested):

PR #4033, opened 2022-03-05
Add walletsPublisher to walletAddressesStore #4032

### LINKED ISSUE #4032: Add `walletsPublisher` to `walletAddressesStore`

oa-s:
Closes #4032

### DIFF
--- a/AlphaWallet/KeyManagement/JsonWalletAddressesStore.swift
+++ b/AlphaWallet/KeyManagement/JsonWalletAddressesStore.swift
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import Combine
 
 struct JsonWalletAddressesStore: WalletAddressesStore {
     private static let walletsFolderForTests = "testSuiteWalletsForWalletAddresses"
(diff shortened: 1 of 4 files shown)


AlphaWallet/KeyManagement/JsonWalletAddressesStore.swift
```diff
@@ -49,6 +56,8 @@ struct JsonWalletAddressesStore: WalletAddressesStore {
         } else {
             walletAddresses = WalletAddresses()
         }
+        walletsSubject = .init(Set(walletAddresses.wallets))
+        print("[XXX] wallets did init")
```
hboon, line 60: Remember to remove before merging :)

hboon (commented):
There's some refactoring + code to wrap a publisher around the wallets, but the publisher isn't actually used. I suppose there'll be a separate PR for using the publishers? Do you want to split this current PR into 2? — One for the refactoring and one for adding the publisher wrapping code (even if you are going to have a followup, 3rd PR to use the publishers?)

AlphaWallet/KeyManagement/JsonWalletAddressesStore.swift
oa-s: Yep, want to make small PRs

PR #4059, opened 2022-03-10
Remove force unwrap when encodeType(primaryType for `EIP712Domain` and `EIP712Domain` is missing in typed data #4057

### LINKED ISSUE #4057: WalletConnect 2 eth_signTypedData crash
name: Bug Report
about: It crashed, and rerun got a blank screen.

---

**Describe the bug**

**Steps to reproduce (REQUIRED)**
Steps to reproduce the behavior

1. Connect to WalletConnect 2 test Dapp https://react-app.walletconnect.com/
2. Invoke eth_signTypedData
3. App crashed
4. Rerun

**Expect**
App started.

**Observed**
Blank screen with nothing.

**Screenshots**
image

3.36

oa-s:
Closes #4057

### DIFF
--- a/AlphaWallet/WalletConnect/Types/EIP712TypedData.swift
+++ b/AlphaWallet/WalletConnect/Types/EIP712TypedData.swift
@@ -78,8 +78,9 @@ extension EIP712TypedData {
         var depSet = findDependencies(primaryType: primaryType)
         depSet.remove(primaryType)
         let sorted = [primaryType] + Array(depSet).sorted()
-        let encoded = sorted.map { type in
-            let param = types[type]!.map { "\($0.type) \($0.name)" }.joined(separator: ",")
+        let encoded = sorted.compactMap { type in
+            guard let values = types[type] else { return nil }
+            let param = values.map { "\($0.type) \($0.name)" }.joined(separator: ",")
             return "\(type)(\(param))"
         }.joined()
         return encoded.data(using: .utf8) ?? Data()


AlphaWallet/WalletConnect/Types/EIP712TypedData.swift
```diff
@@ -78,8 +78,9 @@ extension EIP712TypedData {
         var depSet = findDependencies(primaryType: primaryType)
         depSet.remove(primaryType)
         let sorted = [primaryType] + Array(depSet).sorted()
-        let encoded = sorted.map { type in
-            let param = types[type]!.map { "\($0.type) \($0.name)" }.joined(separator: ",")
+        let encoded = sorted.compactMap { type in
```
hboon, line 81: Is this because there is an unsupported type that we should support?

oa-s: i am not quite sure that type is not supported, when we calculate `digest` for typedData we check for `EIP712Domain` that is missing for typed data received from Wallet Connect v2, as i know it should present there

hboon: Hmm.. OK, can you help leave a comment there before the compactMap to mention the problem we are avoiding? So when we see the code (and the comment) the next time we know?

hboon (approved):

AlphaWallet/WalletConnect/Types/EIP712TypedData.swift
hboon: Normally I'd rely on `git log -S`, but seems hard to think of searching git log for this :)

PR #4064, opened 2022-03-11
Replace nativeCryptoCurrencyPrices and nativeCryptoCurrencyBalances with fetching balance and rate with balanceCoordinator #4063

### LINKED ISSUE #4063: Replace nativeCryptoCurrencyPrices and nativeCryptoCurrencyBalances with fetching balance and rate with balanceCoordinator

oa-s:
Closes #4063


oa-s:
The PR removes dependency from  serverDictionary of native crypto balances and native prices, and uses source balance coordinator to fetch data

hboon (approved):
Didn't check.

But could you (or anyone) maybe in a future PR since the current one is quite big already, rename `BalanceCoordinator`? It's one of those things that aren't really a coordinator but called one.

PR #4088, opened 2022-03-16
Created VerticalButtonsBar for VerifySeedPhraseViewController, renamed ButtonsBar to HorizontalButtonsBar

### LINKED ISSUE #3819: A proper clear/secondary button in verify seed phrase screen
@colourfreak could we one for that?

Maybe the same (or different) style for the "Import from iCloud…" button too when importing:

Screenshot 2022-01-26 at 11 31 53 AM

eviltofu:
Closes #3819
Also part of #4089
https://user-images.githubusercontent.com/1050309/158499074-b78eaae5-d6e2-4ce4-9edf-6830ef59a2a0.mp4

### DIFF
--- a/AlphaWallet/Core/Views/ButtonsBar/VerticalButtonsBar.swift
+++ b/AlphaWallet/Core/Views/ButtonsBar/VerticalButtonsBar.swift
@@ -1,0 +1,108 @@
+//
+//  VerticalButtonsBar.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 9/3/22.
+//
+
+import UIKit
+
+class VerticalButtonsBar: UIView {
+    
+    // MARK: - Public properties
+    var buttons: [UIButton] = []
+    
+    // MARK: Private properties
+    private var stackView: UIStackView = UIStackView()
+    
+    // MARK: - Init
+    init(numberOfButtons: Int) {
+        super.init(frame: .zero)
+        setup(numberOfButtons: numberOfButtons)
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    // MARK: - Setup and creation
+    private func setup(numberOfButtons: Int) {
+        var buttonViews: [ContainerViewWithShadow<BarButton>] = [ContainerViewWithShadow<BarButton>]()
+        guard numberOfButtons > 0 else { return }
+        let primaryButton = createButton(viewModel: .primaryButton)
+        primaryButton.childView.borderWidth = 0
+        buttonViews.append(primaryButton)
+        guard numberOfButtons > 1 else { return }
+        (1 ..< numberOfButtons).forEach { _ in
+            buttonViews.append(createButton(viewModel: .secondaryButton))
+        }
+        setupStackView(views: buttonViews)
+        setupButtons(views: buttonViews)
+        setupView(numberOfButtons: numberOfButtons)
+    }
+    
+    private func setupStackView(views: [UIView]) {
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.removeAllArrangedSubviews()
+        stackView.addArrangedSubviews(views)
+        stackView.axis = .vertical
+        stackView.distribution = .equalSpacing
+        stackView.spacing = 16.0
+        addSubview(stackView)
+        NSLayoutConstraint.activate([
+            topAnchor.constraint(equalTo: stackView.topAnchor),
+            bottomAnchor.constraint(equalTo: stackView.bottomAnchor),
+            leadingAnchor.constraint(equalTo: stackView.leadingAnchor),
+            trailingAnchor.constraint(equalTo: stackView.trailingAnchor),
+        ])
+    }
+
+    private func setupButtons(views: [ContainerViewWithShadow<BarButton>]) {
+        buttons.removeAll()
+        views.forEach { view in
+            buttons.append(view.childView)
+        }
+    }
+
+    private func setupView(numberOfButtons: Int) {
+        translatesAutoresizingMaskIntoConstraints = false
+    }
+
+    private func createButton(viewModel: ButtonsBarViewModel) -> ContainerViewWithShadow<BarButton> {
+        let button = BarButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
+        let containerView = ContainerViewWithShadow(aroundView: button)
+        containerView.translatesAutoresizingMaskIntoConstraints = false
+        setup(viewModel: viewModel, view: containerView)
+        NSLayoutConstraint.activate([
+            button.heightAnchor.constraint(equalToConstant: 50.0),
+        ])
+        return containerView
+    }
+
+    // TODO: This function duplicates a function in ButtonsBar. Merge these two functions.
+
+    private func setup(viewModel: ButtonsBarViewModel, view: ContainerViewWithShadow<BarButton>) {
+        view.configureShadow(color: viewModel.buttonShadowColor, offset: viewModel.buttonShadowOffset, opacity: viewModel.buttonShadowOpacity, radius: viewModel.buttonShadowRadius, cornerRadius: viewModel.buttonCornerRadius)
+
+        let button = view.childView
+        button.setBackgroundColor(viewModel.buttonBackgroundColor, forState: .normal)
+        button.setBackgroundColor(viewModel.disabledButtonBackgroundColor, forState: .disabled)
+
+        viewModel.highlightedButtonBackgroundColor.flatMap { button.setBackgroundColor($0, forState: .highlighted) }
+        viewModel.highlightedButtonTitleColor.flatMap { button.setTitleColor($0, for: .highlighted) }
+
+        button.setTitleColor(viewModel.buttonTitleColor, for: .normal)
+        button.setTitleColor(viewModel.disabledButtonTitleColor, for: .disabled)
+        button.setBorderColor(viewModel.buttonBorderColor, for: .normal)
+        button.setBorderColor(viewModel.disabledButtonBorderColor, for: .disabled)
+        button.titleLabel?.font = viewModel.buttonFont
+        button.titleLabel?.adjustsFontSizeToFitWidth = true
+        //So long titles (that cause font to be adjusted) have some margins on the left and right
+        button.contentEdgeInsets = .init(top: 0, left: 3, bottom: 0, right: 3)
+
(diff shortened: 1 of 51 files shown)


AlphaWallet/Core/Views/ButtonsBar/VerticalButtonsBar.swift
```diff
@@ -0,0 +1,108 @@
...
+//
+//  Created by Jerome Chan on 9/3/22.
+//
+
+import UIKit
+
+class VerticalButtonsBar: UIView {
+    
+    // MARK: - Public properties
+    var buttons: [UIButton] = []
```
hboon, line 13: private before public

AlphaWallet/Foundation/Repeat.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  Repeat.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 14/3/22.
+//
+
+import Foundation
+
+func repeatTimes(_ times: Int, block: () -> Void) {
```
hboon, line 10: I'm skeptical about adding this. The `for-each` look is perfectly fine. I'd much rather not create new dialects. For some exceptions, it's useful to create some utility functions; in this case, the old version requires an explicit `0` and choice of inclusive/exclusive-ended loops, so maybe it's useful. So let's keep it for now.

hboon (changes requested):
52 files are modified in the PR There's too much noise and takes too long to review.

1 thing before I check: Is it possible to continue to call `ButtonsBar` that and then pass in an argument or configuration which determines if the layout is vertical or horizontal, and defaulting to horizontal? That seems like it would avoid duplicating or having very similar code and affecting so many files? Is there a reason why we don't do that instead?

And assuming that it is necessary and for future similar changes, this approach should help a lot:

* Keep the names `ButtonsBarConfiguration` and `BarButton` and make the changes required, i.e. add `VerticialBarButton`. Make a commit. This will be the core part of the PR to review. Note the commit hash and point out to look at this commit in the PR.
* Have a separate commit to rename `ButtonsBarConfiguration` and `BarButton`. Note that this is different from multiple "WIP" commits pushed, which doesn't help reviewing the changes.
* We have the option to squash the 2 commits if necessary before merge

We ought to strive to get PRs merged as soon as we reasonably can. And one of the ways to do this is to make it quicker and safer to merge changes.

---

I'm getting this in the console:

> User defaults from command line:
>     IDECustomDerivedDataLocation = /Users/hboon/Library/Caches/JetBrains/AppCode2021.2/DerivedData
> 
> Resolve Package Graph
> Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
> Resolved source packages:
>   WalletConnect: https://github.com/WalletConnect/WalletConnectSwiftV2 @ 0.5.0
> 
> xcodebuild: error: Could not resolve package dependencies:
>   Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2

Is it because you have upgraded Xcode and I haven't? Is the upgrade necessary? And anyway this change doesn't look like it should be part of this PR, can you remove it?

eviltofu:
> 52 files are modified in the PR. There's too much noise and takes too long to review.
> 
> 1 thing before I check: Is it possible to continue to call `ButtonsBar` that and then pass in an argument or configuration which determines if the layout is vertical or horizontal, and defaulting to horizontal? That seems like it would avoid duplicating or having very similar code and affecting so many files? Is there a reason why we don't do that instead?
> 
> And assuming that it is necessary and for future similar changes, this approach should help a lot:
> 
> * Keep the names `ButtonsBarConfiguration` and `BarButton` and make the changes required, i.e. add `VerticialBarButton`. Make a commit. This will be the core part of the PR to review. Note the commit hash and point out to look at this commit in the PR.
> * Have a separate commit to rename `ButtonsBarConfiguration` and `BarButton`. Note that this is different from multiple "WIP" commits pushed, which doesn't help reviewing the changes.
> * We have the option to squash the 2 commits if necessary before merge
> 
> We ought to strive to get PRs merged as soon as we reasonably can. And one of the ways to do this is to make it quicker and safer to merge changes.
> 
> I'm getting this in the console:
> 
> > User defaults from command line:
> > IDECustomDerivedDataLocation = /Users/hboon/Library/Caches/JetBrains/AppCode2021.2/DerivedData
> > Resolve Package Graph
> > Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
> > Resolved source packages:
> > WalletConnect: https://github.com/WalletConnect/WalletConnectSwiftV2 @ 0.5.0
> > xcodebuild: error: Could not resolve package dependencies:
> > Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
> 
> Is it because you have upgraded Xcode and I haven't? Is the upgrade necessary? And anyway this change doesn't look like it should be part of this PR, can you remove it?

I could rename them back to ButtonsBar. Then after everything is merged, just do a PR for the name change.

AlphaWallet/Core/Views/ButtonsBar/VerticalButtonsBar.swift
eviltofu: > Is it because you have upgraded Xcode and I haven't? Is the upgrade necessary? And anyway this change doesn't look like it should be part of this PR, can you remove it?

@hboon Do you want me to downgrade back to 13.2.1? If so, you should announce this to the entire team. I didn't make this change, Xcode changed it by itself.

hboon: > I didn't make this change, Xcode changed it by itself.

You should (have) check your PR for what has changed and excluded changes you didn't expect :)

> Do you want me to downgrade back to 13.2.1

Current options for you is to use the older Xcode (and use this https://gist.github.com/steipete/d9b44d8e9f341e81414e86d7ff8fb62d to make it work), install the older simulators too. Or use the newer Xcode version as long as PRs are compatible. I think the change with the SPM metadata is just a one-off incompatibility.

Also, better solution for longer term: can you help add a `.xcode-version` file #4093

hboon:
> > 52 files are modified in the PR. There's too much noise and takes too long to review.
> 
> @hboon I could rename them back to ButtonsBar. Then after everything is merged, just do a PR for the name change.

I had to do the rename in order to check what had change, but there were many whitespace changes introduced (which are fine in this PR because the files included other changes, but those changes were actually because of the rename in the first place so would have been excluded if the rename was a separate commit). So yeah, if they were a separate commit, it'd be much more efficient for both the developer and the reviewer(s).

I'd say 1 PR might be slightly better because sometimes the first commit (that does the actual change, adding the vertical version) might not so clean cut without the rename done, so 1 PR lets us clean it up before merge. Up to you if you want 1 PR or 2 separate PRs as long as it works well and the result is clean.

eviltofu:
> > > 52 files are modified in the PR. There's too much noise and takes too long to review.
> > 
> > 
> > @hboon I could rename them back to ButtonsBar. Then after everything is merged, just do a PR for the name change.
> 
> I had to do the rename in order to check what had change, but there were many whitespace changes introduced (which are fine in this PR because the files included other changes, but those changes were actually because of the rename in the first place so would have been excluded if the rename was a separate commit). So yeah, if they were a separate commit, it'd be much more efficient for both the developer and the reviewer(s).
> 
> I'd say 1 PR might be slightly better because sometimes the first commit (that does the actual change, adding the vertical version) might not so clean cut without the rename done, so 1 PR lets us clean it up before merge. Up to you if you want 1 PR or 2 separate PRs as long as it works well and the result is clean.

So if you have already checked out what changed due to the ButtonsBar rename, do you still want me to do a second PR?

AlphaWallet/Foundation/Repeat.swift
eviltofu: I'll take this out.

hboon:
> @hboon So if you have already checked out what changed due to the ButtonsBar rename, do you still want me to do a second PR?

I didn't continue checking. I gave up because I didn't rename all of them.

AlphaWallet/Core/Views/ButtonsBar/VerticalButtonsBar.swift
eviltofu: > I'm getting this in the console:
> 
> > User defaults from command line:
> > IDECustomDerivedDataLocation = /Users/hboon/Library/Caches/JetBrains/AppCode2021.2/DerivedData
> > Resolve Package Graph
> > Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
> > Resolved source packages:
> > WalletConnect: https://github.com/WalletConnect/WalletConnectSwiftV2 @ 0.5.0
> > xcodebuild: error: Could not resolve package dependencies:
> > Package.resolved file is corrupted or malformed; fix or delete the file to continue: unsupported schema version 2
> 
> Is it because you have upgraded Xcode and I haven't? Is the upgrade necessary? And anyway this change doesn't look like it should be part of this PR, can you remove it?

@hboon Should we put those type of files into gitignore? It appears to be reconstructed if missing right?

eviltofu:
@hboon HorizontalButtonsBar have been renamed back.

hboon:
@eviltofu thanks. Have you seen this?:

> Is it possible to continue to call ButtonsBar that and then pass in an argument or configuration which determines if the layout is vertical or horizontal, and defaulting to horizontal? That seems like it would avoid duplicating or having very similar code and affecting so many files? Is there a reason why we don't do that instead?

eviltofu:
> @eviltofu thanks. Have you seen this?:
> 
> > Is it possible to continue to call ButtonsBar that and then pass in an argument or configuration which determines if the layout is vertical or horizontal, and defaulting to horizontal? That seems like it would avoid duplicating or having very similar code and affecting so many files? Is there a reason why we don't do that instead?

@hboon That would require refactoring the original `ButtonsBar` and make the VerticalButtonsBar a subclass of ButtonsBar which requires me to do more work. Have you reviewed the current code so I can proceed with the next step of upgrading the other buttons in other view controllers?

hboon:
No. It seems to be pretty much self-contained, so go ahead with the rest.

eviltofu:
> No. It seems to be pretty much self-contained, so go ahead with the rest.

@hboon Should I do the rename for HorizontalButtonsBar now?

hboon:
> @hboon Should I do the rename for HorizontalButtonsBar now?

Sure. Can you rename, then merge this PR? If you have follow up changes like:

>  so I can proceed with the next step of upgrading the other buttons in other view controllers?

… create a new PR.

hboon (approved):

PR #4098, opened 2022-03-17
Added .xcode-version file with 13.2.1 as target

### LINKED ISSUE #4093: Add .xcode-version file
https://github.com/fastlane/ci/blob/master/docs/xcode-version.md and point it to 13.2.1?

We can change it after this, but with another PR and so everyone (just the 3 of us for now) a chance to know they have to and what is involved.

eviltofu:
Closes #4093

I can't find anyway to test this.


hboon (approved):
We can verify it with tools like https://github.com/xcpretty/xcode-install

PR #4101, opened 2022-03-18
Added Package.resolved to .gitignore file and deleted Package.resolved from repository

### LINKED ISSUE #4099: Place Package.resolved into .gitignore file
Between Xcode 13.2.1 and 13.3, the format for Package.resolved changed from version 1 to version 2. This change is not backward compatible. By placing Package.resolved into the .gitignore file, we can workaround this issue.

eviltofu:
Closes #4099


hboon:
@eviltofu I read [this](https://forums.swift.org/t/package-resolved-should-go-in-the-gitignore/14699/11) and it suggests it's the equivalent of `Gemfile.lock` and `Podfile.lock`. Can you help verify? If that's true then they shouldn't be ignored and we might just have to force everyone to upgrade Xcode this time, hoping it's an unfortunately one-off incompatibility in the file format — if it's really the same as those 2 files, it basically means if someone build a project that was maintained using the older Xcode (with v1 file format) with the newer Xcode (with v2 file format), the dependencies used might unknowingly change as that file is silently/automatically regenerated.

hboon:
Sorry, didn't see the URL earlier when we looked at `http://github.com/github/gitignore`

eviltofu:
@hboon [this](https://github.com/apple/swift-evolution/blob/master/proposals/0175-package-manager-revised-dependency-resolution.md) seems to indicate that the Package.resolved is used to resolve versions if it exists. Placing it in `.gitignore` is not advised.

So what is the decision on which version of Xcode to proceed?

hboon:
> @hboon this seems to indicate that the Package.resolved is used to resolve versions if it exists. Placing it in .gitignore is not advised.

Right, let's keep it out of `.gitignore`.

(I wrote the following because I mis-read your comment, but keeping it for posterity):

Right it says this:

> The version of every resolved dependency will be recorded in a Package.resolved file in the top-level package, and when this file is present in the top-level package it will be used when performing dependency resolution, rather than the package manager finding the latest eligible version of each package. swift package update will update all dependencies to the latest eligible versions and update the Package.resolved file accordingly.

and:

> If this file is gitignored, each user will separately choose when to get new versions based on when they run the swift package update command, and new users will start with the latest eligible version of each dependency.

So given this dependency:

```
FooPackage
```

With no `Package.resolved` file:

Xcode resolves it, pulling the latest and writes `Package.resolved`

```
FooPackage = version N
```

The maintainer of `FooPackage` publishes the version to N+1

If `Package.resolved` is not in `.gitignore`, everyone building the app will use continue to use N.

But if `Package.resolved` is in `.gitignore`, Xcode will create a new `Package.resolved` like this:

```
FooPackage = version N+1
```

And this is not what we want. And this is why we must not add `Package.resolved` (or `Gemfile.lock`, `Podfile.lock`, `yarn.lock` and any similar file) into `.gitignore`.

hboon:
> So what is the decision on which version of Xcode to proceed?

Close this, prepare a separate PR to upgrade (.xcode-version) and note that `Package.resolved` incompatibility. If anyone has to work with an older Xcode, they can figure out to do a `git revert` themselves.

hboon:
> > So what is the decision on which version of Xcode to proceed?
> 
> Close this, prepare a separate PR to upgrade (.xcode-version) and note that `Package.resolved` incompatibility. If anyone has to work with an older Xcode, they can figure out to do a `git revert` themselves.

^ Assuming that you are confident the newer Xcode works for us.

eviltofu:
> > > So what is the decision on which version of Xcode to proceed?
> > 
> > 
> > Close this, prepare a separate PR to upgrade (.xcode-version) and note that `Package.resolved` incompatibility. If anyone has to work with an older Xcode, they can figure out to do a `git revert` themselves.
> 
> ^ Assuming that you are confident the newer Xcode works for us.

It does not out of the box. The tests don't work for 15.4 as the default 15.2 OS is not included. I will make a new PR to update the tests.

hboon:
> @hboon It does not out of the box. The tests don't work for 15.4 as the default 15.2 OS is not included. I will make a new PR to update the tests to 15.4.

Do it in the same PR? Atomic commits, remember.

eviltofu:
This issue can be closed since we have moved to Xcode 13.3.1.

PR #4114, opened 2022-03-18
Add Swap feature #3919

### LINKED ISSUE #3919: Add Swap feature
### How to access Swap feature?

- For now, we let users to access the swap feature from the token menu (we will keep Send and Swap as main actions)
ETH-Short

- The second option will be in the dotted menu at the top
Wallet-Menu

- When we will test it positively (I know how bad it sounds nowadays), we can transform the tab bar so it has one more button int the centre. 
Wallet-New
Wallet-Actions

### Invision prototype
https://invis.io/Z412DM7HGXTF

### Screenshots

- Default view
Swap-Empty

- Token selection
Swap-Select-Token

- Settings (you access in the top right icon)
Swap-Settings

- Tokens selected, no amount entered
Swap-Default

- Entering amount
Swap-Entered

- You can expand the Fees view
Swap-Expanded

- Confirmation
Swap-Confirmation

### Zeplin:
https://zpl.io/7wkvBnW

### Working on:
Few more states, when the price has been updated, no search results when selecting a token, switching to networks etc.

### LINKED ISSUE #4004: Build Swap UI
Refer to: #3919

Use a feature flag to disable it. Might also need to use the feature flag to switch between the current swap (opening a browser) and this new one.

Boon is still exploring the backend/API.

oa-s:
Closes #3919, Closes #4004


AlphaWallet/Settings/Types/Constants.swift
```diff
@@ -225,4 +225,11 @@ public struct Constants {
     static let fetchContractDataTimeout = TimeInterval(4)
 
     static let refreshTokensThresholdSec: Int = 1
+
+    enum AmountTextField {
+        static let allowedCharacters: String = {
+            let decimalSeparator = Config.locale.decimalSeparator ?? ""
+            return "0123456789" + decimalSeparator + EtherNumberFormatter.decimalPoint
```
hboon, line 232: Can you check if this might break handling for "," and "."? I forgot if we wanted to let users enter both.

hboon:
It looks like some changes like in `KeyboardChecker.swift` can be moved into a separate PR and merged first. If you have the chance, can you help to create those? Then this one becomes a bit more manageable.

AlphaWallet/Settings/Types/Constants.swift
oa-s: that is what was in AmountTextField, just extracted it to constants, no other changes for it.

oa-s:
> KeyboardChecker.swift

there were just experiments, there will be no changes

hboon:
Strange. I thought I posted the following. DId I post them to the wrong PR/issue. Any way, can you help to check:

---

Looks great. Thanks! Couple of things:

- [  ] There's a strange animation after tapping "swap any tokens" to the Swap UI appearing. Maybe it's the navigation controller for the swap UI being animated in

- [ ] After Tapping "Select Token" and then tapping outside the actionsheet of tokens, the swap UI pops and go back to the Wallet tab. It shouldn't. It works correctly if the x button in the actionsheet is tapped — the actionsheet is closed.

 - [ ] Need to tweak the label width for longer symbols:
 - [ ] The title "Select Asset" is too close to the "handlebar" above it:
 - [ ] Tapping the FROM section doesn't show an actionsheet like "TO"'s:
 - [ ] Balance for "FROM" isn't displayed:
 - [ ] "All Funds" doesn't work:
 - [ ] Tapping the down arrow should swap the 2 tokens:

Screenshot 2022-03-22 at 1 59 39 PM

 - [ ] The checkmark (filled and empty) looks like it's smaller than in Tomek's design:
 - [ ] The title "Settings" is too close to the "handlebar" above it:
 - [ ] Tapping outside the actionsheet should close it:

Screenshot 2022-03-22 at 2 01 25 PM

- [ ] Is the "Fees" and "Current Price" section built too? How can I make it appear?:



- [ ] This too. How can I make it appear?:

oa-s:
Currently we are using an updated `AmountTextField` control to select `from` and `to` tokens. And we are able to select only to token (tokens list appears). For `from` token we able toggle beetween feat and crypto currencies, do we need such functionality here?

oa-s:
>  Is the "Fees" and "Current Price" section built too? How can I make it appear?:

Entering amount to swap with selected `to` token displays these sections

hboon:
> Currently we are using an updated AmountTextField control to select from and to tokens. And we are able to select only to token (tokens list appears). For from token we able toggle beetween feat and crypto currencies, do we need such functionality here?

Ah I see. OK, can we make the `FROM` also tap to show the selector? We might (or might not) use it right away when we launch the swap feature since it's triggered from a token's screen but later when we have the Swap tab, it must be configurable. Also easier to test with it :)

This leaves tapping the little triangle to tap the fiat/crypto values, but we can figure out how to fix it if it becomes a problem.

oa-s:
@hboon Does native crypto supported for all `Swap Pairs`? for now when we select `from` token as native crypto (`bsc` server for example) `connection.toTokens` doesn't contain address and `0x0000000000000000000000000000000000000000` for server: `bnb`?

hboon:
Not sure about bnb. Not sure. Have to check the API output. You can assume it’s not supported. If there’s a bug, it’s in the parsing to form SwapPairs

oa-s:
Seems like `0x0000000000000000000000000000000000000000` returns only for `mainnet`

oa-s:
@hboon  i a little bit confused with swap feature, do i understand right, it's swap between tokens on single chain, and not with different chains (not bridge between chains), right?

hboon:
Correct. But the API does support bridging. But we are not implementing it yet.

oa-s:
> Correct. But the API does support bridging. But we are not implementing it yet.

ok, i see, i was a bit confused with `from` and `to` servers in model ;)

oa-s:
Yes, working on it

oa-s:
@hboon pr is ready for review

hboon (commented):
Looks great. Need some more testing. Is this completed, only missing making it available in individual token screens?

- [ ] Can you keep the functionality available like it is as a 5th, middle button, but hidden with a feature flag? We can use Settings > Advanced > Features to enable it for TestFlights. The PR doesn't make the swap functionality accessible from elsewhere, right?
- [ ] For Polygon, it says 1 WETH = 0 USDC (seems to show reasonable values for Ethereum mainnet):

Screenshot 2022-05-25 at 2 02 59 PM

- [ ] Upon tapping the swap tab button and showing ETH (native token) on Ethereum mainnet, the destination list is empty. Is it meant to be supported? (I forgot if li.fi allows ETH => X)

- [x] There should not be the AlphaWallet Fee row since we are not getting a cut:

Screenshot 2022-05-25 at 2 07 01 PM

hboon:
This PR's quite big, after we hide it behind a feature flag and if the code is cleaned up, we should merge it since/if it doesn't break the app.

oa-s:
> Looks great. Need some more testing. Is this completed, only missing making it available in individual token screens?

i will double check, but it should. loading may take some time, (to ckeck if token supports swap)

oa-s:
> * Can you keep the functionality available like it is as a 5th, middle button, but hidden with a feature flag? We can use Settings > Advanced > Features to enable it for TestFlights. The PR doesn't make the swap functionality accessible from elsewhere, right?

it accessible from token screen too. But i got your idea.

oa-s:
>  For Polygon, it says 1 WETH = 0 USDC (seems to show reasonable values for Ethereum mainnet):

haven't tested this case, but i got an another pr, (pretend watch wallet is real), will check it

oa-s:
> Upon tapping the swap tab button and showing ETH (native token) on Ethereum mainnet, the destination list is empty. Is it meant to be supported? (I forgot if li.fi allows ETH => X)

its an issue with swap pairs json. there is an empty response for ETH on Ethereum mainnet

oa-s:
>  There should not be the AlphaWallet Fee row since we are not getting a cut:

got it

PR #4135, opened 2022-03-22
Prepend current token balance value when subscribe for new updates #4134

### LINKED ISSUE #4134: Prepend current token balance value when subscribe for new updates

oa-s:
Closes #4134


hboon:
@oa-s can you elaborate what it means by "prepend current token balance value" ?

oa-s:
`prepend` i took this word from combine operator called `prepend` its an alternative for `startWith` in RxSwift.
the reason we need to use it is that when we do subscribe updates it need to return an initial value right after subcribing

hboon (approved):

PR #4162, opened 2022-03-26
Notification appear in loop after receving ether #4161

### LINKED ISSUE #4161: Notification appear in loop after receving ether
Sent transaction doesn't get updated with its latest state, and starts displaying local notification in loop:

When used send ether from wallet **A** to wallet **B**, and after send switch to wallet **B** the app starts receive local notification in loop. That happens bacause the app doesn't get updated with latest transaction state.

oa-s:
Closes #4161

### DIFF
--- a/AlphaWallet/Transactions/Storage/TransactionsStorage.swift
+++ b/AlphaWallet/Transactions/Storage/TransactionsStorage.swift
@@ -39,6 +39,11 @@ class TransactionDataStore {
                 TransactionDataStore.functional.nonEmptyIdTransactionPredicate(server: tokenObject.server),
                 filter.predicate
             ])
+        case .predicate(let p):
+            predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
+                TransactionDataStore.functional.nonEmptyIdTransactionPredicate(servers: servers),
+                p
+            ])
         case .all:
             predicate = TransactionDataStore.functional.nonEmptyIdTransactionPredicate(servers: servers)
         }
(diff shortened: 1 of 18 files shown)


AlphaWallet/Transactions/Storage/TransactionsStorage.swift
```diff
@@ -216,7 +225,11 @@ class TransactionDataStore {
             if each.isERC20Interaction {
                 return true
             } else {
-                return realm.object(ofType: Transaction.self, forPrimaryKey: each.primaryKey) == nil
```
oa-s, line 219: I am not quite sure if this chages  could brake any logic. But its required to make transaction to be updated. @hboon any suggestions?

hboon: Comparing `primaryKey` ensures it's the same transaction (because it has both transaction hash and chain ID).

The blockNumber comparison looks a little odd, can you explain what it is checking for? This:

```
return each.blockNumber != tx.blockNumber && each.blockNumber != 0
```

oa-s: when we sent ether from wallet A to wallet B, we add transaction to storage, added transaction is in panding state, later
when we receive callback for `FetchLatestTransactionsOperation` we receive same transaction, with updated information, updated state, block number etc.
so with current implementation we are not able to update our stored transaction with a new received tx, because we already have added tx in storage.
(as function call `filterTransactionsToNotOverrideERC20Transactions` returns an array with 0 elements)

hboon: Ah I see. Not sure if it will break it actually. Let's just go with it and see what happens. I think it will just have a strange visual effect as described in this comment, so let's try:

>     //We pull transactions data from the normal transactions API as well as ERC20 event log. For the same transaction, we only want data from the latter. Otherwise the UI will show the cell display switching between data from the 2 source as we fetch (or re-fetch)

oa-s: ok. In the other PR we can make subscribing for tx updates, for transaction view controller and in other parts of available.

hboon:
@oa-s Ready for review?

Issue #3948, opened 2022-02-17
Input accessory button in Send screen for amount textbox should say "Next", not "Done"


hboon:
<img width="487" alt="Screenshot 2022-02-17 at 12 03 47 PM" src="https://user-images.githubusercontent.com/56189/154403049-765f752f-ba33-4a23-aa36-39acd067649a.png">

PR #4202, opened 2022-03-31
Added function to update AmountTextField’s InputAccessoryView based on button type. SendViewController modified to use the "next" InputAccessoryView.

eviltofu:
closes: #3948

Simulator Screen Shot - iPhone 11 - 2022-03-31 at 23 42 28


hboon (approved):
I suggest the following commit message instead:

> SendViewController modified to use the "next" InputAccessoryView. 

The detail about adding a function isn't necessary in comparison (though it would be appropriate if it's a separate commit)

PR #4213, opened 2022-04-02
Added highlighted button background colours to Primary, Secondary, and More button styles in ButtonsBarViewModel.

### LINKED ISSUE #3946: Primary/secondary/Ellipsis button colors when pressed and held down
While we are at it, do we want to set a "selected" foreground and background color which is used when the primary and secondary buttons are held down? Apparently there's one for the secondary button already (probably provided by OS), but not for the primary button. Let's do it for both, deliberately, for dark theme too, i.e. 4 combinations:

1. primary foreground + background color for light theme
2. secondary foreground + background color for light theme
3. primary foreground + background color for dark theme
4. secondary foreground + background color for dark theme

Screenshot 2022-02-17 at 11 07 36 AM

eviltofu:
Closes #3946 

Simulator Screen Shot - iPhone 11 - 2022-04-02 at 09 00 27 copy 2 
The primary button when highlighted.

Simulator Screen Shot - iPhone 11 - 2022-04-02 at 09 00 14 copy 2
The secondary button when highlighted.

Simulator Screen Shot - iPhone 11 - 2022-04-03 at 13 31 02 copy 2
The more button when highlighted.


hboon (commented):
Pending https://github.com/AlphaWallet/alpha-wallet-ios/issues/3946#issuecomment-1086564100

oa-s (approved):

PR #4214, opened 2022-04-02
In transaction confirmation actionsheet for a known, decoded function invocation in a transaction, such as approve(), address values should be displayed correctly #3999

### LINKED ISSUE #3999: In transaction confirmation actionsheet for a known, decoded function invocation in a transaction, such as approve(), `address` values should be displayed correctly
Probably reproducible by trying to add liquidity described in: #3998

Should just display 0xa5e, etc without mentioning "EthereumAddress" or ".normal"

Screenshot 2022-02-22 at 5 02 14 PM

oa-s:
Closes #3999


oa-s:
Didn't tested, verify first

hboon (commented):
Still the same problem. I think you should be able to manually create a ERC20 approve() call to trigger this.

oa-s:
> approve

Ah i see,  we had some kind of this issue in past, types missmatch in pods `EthereumAddress` and `web3swift`

oa-s:
@hboon do we have formatter from string representation for `BigUInt`, i know that we able to get formatted value with `EtherNumberFormatter.full.string(from:` for `BigInt` type ?

PR #4215, opened 2022-04-02
TokenViewController sometimes show 0 balance for native token even when Wallet tab has refreshed to show non-zero #4200

### LINKED ISSUE #4166: 0 Matic Balance Shown in USD and Can't Send
AW fails to read the Matic balance in USD and it doesn't allow it sending. If you change the connected wallet for another one and change back again you will see the balance...

### LINKED ISSUE #4200: TokenViewController sometimes show 0 balance for native token even when Wallet tab has refreshed to show non-zero
While testing #4198, I noticed this:

I noticed this:

1. Make sure only Klaytn testnet is enabled
2. Import a new wallet
3. Balance in wallet tab is 0
4. Wait
5. Balance in wallet tab is non-zero
6. Tap native token

Expected
---
7. Balance in TokenViewController is non-zero

Observed
---
7. Balance in TokenViewController is still 0

Going back to the wallet tab, the balance is still (correctly) non-zero, tapping into TokenViewController, it still shows 0. After I switch from Klaytn testnet -> mainnet, the non-zero balance for mainnet is shown correctly in both Wallet tab and TokenViewController. After I switch back to testnet, it now shows non-zero correctly for testnet and I can't reproduce the issue anymore.

Wallet|TokenViewController
-|-
Screenshot 2022-03-31 at 10 40 33 AM|Screenshot 2022-03-31 at 10 40 34 AM

oa-s:
Closes #4200, Closes #4166

### DIFF
--- a/AlphaWallet/Accounts/ViewControllers/AccountsViewController.swift
+++ b/AlphaWallet/Accounts/ViewControllers/AccountsViewController.swift
@@ -101,11 +101,12 @@ class AccountsViewController: UIViewController {
 
     @objc private func pullToRefresh(_ sender: UIRefreshControl) {
         tableViewRefreshControl.beginRefreshing()
-        walletBalanceService.refreshBalance(updatePolicy: .all, force: true).done { _ in
-            // no-op
-        }.cauterize().finally { [weak self] in
-            self?.tableViewRefreshControl.endRefreshing()
-        }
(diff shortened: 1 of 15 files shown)


hboon (approved):
Seems to involve quite a bit of changes across many files.

oa-s:
i found that that it had happened because of to early subscribing for TokenObjects changes, `when the native token hasn't added to data store`. that is why there were needed to change batch of files

PR #4216, opened 2022-04-02
Not handling types correctly when signing EIP712 messages #4000

### LINKED ISSUE #4000: Not handling types correctly when signing EIP712 messages
uint256 are stored in correctly as a floating point number, so when they are signed, it's quite likely the value used is wrong.

Not sure if this is a problem for other types, but likely just affects uint(256).

Description below would sound a little strange as it is phrased to describe the UI display, but it has the same underlying problem:

---

Need to verify if this is v3 or v4, or both

Examples:

* `uint256` should not be in scientific mode (truncate if too long)
* `address` should be displayed as "0xabcd…cdef", not string

1. Go to http://quickswap.exchange in dapp browser
2. Tap Pool
3. Tap Add Liquidity
4. A tiny bit of MATIC + USDC would do (0.01 MATIC + 0.02 USDC)
5. Go through the process of adding liquidity, approving, etc and finally posting the transaction (might be reproducible while adding too)
6. Wait till transaction is confirmed
7. Tap Pool
8. Tap Add Liquidity
9. Tap Manage on the pool, we added previously

    

10. Tap Remove

    

11. Slide to choose something, say 50%

    

12. Tap Approve (Look at the Signature request. Don't proceed with transaction)

    

An example of the EIP712 payload (maybe it's possible to replicate this without going through the steps above):

```
{
   "domain" : {
      "chainId" : 137,
      "name" : "Uniswap V2",
      "verifyingContract" : "0x6e7a5FAFcec6BB1e78bAE2A1F0B612012BF14827",
      "version" : "1"
   },
   "message" : {
      "deadline" : 1645588039,
      "nonce" : "0x00",
      "owner" : "0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362",
      "spender" : "0xa5E0829CaCEd8fFDD4De3c43696c57F7D7A678ff",
      "value" : "2815176794"
   },
   "primaryType" : "Permit",
   "types" : {
      "EIP712Domain" : [
         {
            "name" : "name",
            "type" : "string"
         },
         {
            "name" : "version",
            "type" : "string"
         },
         {
            "name" : "chainId",
            "type" : "uint256"
         },
         {
            "name" : "verifyingContract",
            "type" : "address"
         }
      ],
      "Permit" : [
         {
            "name" : "owner",
            "type" : "address"
         },
         {
            "name" : "spender",
            "type" : "address"
         },
         {
            "name" : "value",
            "type" : "uint256"
         },
         {
            "name" : "nonce",
            "type" : "uint256"
         },
         {
            "name" : "deadline",
            "type" : "uint256"
         }
      ]
   }
}
```

Notice that we can look up the types of the fields "owner" and "deadline" as "address" and "uint256" respectively. Similarly for other attributes. Maybe this has to be expanded to match and hold the available types?

```
extension EIP712TypedData {
    enum JSON: Equatable {
        case string(String)
        case number(Float)
        case object([String: JSON])
        case array([JSON])
        case bool(Bool)
        case null
    }
}
```

in https://github.com/AlphaWallet/alpha-wallet-ios/blob/97cf75a669038e8e6c400601d0001c7d577b4dd4/AlphaWallet/WalletConnect/Types/EIP712TypedData.swift#L237

---

Need to verify if this is v3 or v4, or both

Examples:

* `uint256` should not be in scientific mode (truncate if too long)
* `address` should be displayed as "0xabcd…cdef", not string

1. Go to http://quickswap.exchange in dapp browser
2. Tap Pool
3. Tap Add Liquidity
4. A tiny bit of MATIC + USDC would do (0.01 MATIC + 0.02 USDC)
5. Go through the process of adding liquidity, approving, etc and finally posting the transaction (might be reproducible while adding too)
6. Wait till transaction is confirmed
7. Tap Pool
8. Tap Add Liquidity
9. Tap Manage on the pool, we added previously

    

10. Tap Remove

    

11. Slide to choose something, say 50%

    

12. Tap Approve (Look at the Signature request. Don't proceed with transaction)

    

An example of the EIP712 payload (maybe it's possible to replicate this without going through the steps above):

```
{
   "domain" : {
      "chainId" : 137,
      "name" : "Uniswap V2",
      "verifyingContract" : "0x6e7a5FAFcec6BB1e78bAE2A1F0B612012BF14827",
      "version" : "1"
   },
   "message" : {
      "deadline" : 1645588039,
      "nonce" : "0x00",
      "owner" : "0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362",
      "spender" : "0xa5E0829CaCEd8fFDD4De3c43696c57F7D7A678ff",
      "value" : "2815176794"
   },
   "primaryType" : "Permit",
   "types" : {
      "EIP712Domain" : [
         {
            "name" : "name",
            "type" : "string"
         },
         {
            "name" : "version",
            "type" : "string"
         },
         {
            "name" : "chainId",
            "type" : "uint256"
         },
         {
            "name" : "verifyingContract",
            "type" : "address"
         }
      ],
      "Permit" : [
         {
            "name" : "owner",
            "type" : "address"
         },
         {
            "name" : "spender",
            "type" : "address"
         },
         {
            "name" : "value",
            "type" : "uint256"
         },
         {
            "name" : "nonce",
            "type" : "uint256"
         },
         {
            "name" : "deadline",
            "type" : "uint256"
         }
      ]
   }
}
```

Notice that we can look up the types of the fields "owner" and "deadline" as "address" and "uint256" respectively. Similarly for other attributes. Maybe this has to be expanded to match and hold the available types?

```
extension EIP712TypedData {
    enum JSON: Equatable {
        case string(String)
        case number(Float)
        case object([String: JSON])
        case array([JSON])
        case bool(Bool)
        case null
    }
}
```

in https://github.com/AlphaWallet/alpha-wallet-ios/blob/97cf75a669038e8e6c400601d0001c7d577b4dd4/AlphaWallet/WalletConnect/Types/EIP712TypedData.swift#L237

oa-s:
Fixes #4000


AlphaWallet/Transfer/ViewModels/ConfirmSignMessageViewControllerViewModel.swift
```diff
@@ -258,7 +258,7 @@ extension EIP712TypedData.JSON {
             return NSAttributedString(string: fittedString.indented(indentationLevel), attributes: valueAttributes)
 
         case .number(let value):
-            return NSAttributedString(string: String(value).indented(indentationLevel), attributes: valueAttributes)
+            return NSAttributedString(string: value.debugDescription.indented(indentationLevel), attributes: valueAttributes)
```
hboon, line 261: The PR fixes the issue, but can you replace `debugDescription` with something else? The result returned by `debugDescription` is subject to change by Apple.

PR #4225, opened 2022-04-04
Feature/Add-Swapzone.io-to-Dapps-List

vrogov:
Swapzone is a cryptocurrency exchange aggregator. We want to make our integration to your wallet, so that users will be able to exchange their crypto coins and tokens (Swapzone supports over 300 assets) with best rates and prices, custody-free, with no registration needed.


hboon (approved):

Issue #4199, opened 2022-03-31
`RemoteLogger.swift:97:20`: '`DDASLLogger`' was deprecated in iOS 10.0: Use `DDOSLogger` instead


eviltofu:
`DDASLLogger` should be replaced with `DDOSLogger`. The following warning message was given for `RemoteLogger.swift`:

`AlphaWallet/Core/Helpers/RemoteLogger.swift:97:20`: '`DDASLLogger`' was deprecated in iOS 10.0: Use `DDOSLogger` instead

```
        logger.add(DDASLLogger.sharedInstance, with: .debug)
```

PR #4233, opened 2022-04-04
RemoteLogger.swift:97:20: 'DDASLLogger' was deprecated in iOS 10.0: Use DDOSLogger instead #4199

Mac3n:
replace DDASLLogger with DDOSLogger in logger because of DDASLLogger deprecation in iOS 10


Mac3n:
closes #4199

hboon (approved):

hboon:
@Mac3n thank you!

Issue #4259, opened 2022-04-09
Should skip modules/*/Pods/ when running SwiftLint


hboon:
Those are created (but not checked in) if we run `pod install` in inhouse pods

PR #4260, opened 2022-04-09
Skip modules/*/Pods/ when running SwiftLint. Those are created (but not checked in) if we run `pod install` in inhouse pods #4259

hboon:


eviltofu:
Should not the following directories be excluded as well?
```
/docs
/fastlane
/resources
/TokenScriptFiles
/vendor
```
@hboon

hboon:
They could be specified, but they don’t affect things because they are small and don’t contain Swift source files.

eviltofu (approved):

PR #4269, opened 2022-04-10
Fix: URLs in in-house pods

hboon:


hboon:
Force pushed to include app's `Podfile.lock` (as the pods' `Podfile`s has changed, so their content hash changes)

oa-s (approved):

PR #4274, opened 2022-04-10
foo

hboon:
- [Refactor] Move protocol implementations for ENS-related classes into extensions. Easier to follow
- [Refactor] Rename vars and func labels related to ENS to be more intention revealing
- Cache ENS forward and reverse resolution by name and address respectively instead of their hash
- Remove dependency TrustWalletCore from AlphaWalletAddress pod by vendoring some code


hboon:
Closing because of wrong branch name

PR #4275, opened 2022-04-10
Remove dependency TrustWalletCore from AlphaWalletAddress pod

hboon:
Vendors a little code which should be internal to the pod anyway

Depends on and includes: #4273


hboon:
Closing because of wrong branch name

PR #4276, opened 2022-04-10
Remove dependency TrustWalletCore from AlphaWalletAddress pod

hboon:
Vendors a little code which should be internal to the pod anyway

Depends on and includes: https://github.com/AlphaWallet/alpha-wallet-ios/pull/4273


hboon:
Supersedes #4274, #4275

oa-s (approved):

PR #4277, opened 2022-04-10
Move code accessing EethereumAddress pod into AlphaWalletAddress pod so it's hidden in app codebase

hboon:
Depends on and includes #4276


oa-s (approved):

hboon:

PR #4291, opened 2022-04-11
Remove force unwrap for realm write operations, syncronize realm access, perform safeWrite #4290

### LINKED ISSUE #4290: Remove force unwrap for realm write operations, syncronize realm access, perform `safeWrite`

oa-s:
Closes #4290


hboon (approved):
Even if this works, it's still very hard to bear in mind when we should queue (and on which queue) and very easy to introduce bugs though.

But let's try this out first.

PR #4292, opened 2022-04-11
Update Dapps.swift

LGYang:
Add MetaPath and SWFT


hboon (approved):

hboon:

Issue #3546, opened 2021-12-14
Improve Segmented Controls and Token Groups


colourfreak:
Another issue related to the Wallet tab. We are very, very close to the end design I made while designing Samoa.
So based on Hu's recent work, we now have specified token groups. 

- All (Assets, Defi, Governance but no Collectibles)
- Assets (ERC20)
- Collectibles (ERC721 and ERC1155), 
- DeFi (all ctokens, atokens, ptokens, itokens, liquidity pools, etc...), 
- Governance (DAO tokens like COMP, UNI, etc)

**So we have to:**
1. Update Segmented Controls (so they are horizontally scrollable) - ScrollableSegmentedControl
2. Add Token groups

**Zeplin:**
https://zpl.io/bAgG6ox

**Prototype**
https://invis.io/6U125DXFWE2X

<img width="1683" alt="Screenshot 2021-12-14 at 16 50 01" src="https://user-images.githubusercontent.com/51817359/146032038-baaa2a4d-823d-4fd9-a708-589cffa6e022.png">

![Wallet_Long](https://user-images.githubusercontent.com/51817359/146029729-ecb4e36a-0c46-495b-a2f1-8529f38309ca.jpg)

colourfreak:
Examples of scrollablesegmentedcontrol

https://user-images.githubusercontent.com/51817359/146033231-15e38a17-c863-491b-98f8-b530b9993589.MOV

https://user-images.githubusercontent.com/51817359/146033184-487ae25c-5157-4e0b-b900-4a7d6396db63.MOV

JamesSmartCell:
@hboon We need to come up with a way to distinguish these tokens - I don't know of any way to tell from looking at the token, but maybe an API can say? Failing this we should maintain a common list of tokens that are governance or defi.

hboon:
> We need to come up with a way to distinguish these tokens - I don't know of any way to tell from looking at the token, but maybe an API can say? Failing this we should maintain a common list of tokens that are governance or defi.

@JamesSmartCell It's already in the token mapping schema, but the data hasn't been populated yet: https://github.com/AlphaWallet/tokens/blob/main/tokens.json, eg.

```
[
  {
    "contracts": [
      {"address": "0xc18360217d8f7ab5e7c516566761ea12ce7f9d72", "chainId": 1},
      {"address": "0xbD7A5Cf51d22930B8B3Df6d834F9BCEf90EE7c4f", "chainId": 137},
      {
[...]

PR #4304, opened 2022-04-12
Added code to read tokens.json and sort the tokens into defi, goverance, assets, collectibles

### LINKED ISSUE #3581: Read tokens.json and put tokens into the correct group in Wallet tab
Part of #3546
Depends on: #3580

A token matches if the contract and chain ID matches an entry in each array.

This file: https://github.com/AlphaWallet/tokens/blob/main/tokens.json (check update: https://github.com/AlphaWallet/alpha-wallet-ios/issues/3581#issuecomment-1019835025)

"All" should contain every token that is not hidden (like it is before the PR).

If a token is not categorized, display it under "Assets".

Just ignore the tokens tagged as "Collectibles" (if there are any) in the JSON. We still show all and only ERC721 and ERC1155 in the hardcoded Collectibles tab. Otherwise it wouldn't work.

eviltofu:
Closes #3581
Simulator Screen Shot - iPhone 11 - 2022-04-12 at 19 17 52 copy
Old

Simulator Screen Shot - iPhone 11 - 2022-04-12 at 19 16 51 copy
New


eviltofu:
I don't know if there tokens.json file needs to be updated. I can't tell if the defi / governance tabs are properly working.

hboon (changes requested):
> I don't know if the tokens.json file needs to be updated.

It's the correct version.

> I can't tell if the defi / governance tabs are properly working. From the images above, you can see that the assets tabs are both different.

After the changes, it's likely that DeFi and Governance that will be empty or have only a few tokens. You can manually add a toke you know belong to those 2 tabs by:

1. Wallet tab
2. Tap …
3. Add/Hide Tokens
4. Paste and follow through to add (on the right chain)

eviltofu:
I'm looking at a coinbase [article](https://www.coindesk.com/learn/what-are-defi-tokens/) and it says Link, Dai and Luna are DeFi but I've added those and the DeFi tab is still blank. @hboon

eviltofu:
> > I don't know if the tokens.json file needs to be updated.
> 
> It's the correct version.
> 

I've checked the git log. It's not been modified since I added it?

hboon:
Will check when I’m back but check against tokens.json that you embed, not against an external article. If you check against the article, update tokens.json instead.

hboon:
> I've checked the git log. It's not been modified since I added it?

No, the copy you used hasn't been committed yet, so it's the lastest we have.

hboon:
> Will check when I’m back but check against tokens.json that you embed, not against an external article. If you check against the article, update tokens.json instead.

@eviltofu So if you add the contract `0x21bfbda47a0b4b5b1248c767ee49f7caa9b23697` (chaindId = 1), the first entry in `tokens.json` you'll see it appear under the Governance tab. This test will only verify if the tokens.json is parsing correctly of course. Unless you are referring to testing something else?

AlphaWallet.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,16 +1,14 @@
...
-    "pins": [
-      {
-        "package": "WalletConnect",
-        "repositoryURL": "https://github.com/WalletConnect/WalletConnectSwiftV2",
-        "state": {
-          "branch": null,
-          "revision": "5ea0265d34f18ec48d71fc299adacecdacc904bb",
-          "version": "0.5.0"
-        }
+  "pins" : [
```
hboon, line 2: Can you rebase master (or one of the earlier ones which still builds like `31d3fe25affce53c9ce42464b653b9bd97b87537`)? It'll get rid of this in the PR.

hboon (approved):
Can merge after rebasing.

hboon:
Closes #3546

PR #4311, opened 2022-04-13
Fix Swift Package Manager error when building

### LINKED ISSUE #4310: SPM error when building
Screenshot 2022-04-13 at 12 22 16 PM

hboon:
Fixes #4310

@oa-s do you know why? Does both before and after PR work for you?

### DIFF
--- a/AlphaWallet.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/AlphaWallet.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/firebase/abseil-cpp-SwiftPM.git",
       "state" : {
-        "revision" : "fffc3c2729be5747390ad02d5100291a0d9ad26a",
-        "version" : "0.20200225.4"
+        "revision" : "d302de612e3d57c6f4afaf087da18fba8eac72a7",
+        "version" : "0.20220203.1"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


oa-s (approved):

oa-s:
I will check

hboon:
Ok, buzz here once you have confirmed post-PR works for you then I'll merge.

eviltofu:
Can we use one of these flags to prevent the Packages.resolved file to be updated every build?
```
    -resolvePackageDependencies                              resolves any Swift package dependencies referenced by the project or workspace
    -disableAutomaticPackageResolution                       prevents packages from automatically being resolved to versions other than those recorded in the `Package.resolved` file
    -onlyUsePackageVersionsFromResolvedFile                  prevents packages from automatically being resolved to versions other than those recorded in the `Package.resolved` file
    -disablePackageRepositoryCache                           disable use of a local cache of remote package repositories
    -packageCachePath                                        path of caches used for package support
```

hboon:
Force pushed after rebase master with no other modifications to make it clearer there is just 1 commit.

hboon:
> Can we use one of these flags to prevent the Packages.resolved file to be updated every build to solve this issue?

Xcode isn't doing that. And it shouldn't. I *think* the issue was that it read `Packages.resolved`, think it's updated (so didn't resolve, rightfully), but somehow fail to fetch the specified dependencies; and my fix by upgrading it (it worked though) was the wrong way to go. Anyway, this seems to have fixed it self and the pre-PR `Package.resolved` builds now. So closing it.

PR #4316, opened 2022-04-13
Make transaction store safe #4315

### LINKED ISSUE #4315: Make transaction store safe

oa-s:
Closes #4315


hboon (approved):
All those `queue` are scary :)

oa-s:
verifying to make all calls from single queue

PR #4334, opened 2022-04-13
Switch wallet Connect v2 to develop branch #4333

### LINKED ISSUE #4333: Switch wallet Connect v2 to develop branch
Develop branch contains crash fix

oa-s:
Closes #4333

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -7488,8 +7488,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/WalletConnect/WalletConnectSwiftV2";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.5.0;
+				branch = develop;
+				kind = branch;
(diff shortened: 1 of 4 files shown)


oa-s:
Not able to test, example app returns invalid connection url

hboon:
@oa-s it's also possible they broke or modified their example though, so might need to check with them.

oa-s:
i was able to reproduce it when you have connected v2 session, on apps launch it causes crash `Bad Access`
Thiss PR fixes the issue https://github.com/WalletConnect/WalletConnectSwiftV2/issues/156

oa-s:
> @oa-s it's also possible they broke or modified their example though, so might need to check with them.

yep, there missing some parts of url, i will create issue.

hboon:
Ok, do you want to merge first?

oa-s:
Will do it later after testing

oa-s:
@hboon I found that we can manually download walletConnect v2 (its 0.5.0). (even for **spm**) change it locally and add to the proj. And later update code base when a new release will be available.

hboon:
Ok sure

PR #4338, opened 2022-04-14
Added code to persist cache for GetIsERC1155ContractCoordinator

### LINKED ISSUE #4285: Add persistent caching to GetIsERC1155ContractCoordinator when the result is true/false (ie. not throwing an error)
Makes the assumption that a token's type wouldn't change.

We currently already cache EIP165 lookups indirectly via `callSmartContract()` but this is just within an app's lifecycle. The token type shouldn't change across launches. I did a simple count and the app made 68 ERC165 checks if tokentype=ERC1155.

Write to somewhere in `FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)[0]`.

Note that caching should be done in `GetIsERC1155ContractCoordinator` and not `GetInterfaceSupported165Coordinator`.

eviltofu:
Closes #4285

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/GetIsERC1155ContractCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/GetIsERC1155ContractCoordinator.swift
@@ -10,17 +10,25 @@ import web3swift
 
 class GetIsERC1155ContractCoordinator {
     private let server: RPCServer
+    private var cache: CachedERC1155ContractDictionary?
 
     private struct ERC165Hash {
         //https://eips.ethereum.org/EIPS/eip-1155
         static let official = "0xd9b67a26"
     }
-    init(forServer server: RPCServer) {
+    init(forServer server: RPCServer, cacheName: String = "ERC1155ContractCache.json") {
         self.server = server
+        cache = CachedERC1155ContractDictionary(fileName: cacheName)
     }
 
     func getIsERC1155Contract(for contract: AlphaWallet.Address) -> Promise<Bool> {
-        return GetInterfaceSupported165Coordinator(forServer: server)
-            .getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
-    } 
+        if let result = cache?.isERC1155Contract(for: contract) {
+            return Promise.value(result)
+        }
+        return firstly {
+            GetInterfaceSupported165Coordinator(forServer: server).getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
+        }.get { result in
+            self.cache?.setContract(for: contract, result)
+        }
+    }
 }
--- a/AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
+++ b/AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
@@ -1,0 +1,71 @@
+//
+//  CachedERC1155ContractDictionary.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 12/4/22.
+//
+
+import Foundation
+
+class CachedERC1155ContractDictionary {
+    private let fileUrl: URL
+    private var baseDictionary: [AlphaWallet.Address: Bool] = [AlphaWallet.Address: Bool]()
+    private var encoder: JSONEncoder
+
+    init?(fileName: String) {
+        do {
+            var url: URL = try FileManager.default.url(for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
+            url.appendPathComponent(fileName)
+            self.fileUrl = url
+            self.encoder = JSONEncoder()
+            if FileManager.default.fileExists(atPath: url.path) {
+                readFromFileUrl()
+            }
+        } catch {
+            return nil
+        }
+    }
+
+    func isERC1155Contract(for address: AlphaWallet.Address) -> Bool? {
+        return baseDictionary[address]
+    }
+
+    func setContract(for address: AlphaWallet.Address, _ result: Bool) {
+        baseDictionary[address] = result
+        writeToFileUrl()
+    }
+
+    func remove() {
+        do {
+            try FileManager.default.removeItem(at: fileUrl)
+        } catch {
+            // Do nothing
+            verboseLog("CachedERC1155ContractDictionary::remove Exception: \(error)")
+        }
+    }
+
+    private func writeToFileUrl() {
+        do {
+            let data = try encoder.encode(baseDictionary)
+            if let jsonString = String(data: data, encoding: .utf8) {
+                try jsonString.write(to: fileUrl, atomically: true, encoding: .utf8)
+            }
+        } catch {
+            // Do nothing
+            verboseLog("[CachedERC1155ContractDictionary] writeToFileUrl error: \(error)")
+        }
+    }
+
+    private func readFromFileUrl() {
+        do {
+            let decoder = JSONDecoder()
+            let data = try Data(contentsOf: fileUrl)
+            let jsonData = try decoder.decode([AlphaWallet.Address: Bool].self, from: data)
+            baseDictionary = jsonData
+        } catch {
+            verboseLog("[CachedERC1155ContractDictionary] readFromFileUrl error: \(error)")
+            baseDictionary = [AlphaWallet.Address: Bool]()
+        }
+    }
+    
+}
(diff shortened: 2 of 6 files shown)


hboon:
@eviltofu master should build now. Can you rebase and check your branch? There should be just 1 (or a few) new commits even if the branch is slightly behind. It shows 26 in the PR.

eviltofu:
Ok don't merge yet. I have just thought of a way to test something.

hboon:
Good to review and merge?

eviltofu:
> Good to review and merge?

@hboon yes

AlphaWallet/Tokens/Coordinators/GetIsERC1155ContractCoordinator.swift
```diff
@@ -10,17 +10,31 @@ import web3swift
...
+            if let result = cache?.isERC1155Contract(for: contract) {
+                seal.fulfill(result)
+                return
+            }
+            GetInterfaceSupported165Coordinator(forServer: server)
+                .getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
+                .done { result in
+                    self.cache?.setContract(for: contract, result)
+                    seal.fulfill(result)
+                    return
```
hboon, line 35: The `return` isn't needed here.

AlphaWallet/Tokens/Coordinators/GetIsERC1155ContractCoordinator.swift
```diff
@@ -10,17 +10,31 @@ import web3swift
...
+                }
+                .cauterize()
```
hboon, line 37: We shouldn't `cauterize()` the result here because the Promise returned by `getIsERC1155Contract() will never resolve when there's an error from `getInterfaceSupported165()`. PromiseKit has this helpful `firstly {}` function so the promise code is indented (probably async is better, but we'll see when we start adopting it). So we can rewrite this as:


```
if let result = cache?.isERC1155Contract(for: contract) {
    return Promise.value(result)
}

return firstly {
    GetInterfaceSupported165Coordinator(forServer: server).getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
}.get { result in
    self.cache?.setContract(for: contract, result)
}
```

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
```diff
@@ -0,0 +1,76 @@
...
+import Foundation
+
+class CachedERC1155ContractDictionary {
+    private var baseDictionary: [AlphaWallet.Address: Bool] = [AlphaWallet.Address: Bool]()
+    private var fileUrl: URL
+    private var encoder: JSONEncoder
+
+    static func loadFromCache(fileName: String) -> CachedERC1155ContractDictionary? {
+        do {
+            var url: URL = try FileManager.default.url(for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
```
hboon, line 17: Should we be using `cacheUrlFor(fileName: String)` here instead?

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
```diff
@@ -0,0 +1,76 @@
...
+            if FileManager.default.fileExists(atPath: url.path) {
+                cache.readFromFileUrl()
+            }
+            return cache
+        } catch {
+            return nil
+        }
+    }
+
+    private init(url: URL) {
```
hboon, line 29: What's the rationale to write a separate factory method instead of putting it into `init()` like this:

```
init?(fileName: String) {
    do {
        var url: URL = try FileManager.default.url(for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
        url.appendPathComponent(fileName)

        self.fileUrl = url
        self.encoder = JSONEncoder()

        if FileManager.default.fileExists(atPath: url.path) {
            readFromFileUrl()
        }
    } catch {
        return nil
    }
}
```

I didn't check if the code above works correctly.

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
```diff
@@ -0,0 +1,76 @@
...
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 12/4/22.
+//
+
+import Foundation
+
+class CachedERC1155ContractDictionary {
+    private var baseDictionary: [AlphaWallet.Address: Bool] = [AlphaWallet.Address: Bool]()
+    private var fileUrl: URL
```
hboon, line 12: These 2 can be `let`s

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    private func writeToFileUrl() {
+        do {
+            let data = try encoder.encode(baseDictionary)
+            if let jsonString = String(data: data, encoding: .utf8) {
+                try jsonString.write(to: fileUrl, atomically: true, encoding: .utf8)
+            }
+        } catch {
+            // Do nothing
+            NSLog("CachedERC1155ContractDictionary::writeToFileUrl Exception: \(error)")
```
hboon, line 60: Would you use `verboseLog()` instead? and format like this so it's easier to search and filter off:

```
verboseLog("[CachedERC1155ContractDictionary] writeToFileUrl error: \(error)")
```

hboon (approved):
Merge when you are done?

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
eviltofu: Only fileUrl can be let. The baseDictionary is mutated when an address not in the baseDictionary is added. 
```
39: baseDictionary[address] = result
```

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
eviltofu: `cacheUrlFor(fileName: String)` is just used for testing. Do you want me to move it to the project?

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
eviltofu: Reveals intent? I think I prefer functions and initialisers to be just simple and do one thing?

AlphaWallet/Tokens/Coordinators/GetIsERC1155ContractCoordinator.swift
```diff
@@ -10,17 +10,26 @@ import web3swift
...
     }
 
     func getIsERC1155Contract(for contract: AlphaWallet.Address) -> Promise<Bool> {
-        return GetInterfaceSupported165Coordinator(forServer: server)
-            .getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
-    } 
+        if let result = cache?.isERC1155Contract(for: contract) {
+            return Promise.value(result)
+        }
+        return firstly { GetInterfaceSupported165Coordinator(forServer: server)
```
hboon, line 28: Can you format this:

```
return firstly { GetInterfaceSupported165Coordinator(forServer: server)
        .getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
}
.get { result in
    self.cache?.setContract(for: contract, result)
}
```

like this instead?

```
return firstly {
    GetInterfaceSupported165Coordinator(forServer: server).getInterfaceSupported165(hash: ERC165Hash.official, contract: contract)
}.get { result in
    self.cache?.setContract(for: contract, result)
}
```

The indented parts show the async/promise-based code

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
hboon: Oh, I see what you mean. Never mind that then.

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
hboon: This logic is sound generally, but apply this logic for initializers, wouldn't all initializers have to be private and quite often properties would have to be made `var` instead of `let` just because they are initialized in the factory method but otherwise not modified elsewhere? For that reason (var vs let) alone, it seems worth doing it in the `init` most of the time.

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
hboon: There's this too:

```
NSLog("CachedERC1155ContractDictionary::remove Exception: \(error)")
```

hboon (commented):
Dummy text to satisfy GitHub

AlphaWallet/Tokens/Helpers/CachedERC1155ContractDictionary.swift
eviltofu: No since this instance we need to use a cache file in a specific area. If you want me to do it in the init way, tell me and I'll do it otherwise I'd leave it like this. @hboon

hboon: init way.

eviltofu:
@hboon this ok to merge?

hboon:
@eviltofu I didn't check again. If you have modified them according, feel free to merge

PR #4339, opened 2022-04-14
Updated makefile to use —repo-update option when installing pods

eviltofu:
Updated makefile to use --repo-update option when installing cocoapods

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ install_gems: check_bundle setup_path
 	fi
 
 install_pods: check_gems
-	@$(bundle_cmd) exec pod install; \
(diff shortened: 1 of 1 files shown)


hboon (changes requested):
`clean_and_check` maybe? unless it supports exclamations: `check!`. Bit too much for `check` to be wiping out the build artefacts.

hboon:
@eviltofu and this can be split into different commits:

> Updated makefile to show clean, to use —repo-update option when install pods, to beautify the build_and_run_booted target, and added maked check to run clean, install, test #4339

hboon:
Or maybe `clean_and_test` is better, since it's really testing?

oa-s (approved):

PR #4340, opened 2022-04-14
Add build run phase to validate correct credentials for Crashlytics used

hboon:


hboon:
@oa-s have kept a copy of the prod credentials and will use it when submitting/TestFlight, so we can keep the one in the repo unchanged, for public/dev. Like we do now for `Constants+Credentials.swift`

oa-s (approved):

PR #4352, opened 2022-04-15
Makes build_and_run_booted target use beautifier

eviltofu:
Makes build_and_run_booted target use beautifier


hboon (approved):

hboon:
Force pushed after rebase master to resolve conflict

hboon (approved):

Issue #4364, opened 2022-04-18
Travis broken because of package.resolved incompatibility


hboon:

eviltofu:
https://docs.travis-ci.com/user/reference/osx seems to indicate that Xcode 13.2.1 is the current highest version Travis has.

hboon:
Forgot to mark closed by #4365, closing

PR #4365, opened 2022-04-18
Fix Travis test. Was not running because `package.resolved` file now requires Xcode 13.3.1 #4364

hboon:


hboon:
Expedite merge to revive Travis

PR #4385, opened 2022-04-19
Add local version of WalletConnect v2, fixes unbalances calls of beginBackgroundTask and endBackgroundTask #3918

### LINKED ISSUE #3918: WalletConnect v2 warning/error message: [BackgroundTask] Background Task 3
This gets logged to the console:

```
[BackgroundTask] Background Task 3 ("Finish Network Tasks"), was created over 30 seconds ago. In applications running in the background, this creates a risk of termination. Remember to call UIApplication.endBackgroundTask(_:) for your task in a timely manner to avoid this.
```

Comes from the WalletConnect v2 SDK. Does it look like a problem? If so, maybe we can log an issue or propose a fix?

oa-s:
Closes #3918


hboon (commented):
Do you think we can modify the dependency to point to your fork instead? Looks like we can modify the URL and set to a particular commit. Much smaller diff :)

Screenshot 2022-04-19 at 1 19 34 PM

oa-s:
i was trying to point it to fork version, but it founds only source url, not able to add fork url. or maybe i made something wrong. tried several times

hboon:
@oa-s can you see if this works? https://github.com/AlphaWallet/alpha-wallet-ios/tree/use-walletconnect-sdk-fork.

Edited the files in an editor. There are build errors, but seems un-related.

oa-s:
@hboon Thanks a lot, it helps!

PR #4391, opened 2022-04-19
Added code to display Chain Icon View in EnabledServersViewController

### LINKED ISSUE #3511: Display chain icons in Settings > Select Active Networks screen
Maybe chain ID too.

Related: https://github.com/AlphaWallet/alpha-wallet-ios/issues/3464#issuecomment-979787211

eviltofu:
Closes #3511 
before after2


eviltofu:
I'm seeing some icons not being updated and being left as a plain circle. For example when I add Classic Ethereum and go to the main screen, I see the icon there. When I go to the select screen in settings, I only see a plain coloured circle.

AlphaWallet/Settings/ViewModels/ServerViewModel.swift
```diff
@@ -13,6 +13,21 @@ protocol ServerTableViewCellViewModelType {
     var selectionStyle: UITableViewCell.SelectionStyle { get set }
 }
 
+protocol ServerImageTableViewCellViewModelType {
```
hboon, line 16: Is a protocol necessary?

AlphaWallet/Settings/Views/ServerImageTableViewCell.swift
```diff
@@ -0,0 +1,155 @@
...
+//
+
+import UIKit
+
+class ServerImageTableViewCell: UITableViewCell {
+
+    // MARK: - Properties
+
+    // MARK: Private
+    private var chainIconView: ImageView = ImageView()
```
hboon, line 15: Some of these can be `let`s

hboon: Better to always default to `let` unless it's not possible.

hboon (changes requested):

AlphaWallet/Settings/ViewModels/ServerViewModel.swift
eviltofu: I'm just following the style of the previous code.

AlphaWallet/Settings/Views/ServerImageTableViewCell.swift
eviltofu: Actually, all of them can be lets...

eviltofu:
> > I'm seeing some icons not being updated and being left as a plain circle. For example when I add Classic Ethereum and go to the main screen, I see the icon there. When I go to the select screen in settings, I only see a plain coloured circle.
> 
> Looks like this line in func `RPCServerImageFetcher.image(server:)`:
> 
> ```
> if let value = server.staticOverlayIcon {
> ```
> 
> should be replaced with:
> 
> ```
> if let value = server.iconImage {
> ```
> 
> for this purpose.
> 
> Can you check other callers of `RPCServerImageFetcher.image(server:)` to be sure? Don't think we should simply change it. I think maybe there has to be 2 versions of this method and they have to be named better.

`AlphaWallet/UI/TokenObject+UI.swift:35:        return RPCServerImageFetcher.instance.image(server: self)` appears to be the only place where this is called. 
```swift
var walletConnectIconImage: Subscribable<Image> {
        return RPCServerImageFetcher.instance.image(server: self)
    }
```
`walletConnectIconImage ` is referenced twice, in `TransactionConfirmationRPCServerInfoViewModel` and `WalletConnectSessionCellViewModel`.

```swift
    var iconImage: Subscribable<Image> {
        server.walletConnectIconImage
    }
```
and
```swift
    var serverIconImages: [Subscribable<Image>] {
        servers.map { $0.walletConnectIconImage }
    }
```

eviltofu:
I've implemented your change. I now get the Ethereum classic icon in the select network screen. The only other place where this function is used is in the `WalletConnectToSessionViewController` but I can't find it.
Simulator Screen Shot - iPhone 11 - 2022-04-20 at 12 11 25 copy 2 Simulator Screen Shot - iPhone 11 - 2022-04-20 at 12 11 18 copy 2

AlphaWallet/Settings/ViewModels/ServerViewModel.swift
hboon: OK. I don't know if you have changed it, but in the future, let's not overdo it and set a new (better) example then :)

hboon:
@eviltofu

1. https://react-app.walletconnect.com
2. Tap to de-select testnet (mainnet is more important here)
3. Tap to select chains
4. Hit the Connect button
5.

    Scan the QR code with the app's camera

    or

    Alternatively, copy the URL under the clipboard (looks like this: wc:235111db399430767c0844f2352557f15e51eb79e3ac0cc1ef0f2c3f02aea864@2?controller=false&publicKey=2f4948949fd7a902e6058ced5e932400d6f873739fa2ecf83a33a34573885e76&relay=%7B%22protocol%22%3A%22waku%22%7D) and then run the simulator with `Config.Development.shouldReadClipboardForWalletConnectUrl = true` and tap the camera button in the wallet tab. Useful for testing in simulator.
6. Tap the "Networks" row to expand and the icons are shown

I *think* it is alright to change it and in fact might be wrong previously.

Buzz me again if it's good to merge?

eviltofu:
> @eviltofu
> 
> 1. https://react-app.walletconnect.com
> 2. Tap to de-select testnet (mainnet is more important here)
> 3. Tap to select chains
> 4. Hit the Connect button
> 
> ```
> Scan the QR code with the app's camera
> 
> or
> 
> Alternatively, copy the URL under the clipboard (looks like this: wc:235111db399430767c0844f2352557f15e51eb79e3ac0cc1ef0f2c3f02aea864@2?controller=false&publicKey=2f4948949fd7a902e6058ced5e932400d6f873739fa2ecf83a33a34573885e76&relay=%7B%22protocol%22%3A%22waku%22%7D) and then run the simulator with `Config.Development.shouldReadClipboardForWalletConnectUrl = true` and tap the camera button in the wallet tab. Useful for testing in simulator.
> ```
> 
>  6. Tap the "Networks" row to expand and the icons are shown
> 
> I _think_ it is alright to change it and in fact might be wrong previously.
> 
> Buzz me again if it's good to merge?

I got to step 5. I'm confused on step 6. Where is this networks row?

hboon:
This screen:

IMG_9093F8369DAD-1

eviltofu:
Screen Shot 2022-04-22 at 2 48 02 PM copy 2

The icon is displayed properly. I think we can merge this. @hboon

hboon:
@eviltofu no, this is the one I'm referring to  (before accepting the connection):

Screenshot 2022-04-22 at 6 18 30 PM

But yes, it seems to be working correctly.

hboon (approved):

PR #4397, opened 2022-04-20
Aligned bottoms of buttons in back up sequence

### LINKED ISSUE #4090: The bottom buttons should be at the same level for the Back up sequence
same level

eviltofu:
Closes #4090

before_1 before_2 before_3 before_4
Before
after_1 after_2 after_3 after_4
After


hboon:
@eviltofu much better. Thanks.

The bottoms of the 1 button on the left and Clear button on the right aren't vertically aligned. Is that intentional?

Screenshot 2022-04-22 at 11 23 27 AM

eviltofu:
> @eviltofu much better. Thanks.
> 
> The bottoms of the 1 button on the left and Clear button on the right aren't vertically aligned. Is that intentional?
> 
> Screenshot 2022-04-22 at 11 23 27 AM

The button bar is aligned properly but the buttons inside the UIStackView always puts a separating space below the buttons. I can't figure out which combination of fill and layoutMargins are needed to flush it against the bottom. The other alternative is to make my own stack view. @hboon

eviltofu:
> @eviltofu much better. Thanks.
> 
> The bottoms of the 1 button on the left and Clear button on the right aren't vertically aligned. Is that intentional?
> 
> Screenshot 2022-04-22 at 11 23 27 AM

The button bar is aligned properly but the buttons inside the UIStackView always puts a separating space below the buttons. I can't figure out which combination of fill and layoutMargins are needed to flush it against the bottom. The other alternative is to make my own stack view. @hboon

hboon:
I see. Never mind. It doesn't look obvious unless aligned side by side.

hboon (approved):

PR #4414, opened 2022-04-25
Fix crash on update activity #4413

### LINKED ISSUE #4413: Fix crash on update activity

oa-s:
Fixes #4413


AlphaWallet/Activities/ActivitiesService.swift
```diff
@@ -427,11 +427,15 @@ class ActivitiesService: NSObject, ActivitiesServiceType {
...
             //Ugly, but should be safe
-            executeThreadSafe({ [unowned self] in
-                self.activities[index] = updatedActivity
-                self.reloadViewController(reloadImmediately: false)
-
-                self.subscribableUpdatedActivity.value = updatedActivity
+            executeThreadSafe({ [weak self] in
+                guard let strongSelf = self else { return }
+                
+                if strongSelf.activities.indices.contains(index) {
```
hboon, line 433: Doesn't this check here and a few lines above imply that the index here and above could be wrong?

hboon (approved):

PR #4419, opened 2022-04-25
Fix crash when trying to get session of unavailable server #4418

### LINKED ISSUE #4418: Fix crash when trying to get session of unavailable server
Crash happens in function `private func generateImplicitAttributesForToken(forContract contract: AlphaWallet.Address, server: RPCServer, symbol: String) -> [String: AssetInternalValue]`

oa-s:
Fixes #4418


oa-s:
> When is the session for a server unavailable?

Captured it from fire base , one crash related to it. I guess it happens when we switch between servers

PR #4422, opened 2022-04-25
Modified GetENSOwnerCoordinatorTests so that if there is a network or response error, we don’t wait 20 seconds for the expectation to time out.

eviltofu:
When GetENSOwnerCoordinatorTests is run, some of the existing tests waits 20 seconds if there is a network  or response error. This rewrite removes the 20 second wait to time out. The following are the tests:

1. testResolution
2. testResolutionThatHasDifferentOwnerAndResolver
3. testEnsIp10WildcardAndEip3668CcipRead

### DIFF
--- a/AlphaWalletTests/Tokens/Coordinators/GetENSOwnerCoordinatorTests.swift
+++ b/AlphaWalletTests/Tokens/Coordinators/GetENSOwnerCoordinatorTests.swift
@@ -21,13 +21,11 @@ class GetENSOwnerCoordinatorTests: XCTestCase {
         firstly {
             ENSResolver(server: server).getENSAddressFromResolver(forName: ensName)
         }.done { address in
-            if address.sameContract(as: "0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362") {
-                expectation.fulfill()
-            } else {
-                XCTFail("ENS name did not resolve correctly")
-            }
+            XCTAssertTrue(address.sameContract(as: "0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362"), "ENS name did not resolve correctly")
+        }.ensure {
+            expectation.fulfill()
         }.catch { error in
-            XCTFail("ENS name did not resolve correctly: \(error)")
+            XCTFail("Unknown error: \(error)")
         }
         wait(for: expectations, timeout: 20)
     }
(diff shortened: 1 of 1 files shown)


AlphaWalletTests/Tokens/Coordinators/GetENSOwnerCoordinatorTests.swift
```diff
@@ -38,18 +41,21 @@ class GetENSOwnerCoordinatorTests: XCTestCase {
         expectations.append(expectation)
         let ensName = "ethereum.eth"
         let server = makeServerForMainnet()
+        var fetchedAddress: AlphaWallet.Address?
+        var fetchedError: Error?
         firstly {
             ENSResolver(server: server).getENSAddressFromResolver(forName: ensName)
         }.done { address in
-            if address.sameContract(as: "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe") {
```
hboon, line 44: Would it work and be simpler if we move the `expectation.fulfill()` to a `.ensure {}` block at the end of the promise chain so it's always executed.

hboon (commented):
<Text to make GitHub happy>

AlphaWalletTests/Tokens/Coordinators/GetENSOwnerCoordinatorTests.swift
eviltofu: Put in `expectation.fufilled()` in an ensure block.

AlphaWalletTests/Tokens/Coordinators/GetENSOwnerCoordinatorTests.swift
```diff
@@ -58,18 +64,21 @@ class GetENSOwnerCoordinatorTests: XCTestCase {
...
-        }.catch { _ in
-            XCTFail("ENS name relying on ENSIP-10 did not resolve correctly")
+            fetchedAddress = address
+        }.ensure {
+            expectation.fulfill()
+        }.catch {error in
+            fetchedError = error
         }
         wait(for: expectations, timeout: 20)
+        XCTAssertNotNil(fetchedAddress)
```
hboon, line 79: Move the asserts up according like https://github.com/AlphaWallet/alpha-wallet-ios/pull/4423#discussion_r858311795 ? Then merge when ready

hboon (approved):

PR #4423, opened 2022-04-25
Changed timeout in GetIsERC1155ContractCoordinatorTestCase to 20 seconds instead of 10 minutes, fixed up some style issues.

eviltofu:
Changed time out for GetIsERC1155ContractCoordinatorTestCase from 10 minutes to 20 seconds. Fixed up some style issues with PromiseKit calls.

### DIFF
--- a/AlphaWalletTests/Tokens/Coordinators/GetIsERC1155ContractCoordinatorTestCase.swift
+++ b/AlphaWalletTests/Tokens/Coordinators/GetIsERC1155ContractCoordinatorTestCase.swift
@@ -30,20 +30,23 @@ class GetIsERC1155ContractCoordinatorTestCase: XCTestCase {
     func testgetIsERC1155Contract() throws {
         let coordinator = GetIsERC1155ContractCoordinator(forServer: .main, cacheName: fileName)
         let expectation = expectation(description: "Waiting for server response")
-        coordinator.getIsERC1155Contract(for: address1)
-            .done { returnValue in
-                self.result = returnValue
-                expectation.fulfill()
-            }
-            .cauterize()
-        wait(for: [expectation], timeout: 10 * 60)
+        firstly {
+            coordinator.getIsERC1155Contract(for: address1)
+        }.done { returnValue in
+            self.result = returnValue
+        }.ensure {
+            expectation.fulfill()
+        }.catch {error in
+            XCTFail("Unknown error: \(error)")
+        }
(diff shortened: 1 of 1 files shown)


AlphaWalletTests/Tokens/Coordinators/GetIsERC1155ContractCoordinatorTestCase.swift
```diff
@@ -30,13 +30,18 @@ class GetIsERC1155ContractCoordinatorTestCase: XCTestCase {
...
-                expectation.fulfill()
-            }
-            .cauterize()
-        wait(for: [expectation], timeout: 10 * 60)
+        var fetchedError: Error?
+        firstly {
+            coordinator.getIsERC1155Contract(for: address1)
+        }.done { returnValue in
+            self.result = returnValue
+            expectation.fulfill()
```
hboon, line 38: If https://github.com/AlphaWallet/alpha-wallet-ios/pull/4422#discussion_r858264456 works, would you apply it here too?

eviltofu: Applied.

AlphaWalletTests/Tokens/Coordinators/GetIsERC1155ContractCoordinatorTestCase.swift
```diff
@@ -30,13 +30,18 @@ class GetIsERC1155ContractCoordinatorTestCase: XCTestCase {
...
+            coordinator.getIsERC1155Contract(for: address1)
+        }.done { returnValue in
+            self.result = returnValue
+        }.ensure {
+            expectation.fulfill()
+        }.catch {error in
+            fetchedError = error
+        }
+        wait(for: [expectation], timeout: 20)
+        XCTAssertNil(fetchedError, "Unknown error: \(fetchedError!)")
```
hboon, line 44: How about moving the assert up into the `catch {}`? Merge when ready.

hboon (approved):

AlphaWalletTests/Tokens/Coordinators/GetIsERC1155ContractCoordinatorTestCase.swift
eviltofu: I prefer to do all the asserts after an expectation is fulfilled.

PR #4430, opened 2022-04-26
Suppose to fix crash when update nft balance #4429

### LINKED ISSUE #4429: Fix crash when update nft balance

oa-s:
Closes #4429

### DIFF
--- a/AlphaWallet/Tokens/Types/TokensDataStore.swift
+++ b/AlphaWallet/Tokens/Types/TokensDataStore.swift
@@ -406,12 +406,11 @@ enum TokenUpdateAction {
                 }
             }
 
-            result = true
+            tokenObject.balance.removeAll()
+            if !newBalance.isEmpty {
+                tokenObject.balance.append(objectsIn: newBalance)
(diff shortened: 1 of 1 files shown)


AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -406,12 +406,11 @@ enum TokenUpdateAction {
                 }
             }
 
-            result = true
-
-            realm.delete(tokenObject.balance)
-            tokenObject.balance.append(objectsIn: newBalance)
+            tokenObject.balance.removeAll()
```
hboon, line 409: Do we still need `result = true` ?

hboon (commented):
And don't forget to fix the commit message :)

AlphaWallet/Tokens/Types/TokensDataStore.swift
oa-s: for now yes, but i got another non merged pr, where it not needed

PR #4446, opened 2022-04-27
Replace with using single instance for Enjin provider, current impl causes memory leaks #4445

### LINKED ISSUE #4445: Replace with using single instance for Enjin provider, current impl causes memory leaks

oa-s:
Closes #4445


hboon (approved):

oa-s:
> @oa-s Travis is failing

now it should be ok

hboon:
Mege when ready

PR #4457, opened 2022-04-28
Move fetchXMLs to background queue, remove singleton from `AssetDefinitionStore` #4456

### LINKED ISSUE #4456: Move fetchXMLs to background queue

oa-s:
Closes #4456


hboon (approved):
Not quite sure. But let's try. I see it includes a commit from another #4455 though, so have to merge that first.

PR #4459, opened 2022-04-28
Move promise callback to background for EtherscanSingleChainTransactionProvider #4458

### LINKED ISSUE #4458: move promide callback to background EtherscanSingleChainTransactionProvider

oa-s:
Closes #4458

### DIFF
--- a/AlphaWallet/Transactions/EtherscanSingleChainTransactionProvider.swift
+++ b/AlphaWallet/Transactions/EtherscanSingleChainTransactionProvider.swift
@@ -290,13 +290,13 @@ class EtherscanSingleChainTransactionProvider: SingleChainTransactionProvider {
 
             firstly {
                 EtherscanSingleChainTransactionProvider.functional.fetchTransactions(startBlock: startBlock, sortOrder: sortOrder, session: coordinator.session, alphaWalletProvider: coordinator.alphaWalletProvider, tokensDataStore: coordinator.tokensDataStore, queue: coordinator.queue)
-            }.done { [weak self] transactions in
+            }.done(on: queue, { [weak self] transactions in
                 guard let strongSelf = self else { return }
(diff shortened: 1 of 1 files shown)


AlphaWallet/Transactions/EtherscanSingleChainTransactionProvider.swift
```diff
@@ -290,13 +290,13 @@ class EtherscanSingleChainTransactionProvider: SingleChainTransactionProvider {
...
-            }.done { [weak self] transactions in
+            }.done(on: queue, { [weak self] transactions in
                 guard let strongSelf = self else { return }
                 guard !strongSelf.isCancelled else { return }
                 coordinator.addOrUpdate(transactions: transactions)
-            }.catch { e in
+            }).catch(on: queue, { e in
                 error(value: e, rpcServer: coordinator.session.server, address: self.session.account.address)
-            }.finally { [weak self] in
+            }).finally(on: queue, { [weak self] in
```
hboon, line 299: Maybe this should be back on main?

hboon (commented):
Also typo in "promide" :)

AlphaWallet/Transactions/EtherscanSingleChainTransactionProvider.swift
oa-s: since we made ability to update reqlm on synced background queue, we are also able to add objects to realm, on background threads

hboon: Hmm.. wouldn't that be hard to keep track of and introduce bugs? For example line 296:

```
coordinator.addOrUpdate(transactions: transactions)
```

So the body of this function `addOrUpdate(transactions:) and everything it calls might run in the background.

PR #4492, opened 2022-05-03
[Refactor] Rename NFT related files #4491

### LINKED ISSUE #4491: [Refactor] Rename NFT related files

oa-s:
Closes #4491


AlphaWallet/Tokens/Collectibles/Coordinators/NFTAssetSelectionCoordinator.swift
```diff
@@ -51,9 +51,9 @@ class TokenCardSelectionCoordinator: Coordinator {
     }
 }
 
-extension TokenCardSelectionCoordinator: TokenCardSelectionViewControllerDelegate {
+extension NFTAssetSelectionCoordinator: NFTAssetSelectionViewControllerDelegate {
```
hboon, line 54: @oa-s @eviltofu terminology such as "token card" came from TokenScript (for context). But still it's good to rename those. We might bring back the "token card", "token view" etc names when we have better separation for TokenScript

hboon (approved):

PR #4497, opened 2022-05-04
Renamed “Run Script” build phase to “Upload debug symbols to Crashlytics"

eviltofu:
Made the intention of the "Run script" clearer.

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -5429,7 +5429,7 @@
 				8A3C752ACC5CACBD729A8F25 /* [CP] Embed Pods Frameworks */,
 				F6006D07F1DCADBC1150AD02 /* [CP] Copy Pods Resources */,
 				878EE8F6255ADA49000210DE /* Embed App Extensions */,
-				87FBF2B828056D8600616B7D /* ShellScript */,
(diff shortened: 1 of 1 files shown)


eviltofu:
I am thinking that this build phase should be removed. It uploads the debug symbols each time we compile. How will Crashlytics know which is the released debug symbols and which are from the branches developers are working on? No doubt the last upload will be used. We should fold this into the release script of fast lane. @hboon @oa-s

hboon (approved):

hboon:
> I am thinking that this build phase should be removed. It uploads the debug symbols each time we compile. How will Crashlytics know which is the released debug symbols and which are from the branches developers are working on? No doubt the last upload will be used. We should fold this into the release script of fast lane. @hboon @oa-s

Definitely can't move it into fastlane until it works, but even then, I don't think we need to do that. We can just do a check like this:

```
if [ "${CONFIGURATION}" = "Release" ]; then
```

and the developer who wants to upload it while debugging can comment this out.

Also this build phrase (pre-PR) doesn't work. I have to manually upload the dSYMs for release builds.

eviltofu:
> Also this build phrase (pre-PR) doesn't work. I have to manually upload the dSYMs for release builds.

So this should be filed as an issue? @hboon

hboon:
Sure. Would you help assign it to me?

PR #4513, opened 2022-05-05
Move fetching events to background queue #4512

### LINKED ISSUE #4512: Move fetching events to background queue

oa-s:
Closes #4512

### DIFF
--- a/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
+++ b/AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
@@ -37,7 +37,7 @@ final class EventSourceCoordinator: NSObject {
     private func setupWatchingTokenChangesToFetchEvents() {
         tokensDataStore
             .enabledTokenObjectsChangesetPublisher(forServers: enabledServers)
-            .receive(on: RunLoop.main)
+            .receive(on: queue)
             .sink { [weak self] _ in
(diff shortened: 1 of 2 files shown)


AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
```diff
@@ -46,24 +46,31 @@ final class EventSourceCoordinator: NSObject {
     private func setupWatchingTokenScriptFileChangesToFetchEvents() {
         //TODO this is firing twice for each contract. We can be more efficient
         assetDefinitionStore.bodyChange
-            .receive(on: RunLoop.main)
+            .receive(on: queue)
             .compactMap { [weak self] in self?.tokensDataStore.token(forContract: $0) }
```
hboon, line 50: Is the assumption is that `tokensDatastore.store(forContract:)` (and everything else) will always access Realm in the main thread?

hboon (approved):
Merge when ready

AlphaWallet/TokenScriptClient/Coordinators/EventSourceCoordinator.swift
oa-s: actually nope, its able to access to it from any thread, when we access to it from main thread it just doesn't move to sync queue,  and preforms as is. but whe we access to this part from different from main thread it blocks on sync queue, and performs code sync.

PR #4528, opened 2022-05-05
Move map of GetBlockTimestampCoordinator to non main queue #4527

### LINKED ISSUE #4527: Move map of GetBlockTimestampCoordinator to non main queue

oa-s:
Closes #4527

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/GetBlockTimestamp.swift
+++ b/AlphaWallet/Tokens/Coordinators/GetBlockTimestamp.swift
@@ -20,12 +20,11 @@ class GetBlockTimestamp {
 
         let promise: Promise<Date> = firstly {
             web3swift.web3.Eth(provider: web3.provider, web3: web3).getBlockByNumberPromise(blockNumber)
-        }.map {
-            $0.timestamp
-        }
+        }.map(on: web3.requestDispatcher.queue, { $0.timestamp })
 
         cacheForServer[blockNumber] = promise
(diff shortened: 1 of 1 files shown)


AlphaWallet/Tokens/Coordinators/GetBlockTimestampCoordinator.swift
```diff
@@ -20,12 +20,11 @@ class GetBlockTimestampCoordinator {
 
         let promise: Promise<Date> = firstly {
             web3swift.web3.Eth(provider: web3.provider, web3: web3).getBlockByNumberPromise(blockNumber)
-        }.map {
-            $0.timestamp
-        }
+        }.map(on: web3.requestDispatcher.queue, { $0.timestamp })
```
hboon, line 23: Is this necessary? It's always safe to read `.timestamp` on any thread, right?

oa-s: i found that when we got a lot of added wallets the app become freeze for some times, as we got ton of tasks executed on main thread, we need to free it. According to `.timestamp` - yes, it safe to map it on any thread

hboon: Yeah. Then let’s leave it to the default for now. It’s hard to manage spinning work out to non-main threads in multiple places.. Just leave those that are performance critical.

hboon (approved):
Forgot to approve. In case I'm blocking

Issue #4282, opened 2022-04-11
Rename and move logic that is currently called coordinators that don't deal with UI flow


hboon:
Long running task for all of us.

We have many types called coordinators. Coordinators should be restricted to those that control UI flow. We should rename and move (if necessary) the rest slowly as we go along.

I think the naming should be:

* coordinators — control UI flow. Usually a good sign is they import UIKit (other than for access to UIImage)
* service — some of the things that we used to call coordinators that are longer-running
* <name> — these are just logic and non-UI eg. What used to be GetENSOwnerCoordinator, ENSReverseLookupCoordinator.swift, GetENSTextRecordsCoordinator. They are used once and destroyed. They will soon be renamed to ENSResolver, ENSReverseResolver, GetENSTextRecord. Conventional wisdom says they should all be nouns, but I don't think it's easy to come up with a noun that is better than GetENSTextRecord — GetENSTextRecordManager is silly!

Related to: #3990

eviltofu:
How different is this from #3990 ?

hboon:
> How different is this from #3990 ?

#3990 is about extracting logic from real coordinators

This (#4282) is about renaming misnamed coordinators to something more appropriate eg. #4283

eviltofu:
AnalyticsCoordinator does not inherit from Coordinator. This should be renamed AnalyticsService?

eviltofu:
`UniversalLinkCoordinator` does not inherit from `Coordinator` but from `UniversalLinkCoordinatorType`.

hboon:
@eviltofu similar to https://github.com/AlphaWallet/alpha-wallet-ios/issues/3990#issuecomment-1132479898, make a list of checkboxes here.

hboon:
> `AnalyticsCoordinator` does not inherit from `Coordinator` but is a protocol. `AnalyticsCoordinator` should be renamed `Analytics` and concrete implementations should be named `AnalyticsService`.

Yes, but `Analytics` and `MixpanelAnalytics`.

hboon:
> `UniversalLinkCoordinator` does not inherit from `Coordinator` but from `UniversalLinkCoordinatorType`.

Not coordinator, so maybe `UniversalLinkHandler`? But let's not call everything "handlers" (or managers!). Just seems appropriate for this one case.

hboon:
Refer to #6606

PR #4530, opened 2022-05-05
[Refactor] Rename fetch smartcontract data coordinator with provider suffix #4529

### LINKED ISSUE #4529: [Refactor] Rename fetch smartcontract data coordinator with provider suffix

oa-s:
Closes #4529


hboon (commented):
This goes towards #4282 right? Do you think we can just NativeCryptoCurrencyBalanceProvider

Instead of `NativeCryptoCurrencyBalanceProvider`, do you think it's better to call it: `GetNativeCryptoCurrencyBalance`? It's shorter (not a big deal), violates the noun convention (not a big deal I think), but can attach more meaning eg. "GetXXX" fetches some value, "ENSResolver" resolves.

Going by this rationale:

> — these are just logic and non-UI eg. What used to be GetENSOwnerCoordinator, ENSReverseLookupCoordinator.swift, GetENSTextRecordsCoordinator. They are used once and destroyed. They will soon be renamed to ENSResolver, ENSReverseResolver, GetENSTextRecord. Conventional wisdom says they should all be nouns, but I don't think it's easy to come up with a noun that is better than GetENSTextRecord — GetENSTextRecordManager is silly!

Similar for the other renamed types.

Also, remember to move the files out of the `coordinators` directories either in this PR.

oa-s:
@oa-s is this related only to NativeCryptoCurrencyBalanceProvider? or the rest too?

hboon:
@oa-s the rest too. Generally. Let provider bring a fallback when we don’t have a better name.

oa-s:
@hboon pr is ready for review

hboon (approved):

PR #4532, opened 2022-05-05
User shared realm for browser #4531

### LINKED ISSUE #4531: User shared realm for browser
visited links dissapear when change active wallet

oa-s:
Closes #4531


hboon (commented):
Commented in https://github.com/AlphaWallet/alpha-wallet-ios/issues/4531#issuecomment-1119220722

oa-s:
@hboon pr is ready for review

hboon (changes requested):
Looks good, except, can we make `config.hasMigratedToSharedRealm(for:)` (renamed) be a 1 single bool that marks that migration for all the wallets in the device has been done and then a maybe, static in-memory `Dictionary` to hold what `config.hasMigratedToSharedRealm(for:)` previously did?

Because with the current implementation, even after migration, every time the app launches, we still scan through all the URLs.

PR #4535, opened 2022-05-05
Update the Welcome screen #4505

### LINKED ISSUE #4505: Update the Welcome screen
**Problem:**
- we have an outdated logo
- users (tested with developers) have no idea what is "Watch"
- buttons are not aligned and in different width

----
**Solution:**
- update the logo
- remove that top line (thin gray line)
- make two buttons at the bottom in the same width

a) Create a new Wallet
b) I already have a Wallet

Welcome

When you hit b) you will be prompt with options to choose between

* Import existing Wallet 
* Track Wallet Address

Import-Wallet

Zeplin:
https://zpl.io/p1m8vKM
https://zpl.io/noPny1Y
https://zpl.io/XYZe5oK

oa-s:
Closes #4505


hboon (changes requested):
Added a few commits to tweak the button labels. Can you help verify and squash?

We also need to Add the "AlphaWallet" text below the logo. It used to be "Welcome to AlphaWallet". It's missing now.

Merge when ready.

Issue #4565, opened 2022-05-11
`RLMException, reason: String too big`


oa-s:
@hboon do you have any ideas of fixing this issue?

https://console.firebase.google.com/u/0/project/alphawallet-ios-prod/crashlytics/app/ios:com.stormbird.alphawallet/issues/087983306e9dd43e58732fcfac54a7b9?time=last-twenty-four-hours&sessionEventKey=90639cde2eb6426081d6f36049c5d468_1674795446825436627

oa-s:
@hboon does it means that we trying to store single string that is larger then 16mb?

hboon:
Looking… Strange that it says it's a regression from v3.40 (427) and was closed. I checked the closed issues and it's not there.

<img width="891" alt="Screenshot 2022-05-11 at 3 58 27 PM" src="https://user-images.githubusercontent.com/56189/167799007-f898e70c-1a9c-4a41-aa8f-949c564547aa.png">

oa-s:
it couble be good to find our that exactly is large for storing, added pr to track users selected wallet address, for later watch ing and trying to reproduce

hboon:
OK

oa-s:
we also can track json strings that is greater than 16 mb, and skip such balances, but i don't think that its a good idea.

oa-s:
it might happen, when json contains base64 data for images, that might  happen quite ofter

hboon:
>@hboon i have suggestion about this issue, what if we update our code base to save balance json as file, like we do for erc1155, and store file url in realm instead whole json string?

@oa-s hmm. But wondering if that will end up with many loose files. Is there a Blob type? What if we convert to Data when we store JSON strings? I suppose the Realm size limitation is because of indexing performance rather than storage?

hboon:
> not sure, as there same limitation for String, Data types, max file size is 16 mb

I see. Since we are guessing that it's images that are base64 encoded in the JSON, do you want to do a size check for them and log if it exceeds a certain size? (say 2MB?) to Firebase? Then if it crashes we ought to pick it up. Maybe log these:

* URL where we picked up the JSON from
* Contract + chain
* Token ID

?

PR #4567, opened 2022-05-11
Add tracking active users wallet address for firebase reports #4566

### LINKED ISSUE #4566: Add tracking active users wallet address for firebase reports

oa-s:
Closes #4566

### DIFF
--- a/AlphaWallet/Core/Helpers/FirebaseReportService.swift
+++ b/AlphaWallet/Core/Helpers/FirebaseReportService.swift
@@ -8,7 +8,7 @@
 import Firebase
 
 extension AlphaWallet {
-    class FirebaseReportService: ReportService {
+    final class FirebaseReportService: ReportService {
         private let options: FirebaseOptions
         // NOTE: failable initializer allow us easily configure with different plist files for different configurations of project
         init?(contents: String? = Constants.googleServiceInfoPlistContent) {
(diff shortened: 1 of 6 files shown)


hboon (commented):
Not sure this is a good idea due to privacy. Is this so we can reproduce it with the right wallet contents?

hboon:
I thought of checking for a hardcoded list of wallets to see if it's present with a key so we can tag Crashlytics with just a few of our team members but not actual users. But it might not cover what you intended with this PR.

AlphaWallet/Core/Helpers/FirebaseReportService.swift
```diff
@@ -22,5 +26,14 @@ extension AlphaWallet {
         func configure() {
             FirebaseApp.configure(options: options)
         }
+
+        static func track(wallet: Wallet) {
+            let keysAndValues: [String: Any] = [
+                ReportKey.walletAddress.rawValue: wallet.address.eip55String,
```
hboon, line 32: Could you track if it's a "real" or "watch" too?

hboon (approved):
Oh well since we aren't correlating user data, I guess it's for the benefit of users https://github.com/AlphaWallet/alpha-wallet-ios/issues/4565#issuecomment-1123331466

PR #4622, opened 2022-05-18
Resolve realm memory leaks #4621

### LINKED ISSUE #4621: Resolve realm memory leaks
For some reasons creating a copy of realm objects causes memory leaks. Especially with `Balance` instances.

oa-s:
Closes #4621


hboon (approved):
I see some refactoring/renaming. But what was the problem?

oa-s:
We got here a memory leak issue, when we make a new copy of realm object (like we did with detached() ) the instances don't get deleted. Even if we don't have manual strong ref to them. Is especially related to token, and tokenBalance. After passing some time, after apps launch, there is more than 10k Token balances, and from 1-2k token objects.

oa-s:
Renaming it's mostly related of what we passing, token as(activity.token) or tolenObject

oa-s:
Guess that passing structs peer the app should be better

hboon (approved):
OK. Trust you tested :)

PR #4636, opened 2022-05-20
Tweak pod url to fix memory leak for [EthereumABI.pod] #4635

### LINKED ISSUE #4616: [EthereumABI] Memory leak for NSRegularExpression
@hboon  can you please make a fork from EthereumABI, so we would be able to make changes.
creating an instance of **NSRegularExpression** for every function call a lot of non released resources.

### LINKED ISSUE #4635: Tweak pod url to fix memory leak for [EthereumABI.pod]

oa-s:
Closes #4635, Closes #4616

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -29,7 +29,7 @@ target 'AlphaWallet' do
   pod 'TrustWalletCore', '2.6.34'
   pod 'AWSSNS', '2.18.0'
   pod 'Mixpanel-swift', '~> 3.1'
-  pod 'EthereumABI', '1.3.0'
+  pod 'EthereumABI', :git => 'git@github.com:AlphaWallet/EthereumABI.git', :commit => '877b77e8e7cbc54ab0712d509b74fec21b79d1bb'
   pod 'BlockiesSwift'
   pod 'PaperTrailLumberjack/Swift'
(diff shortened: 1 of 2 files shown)


Podfile
```diff
@@ -29,7 +29,7 @@ target 'AlphaWallet' do
   pod 'TrustWalletCore', '2.6.34'
   pod 'AWSSNS', '2.18.0'
   pod 'Mixpanel-swift', '~> 3.1'
-  pod 'EthereumABI', '1.3.0'
+  pod 'EthereumABI', :git => 'git@github.com:AlphaWallet/EthereumABI.git'
```
hboon, line 32: Better to include the commit hash too, like we do for `KeychainSwift`, etc.

hboon (approved):

Podfile
hboon: Oh, but maybe you didn't have it when you wrote this :)

hboon:
Why would it be a memory leak though? Are the regex retained? I thought it would just be slower at runtime due to the regex compilation.

oa-s:
i am not quite sure where actually its retained, as its allocated in single function, and uses only there, but what i am facing im memory graph is that there is more than 16k objects of RegexExp. i remember that we already done such changes with the rest of regexes, that time there were more that 50k instances.

oa-s:
and in summary it caused 11 mb memory leak

PR #4639, opened 2022-05-22
Updated Add PHI Network

Phinetwork:
Added PHI Network Smart Chain


hboon:
@Phinetwork Thanks for the contribution. We build chains.zip from https://chainid.network/chains.json which is sourced from https://chainlist.org. The data is available at https://github.com/ethereum-lists/chains, under _data/chains. So the way to do this is to make a PR there and we can update accordingly here.

Alternatively you can add it to https://github.com/AlphaWallet/alpha-wallet-ios/blob/19add6206d5d8a96ab2d95ee06be4cffeb3fa6c4/AlphaWallet/Settings/Types/RPCServers.swift which would involve Swift native code across multiple files. Then the app would come bundled with the chain (chains.zip just provides an easy way for the user to search for the chain in the app and add it)

Phinetwork:
ok Thank you we have done so waiting for chainlist to update t

hboon:
Superseded by #4881

PR #4644, opened 2022-05-22
Changed url from git@github.com:AlphaWallet/EthereumABI.git to https://github.com.AlphaWallet/EthereumABI.git in Podfile.

eviltofu:
Fix pod install for Github non-SSH key developers.


hboon (approved):
Didn't test. But can you make the commit message and PR title be:

"Fix `pod install` for non-AlphaWallet developers"

The reason or summary is usually more important.

eviltofu:
It happens to me too. Basically if you are not using SSH keys with GitHub this will happen. So reason should be 
"Fix `pod install` for Github non-SSH key developers." @hboon

PR #4679, opened 2022-05-26
Add searching in nfts json by keyword #4658

### LINKED ISSUE #4658: Searching for name of NFT should match if it's available in tokenURI() even if not in `name` contract function
Screenshot 2022-05-25 at 11 12 01 AM

Eg. https://resources.carlachan.com/1/0x28d5b2e6f30a9b54dcbe8792543fb0232f4f3658/27.json

1. Search for "Space" in Wallet tab

Expected:

2. Token appear

Observed:

2. Token doesn't appear

oa-s:
Closes #4658

### DIFF
--- a/AlphaWallet/Core/Coordinators/TokensFilter.swift
+++ b/AlphaWallet/Core/Coordinators/TokensFilter.swift
@@ -8,8 +8,40 @@
 import Foundation
 
 class TokensFilter {
-    private enum FilterKeys: String {
+    private enum FilterKeys {
+        case all
         case swap
+        case erc20
+        case erc721
+        case erc875
+        case erc1155
+        case development(Development)
+        case tokenscript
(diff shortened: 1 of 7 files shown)


hboon (commented):
Does this search the `description` field too? Can we limit it only to `name` for the JSON? Or is it too inefficient?

Because searching for "space" brings up https://opensea.io/collection/animetas (my wallet). I wonder if there'll be lots more search noise for basic words

oa-s:
I guess we can, for now we are searching in whole json

hboon:
I suppose we should already have those in the `values` dictionary maybe? If it's too cumbersome, never mind.

hboon (approved):
Right. Ok, if you are satisified, let's merge this after fixing the tests.

PR #4696, opened 2022-05-27
Add object for representing additional token info TokenInfoObject #4695

### LINKED ISSUE #4695: Add object for representing additional token info `TokenInfoObject`
Due to changes maded in #4662 it would be useful to keen such kind of information in inner object

```
class TokenInfoObject: Object {
    @objc dynamic var uid: String = ""
    @objc dynamic var coinGeckoId: String? 
    @objc dynamic var imageUrl: String?
} 
```

```
class TokenObject: Object {
...
@objc dynamic var _info: TokenInfoObject?
...
}
```

Fields can be changed with later needs. An additional model needs to decrease TokenObject and not overload it with fields

oa-s:
Closes #4695


AlphaWallet/Tokens/Types/TokenObject.swift
```diff
@@ -111,3 +118,51 @@ class TokenObject: Object {
         return .init(chainID: chainId)
     }
 }
+
+class TokenInfoObject: Object {
+    @objc dynamic var uid: String = ""
+    @objc dynamic var coinGeckoId: String?
```
hboon, line 124: Are these just 2 values that are used to store inside `TokenObject` in the database? Why do we need them now (and not before)?

oa-s: nope, not only these two. there might be fields related to `Kleros Curate Service`, tokenImage and any other .

hboon: > And not storing them directly in `TokenObject` is just for organization purpose right? Not because there is some kind of sharing (eg. 2 `TokenObject` refers to the same `TokenInfoObject` instance)?

Forgot to send out this one earlier. Sorry.

hboon: > nope, not only these two. there might be fields related to `Kleros Curate Service`, tokenImage and any other .

And `TokenObject` is always 1:1 with `TokenInfoObject`?

oa-s: > And TokenObject is always 1:1 with TokenInfoObject?

yes, TokenInfoObject creates only one time for each token object

hboon:
Ok. Sound good :)

PR #4705, opened 2022-05-30
Add support for WalletConnect v2 SDK beta #4669

### LINKED ISSUE #4669: Upgrade (try out?) to WalletConnect v2 SDK beta
When it's available

oa-s:
Closes #4669


AlphaWallet/Core/Types/Session.swift
```diff
@@ -41,6 +41,12 @@ class WalletSession {
     }
 }
 
+extension WalletSession: {
+    var capi10Account: CAIP10Account {
```
hboon, line 45: capi10Account -> caip10Account :)

hboon (commented):
Doesn't build on Travis and locally after clearing cache. Is it a build caching issue?

Is forking the SDK no longer needed?

Only did a quick check, what is `WalletConnectV2Storage` for and why is it added (and not earlier)?

Any feedback for the SDK folks?

---

And for our own purposes, what do you think of using publishers to replace delegates, like this:

```
Sign.instance.sessionRequestPublisher
```

AlphaWallet/Core/Types/Session.swift
oa-s: > WalletConnectV2Storage 

it is code parts that were in WalletConnectV2Provider, just extracted to separate storage.

oa-s: i guess the same for v1 is needed

oa-s: > And for our own purposes, what do you think of using publishers to replace delegates, like this:

i guess it's great feature, we can move forward to use the same approach somewhere. (but not really sure), all our code base is based on delegate pattern, is it good idea to refactor all of these?

oa-s:
@hboon pr is ready for review, tested v2 works for me.

AlphaWallet/Core/Types/Session.swift
hboon: Hold on to delegate for a while. I'm a bit worried it's harder to find (it's actually just 1 more place to look for, not so bad). At least it's not necessary yet, right?

hboon (approved):
Didn't test it. Merge it in then. Since it's not prod yet.

hboon:
> @hboon can you please try to build it for you? it works for me, but not for Travis.

Oh, works for me now. Strange. I have cleared Travi's cache and trying to re-run it, but you can merge if you are ready.

PR #4719, opened 2022-05-31
Remove console warning maximumViewportInset cannot be larger than frame #4718

### LINKED ISSUE #4718: Remove console warning `maximumViewportInset cannot be larger than frame`

oa-s:
Closes #4718

### DIFF
--- a/AlphaWallet/Browser/ViewControllers/BrowserViewController.swift
+++ b/AlphaWallet/Browser/ViewControllers/BrowserViewController.swift
@@ -43,7 +43,7 @@ final class BrowserViewController: UIViewController {
 
     lazy var webView: WKWebView = {
         let webView = WKWebView(
-            frame: .zero,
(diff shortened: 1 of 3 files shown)


AlphaWallet/Browser/ViewControllers/BrowserViewController.swift
```diff
@@ -43,7 +43,7 @@ final class BrowserViewController: UIViewController {
 
     lazy var webView: WKWebView = {
         let webView = WKWebView(
-            frame: .zero,
+            frame: .init(x: 0, y: 0, width: 40, height: 40),
```
hboon, line 46: @oa-s what does the `40` imply?

oa-s: there is could be any non zero value, 40 as its our min image size we using

PR #4747, opened 2022-06-03
[Refactor] Access fonts via Fonts.regular(size:) etc instead of explicitly referring to the font family/name

hboon:
Closes #4746

PR #4748, opened 2022-06-03
Switch app-wide font to Circular

hboon:
Trying it out on TestFlight first. Build 431.

Depends on #4747

Before|After
-|-
Screenshot 2022-06-03 at 2 14 06 PM|Screenshot 2022-06-03 at 1 54 54 PM

Address and amount in textboxes look alright:

Screenshot 2022-06-03 at 2 15 48 PM


oa-s (approved):

hboon:
Force pushed, including:

* rebased master so italic and light weights are replaced with regular
* Source Sans Pro semibold - > Circular medium

hboon:
Closed since we are sticking to Source Sans Pro for now.

PR #4765, opened 2022-06-03
Define realm database related classes for wallet db #4764

### LINKED ISSUE #4764: Define realm database related classes for wallet db

oa-s:
Closes #4764


hboon (approved):
What happens if we don't set it? :) We didn't

oa-s:
in general when we update realm models, that is not related to general realm file, we willn't need to perform migtation for them. for example: if we make some changes in Bookmark model, we also will be needed to update bookmarks in wallet based realm.

PR #4771, opened 2022-06-03
Extract more logic from coordinator, move token importing to separate resolver #4770

### LINKED ISSUE #4770: Extract more logic from coordinator, move token importing to separate resolver

oa-s:
Closes #4770


hboon (approved):
Didn't test.

PR #4793, opened 2022-06-06
Dispay domain.name, and domain.verifyingContract if them non empty or non nill #4744

### LINKED ISSUE #4744: Improve display of domain.name in EIP712 v3 from WalletConnect
Not sure if this if we sign from the dapp browser without involving WalletConnect, but a reproducible example with WalletConnect:

1. In mobile Safari (definitely not dapp browser), visit https://attestation-foxgem.vercel.app
2. Enter email
3. Copy and paste confirmation code received in email
4. Accept WalletConect connection
5. In signing request, the URL (the row after "Network" in the screenshot below) appears as the field name (i.e. on the left column) instead of as the value

Screenshot 2022-06-02 at 11 04 02 AM

Also after tapping on "Show" for the URL, the copy button is clipped off. This screen looks OK when there's a field name (eg. if we had tap on the "Show button" for "payload" instead)

Screenshot 2022-06-02 at 11 04 00 AM

It looks like the EIP712 payload (seems to be `.signTypedMessageV3`) looks like this:

```
{
   "domain" : {
      "name" : "http://wwww.attestation.id",
      "version" : "0.1"
   },
   "message" : {
      "description" : "Creating email attestation",
      "identifier" : "hboon@motionobj.com",
      "payload" : "72f8d41ab26751f57fef3615a58c6015d3fb6597081263bcf2a863f510806a8e",
      "timestamp" : "Fri Jun 03 2022 12:08:26 GMT+0800"
   },
   "primaryType" : "AttestationRequest",
   "types" : {
      "AttestationRequest" : [
         {
            "name" : "payload",
            "type" : "string"
         },
         {
            "name" : "description",
            "type" : "string"
         },
         {
            "name" : "timestamp",
            "type" : "string"
         },
         {
            "name" : "identifier",
            "type" : "string"
         }
      ],
      "EIP712Domain" : [
         {
            "name" : "name",
            "type" : "string"
         },
         {
            "name" : "version",
            "type" : "string"
         }
      ]
   }
}

```
So maybe that URL is `domain.name`?
```

### PR DESCRIPTION
**Pull request by oa-s**
Closes #4744

### DIFF
--- a/AlphaWallet/Transfer/ViewModels/SignatureConfirmationConfirmationViewModel.swift
+++ b/AlphaWallet/Transfer/ViewModels/SignatureConfirmationConfirmationViewModel.swift
@@ -132,14 +132,27 @@ enum SignatureConfirmationViewModel {
     }
 
     struct EIP712TypedDataConfirmationViewModel {
-        var values: [(key: String, value: EIP712TypedData.JSON)]
+        typealias EIP712TypedDataToKey = (key: String, value: EIP712TypedData.JSON)
+        let values: [EIP712TypedDataToKey]
         let walletConnectDappRequesterViewModel: WalletConnectDappRequesterViewModel?
 
         init(data: EIP712TypedData, walletConnectDappRequesterViewModel: WalletConnectDappRequesterViewModel?) {
             self.walletConnectDappRequesterViewModel = walletConnectDappRequesterViewModel
-            values = [
-                (key: data.domainName, value: .string(data.domainVerifyingContract?.truncateMiddle ?? ""))
-            ] + data.message.keyValueRepresentationToFirstLevel
+
+            var _values: [EIP712TypedDataToKey] = []
+            
+            if let verifyingContract = data.domainVerifyingContract, data.domainName.nonEmpty {
+                _values += [(key: data.domainName, value: .string(verifyingContract.truncateMiddle)) ]
+            } else if let verifyingContract = data.domainVerifyingContract {
+                _values += [(key: "domain.verifyingContract", value: .string(verifyingContract.truncateMiddle)) ]
+            } else if data.domainName.nonEmpty {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: AlphaWallet/Transfer/ViewModels/SignatureConfirmationConfirmationViewModel.swift [hunk 1]
```diff
@@ -132,14 +132,28 @@ enum SignatureConfirmationViewModel {
...
 
         init(data: EIP712TypedData, walletConnectDappRequesterViewModel: WalletConnectDappRequesterViewModel?) {
             self.walletConnectDappRequesterViewModel = walletConnectDappRequesterViewModel
-            values = [
-                (key: data.domainName, value: .string(data.domainVerifyingContract?.truncateMiddle ?? ""))
-            ] + data.message.keyValueRepresentationToFirstLevel
+
+            var _values: [EIP712TypedDataToKey] = []
+            
+            if data.domainName.nonEmpty {
```
hboon, line 144: I think there should be one more check if both are nonEmpty? The original:

```
(key: data.domainName, value: .string(data.domainVerifyingContract?.truncateMiddle ?? ""))
```

oa-s: so we don't need to display domain name if there is no  `domainVerifyingContract`, right? not displaying them separatelly?

hboon: Sorry, it's my pseudo code. How about this?

```
if notEmpty(domain.verifyigContract) && notEmpty(domain.name)
    "\(domain.name): \(domain.verifyingContract)"
elsif notEmpty(domain.verifyigContract)
    "domain.verifyingContract: \(domain.verifyingContract)"
elsif notEmpty(domain.name)
    "domain.name: \(domain.name)"
end
```

PR #4803, opened 2022-06-06
ENS name is not displaying in the change wallet page #4726

### LINKED ISSUE #4726: ENS name is not displaying in the change wallet page
v3.42 build 430

Victor

oa-s:
Closes #4726


hboon (commented):
What was the reason it didn't work (sometimes?) ?

oa-s:
> What was the reason it didn't work (sometimes?) ?

cells view model get dellocated right before initialization

hboon:
> > What was the reason it didn't work (sometimes?) ?
> 
> cells view model get dellocated right before initialization

OK. There seems to be quite a lot of unrelated changes in the PR, renaming, etc. Hard to figure that out.

PR #4805, opened 2022-06-06
Add local storage for ens/ens-record/address resolutions #4804

### LINKED ISSUE #4804: Add local storage for ens/ens-record/address resolutions

oa-s:
Closes #4804


hboon (commented):
I can't seem to build it. Maybe the project file is broken?

But this is caching ENS forward and reverse lookups right? How is expiry handled?

oa-s:
@hboon  yes, and ens text records too

hboon:
@oa-s how do we deal with expiry (i.e. when do we refresh the cache containing`something.eth` -> `0xb0b`)? Can you point me to it?

oa-s:
there is no expiry for now, it works like persistent storage

hboon:
Is it in-memory? If not, can we make it be in-memory and not store on disk? Reason being ENS (name) -> address is very vulnerable if the ENS registration expires or if the user intentionally change to a new address

oa-s:
it stores in persisnt store. i got your idea, will make another pr for that

oa-s:
Updated pr with expiring ens records

hboon:
Can't seem to build it. At least `EnsRecordsStorage.swift` isn't in the included in the project

hboon:
Deleted comment. Never mind, my bad.

AlphaWallet/Settings/Types/Constants.swift
```diff
@@ -226,6 +226,10 @@ public struct Constants {
         static let connectionTimeout: TimeInterval = 10
     }
 
+    enum Ens {
+        static let recordExpiration: TimeInterval = -(7 * 24 * 60 * 60)
```
hboon, line 230: I think this should be positive, instead of negative? Because: `NSDate(timeIntervalSinceNow: expirationTime)`

AlphaWallet/Core/ENS/Storage/EnsRecordsStorage.swift
```diff
@@ -0,0 +1,99 @@
...
+    let storage: EnsRecordsStorage = RealmStore.shared
+    return storage
+}()
+
+protocol EnsRecordsStorage: AnyObject {
+    var recordCount: Int { get }
+
+    func record(for key: EnsLookupKey, expirationTime: TimeInterval) -> EnsRecord?
+    func addOrUpdate(record: EnsRecord)
+    func removeRecord(for key: EnsLookupKey)
```
hboon, line 23: Seems unused?

AlphaWallet/Core/Helpers/BlockieGenerator.swift
```diff
@@ -85,24 +91,27 @@ class BlockiesGenerator {
...
-                return .init(error: TokenImageFetcher.ImageAvailabilityError.notAvailable)
+                guard let url = URL(string: url) else { return .init(error: AnyError.blockieCreateFailure) }
+                return .value(.url(url: WebImageURL(url: url, rewriteGoogleContentSizeUrl: .s120), isEnsAvatar: true))
             }
+
             return firstly {
                 OpenSea.fetchAssetImageUrl(for: result, server: .main)
+            }.get { url in //NOTE: cache fetched open sea image url and rewrite ens avatar with new image
+                let key = EnsLookupKey(nameOrAddress: ens ?? address.eip55String, server: .forResolvingEns, record: .avatar)
+                self.storage.addOrUpdate(record: .init(key: key, value: .record(url.absoluteString)))
```
hboon, line 111: Is there a read for this cached value?

hboon (approved):
I think I might have suggested 7 days somewhere, but can we change it to 2 days instead? But still a short term approach, we probably need to figure out how to expire better (or maybe use the the cached record regardless of expiry and then fetch the real one again, but we'll see next time)

hboon:
Good to merge when you are done

AlphaWallet/Core/Types/RealmStore.swift
```diff
@@ -55,8 +59,13 @@ extension Realm {
         var configuration = RealmConfiguration.configuration(name: name)
         configuration.objectTypes = [
             Bookmark.self,
-            History.self
+            History.self,
+            EnsRecordObject.self
         ]
+        configuration.schemaVersion = 2
+        configuration.migrationBlock = { migration, oldSchemaVersion in
+            migration.deleteData(forType: EnsRecordObject.className())
```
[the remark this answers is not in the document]
oa-s: that was need for my local migration, will remove it

AlphaWallet/Core/Helpers/BlockieGenerator.swift
```diff
@@ -85,24 +91,27 @@ class BlockiesGenerator {
...
+            promise = ensTextRecordFetcher.getENSRecord(forName: ens, record: .avatar)
         } else {
-            promise = GetENSTextRecord(server: .forResolvingEns)
-                .getENSRecord(forAddress: address, record: .avatar)
+            promise = ensTextRecordFetcher.getENSRecord(forAddress: address, record: .avatar)
         }
 
         return firstly {
             promise
         }.then { url -> Promise<BlockiesImage> in
```
oa-s, line 101: on next call the url will be here, so we don't need to fetch image from open sea again

AlphaWallet/Settings/Types/Constants.swift
oa-s: since we want to check if date has expired, we get now and subtract `n` timeinterval to figure out `7 day ago`, and later check, if creation date is greater than our date

AlphaWallet/Core/Helpers/BlockieGenerator.swift
oa-s: ` }.then { url -> Promise<BlockiesImage> in`

hboon: Is this responding to my comment at https://github.com/AlphaWallet/alpha-wallet-ios/pull/4805#discussion_r894139983?

I mean `fetchEnsAvatar()` has this to cache the result, but is that cached value ever read from the cache again?

```
let key = EnsLookupKey(nameOrAddress: ens ?? address.eip55String, server: .forResolvingEns, record: .avatar)
self.storage.addOrUpdate(record: .init(key: key, value: .record(url.absoluteString)))
```

hboon: i.e.

```
if isCached
    use_cached //I can't find this
else
    fetch
    cache //it does this
end
```

oa-s: we update it as ens record `avatar`, so it going to be returned in GetEnsTextRecord

AlphaWallet/Settings/Types/Constants.swift
hboon: Oh, Ok.

PR #4811, opened 2022-06-07
Fix crash because dictionary property in `AssetAttributesCache` is accessed in both main and non-main threads concurrently #4810

### LINKED ISSUE #4810: Crash because `AssetAttributesCacheData` (and its `contracts` property) is accessed from both main and other threads
https://console.firebase.google.com/u/0/project/alphawallet-ios-prod/crashlytics/app/ios:com.stormbird.alphawallet/issues/b4d18e4fca5eaeea7e6ae168945bb1c4?time=last-seven-days&versions=3.43%20(431)&sessionEventKey=9a5f34714ad64310995bad3f4ae531eb_1683421825567050960

hboon:
Fixes #4810

`AssetAttributesCacheData` no longer implements `Codable`. It's harmless since we haven't implemented persistance to disk.

I look at switching to `AtomicDictionary ` as a short-term solution though. Refer to: https://github.com/AlphaWallet/alpha-wallet-ios/issues/4810#issuecomment-1148208721

### DIFF
--- a/AlphaWallet/TokenScriptClient/Models/AssetAttributesCache.swift
+++ b/AlphaWallet/TokenScriptClient/Models/AssetAttributesCache.swift
@@ -9,11 +9,10 @@ class AssetAttributesCache {
     private var resolvedAttributesData: AssetAttributesCacheData
     private var functionOriginAttributes: [AlphaWallet.Address: [AttributeId: AssetAttribute]] = .init()
     private var functionOriginSubscribables: [AlphaWallet.Address: [TokenId: [AttributeId: Subscribable<AssetInternalValue>]]] = .init()
-    
+
     init() {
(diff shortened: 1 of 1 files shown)


oa-s (approved):

oa-s:
there also might be usefull to update 
private var functionOriginAttributes: [AlphaWallet.Address: [AttributeId: AssetAttribute]] = .init()
    private var functionOriginSubscribables: [AlphaWallet.Address: [TokenId: [AttributeId: Subscribable<AssetInternalValue>]]] = .init()

with atomic dictionary

hboon:
@oa-s oh shucks. What about `AssetAttributesCache.functionOriginSubscribables`

PR #4815, opened 2022-06-07
Move ens to background queue  #4814

### LINKED ISSUE #4814: Move ens to background queue

oa-s:
Closes #4814


hboon (approved):
An observation: 

Is it necessary for ENS? I'm wary about introducing little bits of work in background threads/queues in many places. It's too easy writing code to forget to make sure sure we are using the right queue and at runtime, hard to figure out where things might be failing. Ideally code should just work on the current thread and most should be on the main thread.

An example when writing code is in the snippet below, we now need to make sure `self.storage.addOrUpdate(value:)` works correctly, either by passing in `queue` (and using it correctly) or `addOrUpdate(value:)` itself always does the right thing on its own queue (which it does)

```
return firstly {
        ENS(delegate: self, chainId: server.chainID, queue: queue).getENSAddress(fromName: name)
    }.get(on: queue, { address in
        let key = EnsLookupKey(nameOrAddress: name, server: self.server)
        self.storage.addOrUpdate(value: .address(address), for: key)
    })
}
```

Just something to be wary of?

hboon:
@oa-s need to fix tests before merging

oa-s:
yes, i agree with you in it, maded these changes as i'v facing long running operation when fetching ens, especially when we make calls of functions `getName(fromAddress` `getTextRecord(forName`, for ens, so to drop load on main thread, moved then background queue

hboon:
> yes, i agree with you in it, maded these changes as i'v facing long running operation when fetching ens, especially when we make calls of functions `getName(fromAddress` `getTextRecord(forName`, for ens, so to drop load on main thread, moved then background queue

Yeah, sure we can fix it like we are now in bit and pieces so it works, but longer term we'll come up with something better :)

PR #4817, opened 2022-06-07
Move write realm calls to background queue #4816

### LINKED ISSUE #4816: App freezes when change servers

oa-s:
Closes #4816

### DIFF
--- a/AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
+++ b/AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
@@ -155,7 +155,9 @@ class WalletBalanceFetcher: NSObject, WalletBalanceFetcherType {
 
         for each in servers {
             //NOTE: when we change servers it might happen the case when native
-            tokensDataStore.addEthToken(forServer: each)
+            queue.async {
+                self.tokensDataStore.addEthToken(forServer: each)
+            }
             getOrCreateBalanceFetcher(server: each)
         }
 
--- a/AlphaWallet/Tokens/Types/TokensDataStore.swift
+++ b/AlphaWallet/Tokens/Types/TokensDataStore.swift
@@ -79,12 +79,13 @@ class MultipleChainsTokensDataStore: NSObject, TokensDataStore {
     init(store: RealmStore, servers: [RPCServer]) {
         self.store = store
         super.init()
-
-        for each in servers {
-            addEthToken(forServer: each)
+        //NOTE: async on queue to avoid `_dispatch_sema4_wait` that blocks mainthread on begin transaction
+        DispatchQueue.global().async {
+            for each in servers {
+                self.addEthToken(forServer: each)
+            }
+            MultipleChainsTokensDataStore.functional.recreateMissingInfoTokenObjects(for: store)
         }
-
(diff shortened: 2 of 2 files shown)


hboon:
> @hboon i started facing strange issue when change servers, ui freeses for me. looks like its because of `_dispatch_sema4_wait` that get called when write to realm on main queue. can u see the same behaviour?

Before this PR? No, I tried a few times on the simulator and don't encounter it.

hboon:
Could it be related to the `Thread.sleep(forTimeInterval: 0.002)` in `RunLoopThread.performSync(::)`?

AlphaWallet/Core/Coordinators/WalletBalance/WalletBalanceFetcher.swift
```diff
@@ -155,7 +155,9 @@ class WalletBalanceFetcher: NSObject, WalletBalanceFetcherType {
 
         for each in servers {
             //NOTE: when we change servers it might happen the case when native
-            tokensDataStore.addEthToken(forServer: each)
+            queue.async {
```
hboon, line 158: I take this one as a temporary fix? :) We should definitely not be doing this in the long run.

AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -79,12 +79,13 @@ class MultipleChainsTokensDataStore: NSObject, TokensDataStore {
     init(store: RealmStore, servers: [RPCServer]) {
         self.store = store
         super.init()
-
-        for each in servers {
-            addEthToken(forServer: each)
+        //NOTE: async on queue to avoid `_dispatch_sema4_wait` that blocks mainthread on begin transaction
+        DispatchQueue.global().async {
```
hboon, line 83: This too.

hboon (approved):

oa-s:
> Could it be related to the `Thread.sleep(forTimeInterval: 0.002)` in `RunLoopThread.performSync(::)`?

it shouldn't, as operation performs on main queue, there is a check so we don't block it
```
func performSync(_ callback: @escaping (Realm) -> Void) {
        if Thread.isMainThread {
            callback(mainThreadRealm)
        } else {
            thread.performSync() {
                guard let realm = self.backgroundThreadRealm else { fatalError("Failure to resolve background realm") }
                callback(realm)
            }
        }
    }
```

hboon:
If it's always on main, then `callback()` would be invoked immediately one after another with the same `mainThreadRealm` value and `_dispatch_sema4_wait` shouldn't be triggered right? Is there a possibility `MultipleChainsTokensDataStore.init(store:servers:)` don't get called on the main thread? For instance, it's store in `ActiveWalletCoordinator.tokensDataStore` which is lazily initiated. Maybe sometimes?

hboon:
But let's merge this if it makes the problem on your side and we'll see.

oa-s:
> But let's merge this if it makes the problem on your side and we'll see.



> If it's always on main, then `callback()` would be invoked immediately one after another with the same `mainThreadRealm` value and `_dispatch_sema4_wait` shouldn't be triggered right? Is there a possibility `MultipleChainsTokensDataStore.init(store:servers:)` don't get called on the main thread? For instance, it's store in `ActiveWalletCoordinator.tokensDataStore` which is lazily initiated. Maybe sometimes?

hmm, let me see, maybe

PR #4822, opened 2022-06-08
Remove WithReusableIdentifier from UICollectionViewCell as conformance is set somewhere else.

eviltofu:
```
/alpha-wallet-ios/AlphaWallet/Extensions/UITableView.swift:26:33: conformance of 'UICollectionViewCell' to protocol 'WithReusableIdentifier' was already stated in the type's module 'UIKit'
extension UICollectionViewCell: WithReusableIdentifier {
```
Pull request to silence the above warning.


hboon (approved):
Ah, I had always thought it's because there's a nameclash. Turns out it's because `UICollectionViewCell` is a subclass of`UICollectionReusableView`

PR #4830, opened 2022-06-08
Added dark mode to initial wallet creation screen

eviltofu:
iwcs

Delete current app to remove wallet data. Check to see if this code style is acceptable. Once approved, I will fix the info.plist flag to change `appearance` to `light` before merging. Also are the colours right? I went by the style guide. Also Dusty is defined as RGB 153, 153, 153 in the colour guide but the style guide had 151, 151, 151. I followed the colour guide.

Simulator Screen Shot - iPhone 11 - 2022-06-08 at 21 12 40 copy Simulator Screen Shot - iPhone 11 - 2022-06-08 at 21 13 00 copy

@hboon @colourfreak

### DIFF
--- a/AlphaWallet/Configuration.swift
+++ b/AlphaWallet/Configuration.swift
@@ -1,0 +1,100 @@
+//
+//  Configuration.swift
+//  AlphaWallet
+//
+//  Created by Jerome Chan on 30/5/22.
+//
+
+import UIKit
+
+fileprivate func colorFrom(trait: UITraitCollection, lightColor: UIColor, darkColor: UIColor) -> UIColor {
+    switch trait.userInterfaceStyle {
+    case .unspecified, .light:
+        return lightColor
+    case .dark:
+        return darkColor
+    }
+}
+
+struct Configuration {
+    struct Color {
+        struct Semantic {
+            static let defaultViewBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.cod()!)
+            }
+            static let defaultForegroundText = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.white()!)
+            }
+
+            static let primaryButtonBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.cod()!, darkColor: R.color.alabaster()!)
+            }
+            static let primaryButtonHighlightedBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.concrete()!)
+            }
+            static let primaryButtonBackgroundInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.mike()!, darkColor: R.color.mine()!)
+            }
+            static let primaryButtonBorderInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.mike()!, darkColor: R.color.mine()!)
+            }
+            static let primaryButtonTextActive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.black()!)
+            }
+            static let primaryButtonTextInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.black()!)
+            }
+            static let primaryButtonBorderActive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.cod()!, darkColor: R.color.alabaster()!)
+            }
+
+            static let secondaryButtonBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.dusty()!)
+            }
+            static let secondaryButtonHighlightedBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.concrete()!, darkColor: R.color.dusty()!)
+            }
+            static let secondaryButtonBackgroundInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.mike()!, darkColor: R.color.dusty()!)
+            }
+            static let secondaryButtonBorderInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.mike()!, darkColor: R.color.dusty()!)
+            }
+            static let secondaryButtonTextActive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.white()!)
+            }
+            static let secondaryButtonTextInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.white()!)
+            }
+            static let secondaryButtonBorderActive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.dusty()!)
+            }
+        }
+    }
+}
+
+class UIKitFactory {
+
+    static func defaultView(autoResizingMarkIntoConstraints: Bool = false) -> UIView {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = autoResizingMarkIntoConstraints
+        return decorateAsDefaultView(view)
+    }
+
+    @discardableResult static func decorateAsDefaultView(_ view: UIView) -> UIView {
+        view.backgroundColor = Configuration.Color.Semantic.defaultViewBackground
+        return view
+    }
+
+    static func defaultLabel(text: String) -> UILabel {
+        let label = UILabel()
+        label.text = text
+        return decorateAsDefaultLabel(label)
+    }
+
+    @discardableResult static func decorateAsDefaultLabel(_ label: UILabel) -> UILabel {
+        label.textColor = Configuration.Color.Semantic.defaultForegroundText
+        return label
+    }
+
+}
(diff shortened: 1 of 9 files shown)


AlphaWallet/Extensions/UIButton.swift
```diff
@@ -13,4 +13,17 @@ extension UIButton {
 
         setBackgroundImage(colorImage, for: forState)
     }
+    func setBackgroundColor(lightColor: UIColor, darkColor: UIColor, for state: UIControl.State) {
```
oa-s, line 16: i am not quite familiar with dark/light modes, but why just not to use `Appearences` for colors, it should apply right color automatically

oa-s: Screenshot 2022-06-08 at 16 50 54

AlphaWallet/Configuration.swift
```diff
@@ -0,0 +1,98 @@
...
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.white()!)
+            }
+
+            static let primaryButtonBackgroundLight = R.color.cod()!
+            static let primaryButtonBackgroundDark = R.color.alabaster()!
+            static let primaryButtonHighlightedBackground = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.black()!, darkColor: R.color.concrete()!)
+            }
+            static let primaryButtonBackgroundInactive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.mike()!, darkColor: R.color.mine()!)
```
hboon, line 35: Similar question as Vlad's. Why are these necessary?

AlphaWallet/Configuration.swift
```diff
@@ -0,0 +1,98 @@
...
+    case .unspecified, .light:
+        return lightColor
+    case .dark:
+        return darkColor
+    }
+}
+
+struct Configuration {
+    struct Color {
+        struct Semantic {
```
hboon, line 21: An additional `Semantic` doesn't seem necessary, unless we are planning on more sibling "layers" soon?

AlphaWallet/Configuration.swift
```diff
@@ -0,0 +1,98 @@
...
+            static let secondaryButtonBorderActive = UIColor { trait in
+                return colorFrom(trait: trait, lightColor: R.color.white()!, darkColor: R.color.dusty()!)
+            }
+        }
+    }
+}
+
+class UIKitFactory {
+
+    static func defaultView(autoResizingMarkIntoConstraints: Bool = false) -> UIView {
```
hboon, line 76: How about rename like these:

```
`defaultView(withAutoResizingMarkIntoConstraints:)` -> `createView(withAutoResizingMarkIntoConstraints:)`
`defaultLabel(text: String)` -> `createLabel()`
`decorateAsDefaultView(_:)` -> `configure(view:)`
`decorateAsDefaultLabel(_:)` -> `configure(label:)`
```

And maybe the last 2 should be private?

hboon (commented):
It looks slightly different for me from your screenshots (the "I already have a Wallet" button in particular). Is that intentional?

&nbsp;|&nbsp;
-|-
Screenshot 2022-06-09 at 10 54 44 AM|Screenshot 2022-06-09 at 10 55 50 AM

>Dusty is defined as RGB 153, 153, 153 in the colour guide but the style guide had 151, 151, 151. I followed the colour guide.

I see `153` in the style guide, can you screenshot or point us to where it shows `151`? Just in case there are other inconsistencies.

Screenshot 2022-06-09 at 11 03 56 AM

AlphaWallet/Extensions/UIButton.swift
eviltofu: Can this be applied to a uibutton in the various states?

hboon: @eviltofu In case you wane me to respond, can you elaborate? I don't understand.

@oa-s if you have any idea.

eviltofu: So the reason why I don't use appearances directly is because sometimes there are errors in the colour definition. I've seen Dusty defined in RGB as 153,153,153 and 151,151,151. In order to correct this, I would have to go through all the colours that use Dusty and correct them manually. 

If you drag and drop colours in the xcassets file, the colours set are copies of the RGB values. If I drag and drop Red to my new color, I will get RGB 255, 0, 0. If I change Red to RGB 250, 0, 0 this change is not reflected in my new colour. 

UIButtons have different states (disabled, enabled, highlighted/pressed, etc). Each of these states have a different colour associated with them. The UIView's backgroundColor attribute won't work. There are also no function calls to set the colour of UIButtons in the various states in iOS 13 (there is one in iOS 15). The ones we have are self written by ourselves by creating and setting background images for UIButtons. The current ones only take in one colour and do not reflect dark/light modes.  So I've created a extension function in UIButton to do that.

AlphaWallet/Configuration.swift
eviltofu: The last 2 are not private because I use them to change the configuration of views constructed in other methods; I didn't want to change all the code.

In the future, when a uiview is required, I would recommend that it be vended from UIKitFactory. We can set the font, background color etc all in one spot.

eviltofu:
> It looks slightly different for me from your screenshots (the "I already have a Wallet" button in particular). Is that intentional?
> 
>  	 
> Screenshot 2022-06-09 at 10 54 44 AM	Screenshot 2022-06-09 at 10 55 50 AM
> > Dusty is defined as RGB 153, 153, 153 in the colour guide but the style guide had 151, 151, 151. I followed the colour guide.
> 
> I see `153` in the style guide, can you screenshot or point us to where it shows `151`? Just in case there are other inconsistencies.
> 
> Screenshot 2022-06-09 at 11 03 56 AM

image

image

AlphaWallet/Configuration.swift
eviltofu: I'm thinking of adding styles to these calls. Similar to the ones Apple is using for UILabels.

AlphaWallet/Extensions/UIButton.swift
hboon: > So the reason why I don't use appearances directly is because sometimes there are errors in the colour definition. I've seen Dusty defined in RGB as 153,153,153 and 151,151,151. In order to correct this, I would have to go through all the colours that use Dusty and correct them manually.

Use `153` for all Dusty, until we are told otherwise. Should be an easy change. And no, it will have to be 1 value, not 2

> If you drag and drop colours in the xcassets file, the colours set are copies of the RGB values. If I drag and drop Red to my new color, I will get RGB 255, 0, 0. If I change Red to RGB 250, 0, 0 this change is not reflected in my new colour.

Is this about the question you asked earlier where you want to have references created within the asset catalog pointing to another color in the asset catalog? If so, why does it matter here?

> UIButtons have different states (disabled, enabled, highlighted/pressed, etc). Each of these states have a different colour associated with them. The UIView's backgroundColor attribute won't work. There are also no function calls to set the colour of UIButtons in the various states in iOS 13 (there is one in iOS 15). The ones we have are self written by ourselves by creating and setting background images for UIButtons. The current ones only take in one colour and do not reflect dark/light modes. So I've created a extension function in UIButton to do that.

I understand there are functions that take in colors to create images. So far so good.

> The current ones only take in one colour and do not reflect dark/light modes.

But I don't follow this. It only takes in 1 color, but that color should be the light mode version if it's currently not in dark mode, and the dark mode version when it's in dark mode. It doesn't need to store both light and dark mode versions at runtime. Or is there a case where you think so? Can you write a short snippet and illustrate the problem? As short as possible with as short comments as possible?

AlphaWallet/Configuration.swift
hboon: Not sure what you mean, but let's go with it and see.

hboon:
@eviltofu Use `153` (or change it to `151` if necessary later), but we definitely wouldn't have 151 and 153 at the same time.

AlphaWallet/Extensions/UIButton.swift
eviltofu: So that the OS will automatically switch the colour for us. Instead of writing software hooks to change the colours ourselves.

eviltofu:
> @eviltofu Use `153` (or change it to `151` if necessary later), but we definitely wouldn't have 151 and 153 at the same time.

Yes I am using the one in the Colour Style Guide. So I'm using the colours in this manner so that changing it will be relatively easy.

AlphaWallet/Extensions/UIButton.swift
hboon: Summarizing a chat elsewhere for why it accepts lightColor and darkColor (instead of the current color):

```
func setBackgroundColor(lightColor: UIColor, darkColor: UIColor, for state: UIControl.State) {
```

The gist of it is this function builds a (actually 2) background image and this background image has to switch when iOS tells it to, to match light or dark mode (like how colors would behave)

And this is a special one-off case.

eviltofu:
So I've written code to extract the light/dark mode colours to create the images for the background of the UIButton. Would this be ok now? @hboon @oa-s

eviltofu:
The other question is iOS 15 supports background colours for UIButton via the new Configuration interface. Should this option be added as well?

AlphaWallet/Configuration.swift
eviltofu: I mean in other places, people use their own code to create and set up views. This call just replaces the required properties for those views with minimal disruption to the code.

hboon:
> So I've written code to extract the light/dark mode colours to create the images for the background of the UIButton. Would this be ok now? @hboon @oa-s

Looks OK.

> The other question is iOS 15 supports background colours for UIButton via the new Configuration interface. Should this option be added as well?

Sure if it's simple to do and localized to `setBackgroundColor(_:forState:)`. Otherwise add a TODO in the function.

hboon:
@eviltofu anything else you need from me on this?

eviltofu:
@hboon ok changed `Info.plist > Appearance` to `Light`. Should be ready to merge.

hboon:

hboon (approved):

PR #4848, opened 2022-06-09
Remove Force unwrap when building configuration for transaction #4847

### LINKED ISSUE #4847: Remove Force unwrap when building configuration for transaction

oa-s:
Closes #4847

### DIFF
--- a/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
+++ b/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
@@ -13,8 +13,15 @@ protocol TransactionConfiguratorDelegate: AnyObject {
     func updateNonce(to nonce: Int, in configurator: TransactionConfigurator)
 }
 
+enum TransactionConfiguratorError: Error {
+    case impossibleToBuildConfiguration
+    
+    var localizedDescription: String {
+        return "Impossible To Build Configuration"
+    }
+}
+
 class TransactionConfigurator {
-    private let account: AlphaWallet.Address
 
     private var isGasLimitSpecifiedByTransaction: Bool {
         transaction.gasLimit != nil
@@ -75,11 +82,12 @@ class TransactionConfigurator {
         gasPriceWarning(forConfiguration: currentConfiguration)
     }
 
-    init(session: WalletSession, transaction: UnconfirmedTransaction) {
+    init(session: WalletSession, transaction: UnconfirmedTransaction) throws {
         self.session = session
-        self.account = session.account.address
         self.transaction = transaction
-        self.configurations = .init(standard: TransactionConfigurator.createConfiguration(server: session.server, transaction: transaction, account: account))
+
+        let standardConfiguration = try TransactionConfigurator.createConfiguration(server: session.server, transaction: transaction, account: session.account.address)
+        self.configurations = .init(standard: standardConfiguration)
     }
 
     func updateTransaction(value: BigInt) {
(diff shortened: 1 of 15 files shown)


hboon (changes requested):
This one's not right. When we throw `.failureToBuildConfiguration` below, it's caught in the `do-catch` and lands at https://github.com/AlphaWallet/alpha-wallet-ios/blob/328cd0e3e79e0dcf81606af1eea93665eef19986/AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift#L359:

```
} catch {
    return .init(transaction: transaction, server: server)
}
```

For all the cases that we might throw `.failureToBuildConfiguration`: `.erc20`, `.erc721Token`, `.erc875Token`, `.erc875TokenOrder` and `.erc1155Token`, they all need to build a transaction with data composed from encoding a function call. e.g. for `.erc20`:

```
let function = Function(name: "transfer", parameters: [ABIType.address, ABIType.uint(bits: 256)])
//Note: be careful here with the BigUInt and BigInt, the type needs to be exact
let encoder = ABIEncoder()
try encoder.encode(function: function, arguments: [recipient, BigUInt(transaction.value)])
let gasLimit = transaction.gasLimit ?? maxGasLimit

//…

return createConfiguration(server: server, transaction: transaction, gasLimit: gasLimit, data: encoder.data)
```

But the `catch`'s version would eventually build the data like this:

```
data: transaction.data ?? Data()
```

So the transaction would fail without the user knowing in the UI. (If it's `Data()`, it would become an ETH transfer, which could cause loss of funds, eg. if it's a smart contract function call that is ETH -> some tokens in return, it might end p sending ETH to the smart contract without calling the function).

---

I think those force unwraps should never fail. It's an assumption by the programmer. I guess when it fails it should either crash (maybe not so good), or at least the user shouldn't be allowed to proceed. The simplest way is to rename `.failureToBuildConfiguration` to `.impossibleFailureToBuildConfiguration` to indicate it should never fail. And then in the `catch`, check for it, call `assertImpossibleCodePath()` and then somehow fail? We can't fail gracefully it seems, since most if not all of the force unwraps are towards building the transactions `data`. A better way might be to modify `UnconfirmedTransaction` or replace it before calling `createConfiguration(server:transaction:account:)` so that the optionals are replaced with non-optionals ;)

Or come up with an alternative :)

oa-s:
Yep, i agree not the really good solution. Updates with making configurator failable, with catching errors on its init method.

hboon:
Ok, buzz me if you want me to check again

oa-s:
@hboon pr is ready

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -202,24 +210,38 @@ class TransactionConfigurator {
     }
 
 // swiftlint:disable function_body_length
-    private static func createConfiguration(server: RPCServer, transaction: UnconfirmedTransaction, account: AlphaWallet.Address) -> TransactionConfiguration {
+    private static func createConfiguration(server: RPCServer, transaction: UnconfirmedTransaction, account: AlphaWallet.Address) throws -> TransactionConfiguration {
```
hboon, line 213: So the main changes are in this function and the rest is just propogating the error/throw as well as displaying the error, right?

hboon (approved):

AlphaWallet/Transfer/Controllers/TransactionConfigurator.swift
oa-s: Correct

PR #4851, opened 2022-06-09
Remove passing queue to `GetContractInteractions` #4850

### LINKED ISSUE #4850: Remove passing queues

oa-s:
Closes #4850

### DIFF
--- a/AlphaWallet/Core/NFT/Enjin/Enjin.swift
+++ b/AlphaWallet/Core/NFT/Enjin/Enjin.swift
@@ -17,14 +17,12 @@ typealias EnjinSemiFungiblesToAddress = [AlphaWallet.Address: [GetEnjinTokenQuer
 typealias EnjinSemiFungiblesToTokenId = [String: GetEnjinTokenQuery.Data.EnjinToken]
 
 final class Enjin {
-    private lazy var networkProvider = EnjinNetworkProvider(queue: queue)
-    private let queue: DispatchQueue
-    
+    private lazy var networkProvider = EnjinNetworkProvider()
+
     typealias EnjinBalances = [GetEnjinBalancesQuery.Data.EnjinBalance]
     typealias MappedEnjinBalances = [AlphaWallet.Address: EnjinBalances]
-
-    init(queue: DispatchQueue) {
-        self.queue = queue
+    init() {
+        //no-op
     }
 
     func semiFungible(wallet: Wallet, server: RPCServer) -> Promise<EnjinSemiFungiblesToAddress> {
@@ -48,7 +46,7 @@ final class Enjin {
                     seal.reject(error)
                 }
             }
-        }.then(on: queue, { [networkProvider] balances -> Promise<EnjinSemiFungiblesToAddress> in
+        }.then(on: .none, { [networkProvider] balances -> Promise<EnjinSemiFungiblesToAddress> in
             let ids = (balances[wallet.address] ?? []).compactMap { $0.token?.id }
             return networkProvider.getEnjinTokens(ids: ids, owner: wallet.address)
         })
--- a/AlphaWallet/Core/NFT/Enjin/EnjinNetworkProvider.swift
+++ b/AlphaWallet/Core/NFT/Enjin/EnjinNetworkProvider.swift
@@ -15,7 +15,6 @@ final class EnjinNetworkProvider {
     static let client = URLSessionClient()
     static let cache = InMemoryNormalizedCache()
     static var store = ApolloStore(cache: cache)
-    private let queue: DispatchQueue
 
     private lazy var graphqlClient: ApolloClient = {
         let provider = NetworkInterceptorProvider(store: EnjinNetworkProvider.store, client: EnjinNetworkProvider.client)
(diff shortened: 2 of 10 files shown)


AlphaWallet/Tokens/Logic/GetContractInteractions.swift
```diff
@@ -37,11 +31,11 @@ extension GetContractInteractions {
 extension GetContractInteractions.functional {
 
     //TODO rename this since it might include ERC721 (blockscout and compatible like Polygon's). Or can we make this really fetch ERC20, maybe by filtering the results?
-    static func getErc20Interactions(walletAddress: AlphaWallet.Address, server: RPCServer, startBlock: Int? = nil, queue: DispatchQueue) -> Promise<[TransactionInstance]> {
+    static func getErc20Interactions(walletAddress: AlphaWallet.Address, server: RPCServer, startBlock: Int? = nil) -> Promise<[TransactionInstance]> {
         guard let etherscanURL = server.getEtherscanURLForTokenTransactionHistory(for: walletAddress, startBlock: startBlock) else { return .value([]) }
         return firstly {
-            Alamofire.request(etherscanURL).validate().responseJSON(queue: queue, options: [])
-        }.map(on: queue) { rawJson, _ in
+            Alamofire.request(etherscanURL).validate().responseJSON(queue: .global(), options: [])
```
hboon, line 37: The `queue` arg for `responseJSON(queue:options:)` is just the queue which Alamofire would send the response after processing right? i.e. the actual waiting for the HTTP response, parsing, converting to JSON, is done in its own queue/thread, we don't control that? If we leave it out, it falls back to the main queue and then the `map() {}` in the next line below will just run it on the argument we pass to `map(on:)`.

Then we can leave out the `.global()` completely so it's just `responseJSON(options: [])`.

The `map(on:)` in the next line, we generally can/should remove the queue and just do `map() {}` which defaults to main for PromiseKit. But in this particular case (I remember this can be a slow operation with many transaction records, so it' some of those exceptions), I think we have to pass in `.global()` so it looks like this:

```
Alamofire.request(etherscanURL).validate().responseJSON(options: [])
}.map(on: .global()) { rawJson, _ in
```

Or alternatively, a separate non-main queue instead of `.global()`. Maybe it makes sense to create one for such processing, so it's easier to find, manage and observe too?

Does that sound right?

hboon: Similarly for the other few changes in this commit (I haven't checked the other commit in the PR as of writing this)

AlphaWallet/Core/NFT/Enjin/Enjin.swift
```diff
@@ -48,7 +46,7 @@ final class Enjin {
                     seal.reject(error)
                 }
             }
-        }.then(on: queue, { [networkProvider] balances -> Promise<EnjinSemiFungiblesToAddress> in
+        }.then(on: .none, { [networkProvider] balances -> Promise<EnjinSemiFungiblesToAddress> in
```
hboon, line 49: We can just do `then() {}` and let it default to the fallback queue and pass in `.global()` or a custom queue like in the comment in the other commit if it's slow.

AlphaWallet/Core/NFT/Enjin/EnjinNetworkProvider.swift
```diff
@@ -24,15 +23,11 @@ final class EnjinNetworkProvider {
...
-    init(queue: DispatchQueue) {
-        self.queue = queue
-    }
-
     typealias GetEnjinBalancesResponse = (balances: Enjin.MappedEnjinBalances, owner: AlphaWallet.Address)
 
     func getEnjinBalances(forOwner owner: AlphaWallet.Address, offset: Int, sum: Enjin.MappedEnjinBalances = [:], limit: Int = 50, completion: @escaping (Swift.Result<GetEnjinBalancesResponse, EnjinError>) -> Void) {
 
-        graphqlClient.fetch(query: GetEnjinBalancesQuery(ethAddress: owner.eip55String, page: offset, limit: limit), queue: queue) { response in
+        graphqlClient.fetch(query: GetEnjinBalancesQuery(ethAddress: owner.eip55String, page: offset, limit: limit), queue: .global()) { response in
```
hboon, line 30: I look down the call hierarchy and eventually, the queue is the `callbackQueue` argument, implying it is used only for invoking the `completionHandler`.

```
func send<Operation: GraphQLOperation>(operation: Operation,
                                         cachePolicy: CachePolicy,
                                         contextIdentifier: UUID?,
                                         callbackQueue: DispatchQueue,
                                         completionHandler: @escaping (Result<GraphQLResult<Operation.Data>, Error>) -> Void) -> Cancellable
```

So we should use `.main` and not `.global()`. But since the arg defaults to `.main`, we can leave it out completely to become:

```
graphqlClient.fetch(query: GetEnjinBalancesQuery(ethAddress: owner.eip55String, page: offset, limit: limit)) { response in
```

Unless you think this is intensive (I don't know, you are more familiar with this), then `.global()` or the same custom queue mentioned for the other commit.

AlphaWallet/Core/NFT/Enjin/EnjinNetworkProvider.swift
```diff
@@ -69,7 +64,7 @@ final class EnjinNetworkProvider {
 
         let promises = ids.map { tokenId in
             return Promise<GetEnjinTokenQuery.Data.EnjinToken> { seal in
-                graphqlClient.fetch(query: GetEnjinTokenQuery(id: tokenId), queue: queue) { response in
+                graphqlClient.fetch(query: GetEnjinTokenQuery(id: tokenId), queue: .global()) { response in
```
hboon, line 67: Similar logic like the other comments

hboon (changes requested):
See if the comments make sense? :)

oa-s:
Too many conflicts, easier to rewrite

PR #4852, opened 2022-06-09
Replace passing queue, for nft fetcher with execution in current context #4849

### LINKED ISSUE #4849: Replace passing queue, for nft fetcher with execution in current context

oa-s:
Closes #4849


hboon (changes requested):
Oh, think I commented in #4851

PR #4860, opened 2022-06-13
Make backgroundThreadRealm as lazy #4859

### LINKED ISSUE #4859: Make backgroundThreadRealm as lazy

oa-s:
Closes #4859


hboon:
Can you help add a comment in the code near the property to explain? Like the explanation you added in the issue :)

PR #4874, opened 2022-06-16
[Refactor] Replace promises with publishers for ens resolver #4873

### LINKED ISSUE #4873: [Refactor] Replace promises with publishers for ens resolver

oa-s:
Closes #4873


hboon:
@oa-s should we use `Future`s for 1:1 `Promise` replacement though?

oa-s:
> `Future`s

yes, in implementation generally we use Future, and map it to AnyPublisher

PR #4880, opened 2022-06-18
add PancakeSwap to browser

thandang:
This is to add our dApp to browsers


hboon (approved):

PR #4884, opened 2022-06-21
Syncronize access to `func performSync(_ block: @escaping (Realm) -> Void) {`, to make sure there is no two or more call at same time #4883

### LINKED ISSUE #4883: Fix crash
*** -[AlphaWallet.RunLoopThread performSelector:onThread:withObject:waitUntilDone:modes:]: target thread exited while waiting for the perform

oa-s:
Fixes #4883

### DIFF
--- a/AlphaWallet/Core/Types/RealmStore.swift
+++ b/AlphaWallet/Core/Types/RealmStore.swift
@@ -9,6 +9,7 @@ import Realm
 import RealmSwift
 import Foundation
 
+fileprivate let queueForRealmStore = DispatchQueue(label: "org.alphawallet.swift.realm.store", qos: .background)
 class RealmStore {
     static func threadName(for wallet: Wallet) -> String {
         return "org.alphawallet.swift.realmStore.\(wallet.address).wallet"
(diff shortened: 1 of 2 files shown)


AlphaWallet/Core/Types/RealmStore.swift
```diff
@@ -31,12 +32,18 @@ class RealmStore {
...
+    func performSync(_ block: @escaping (Realm) -> Void) {
         if Thread.isMainThread {
-            callback(mainThreadRealm)
+            block(mainThreadRealm)
         } else {
-            thread.performSync() {
-                callback(self.backgroundThreadRealm)
+            //NOTE: synchronize calls from different threads to avoid
+            //*** -[AlphaWallet.RunLoopThread performSelector:onThread:withObject:waitUntilDone:modes:]: target thread exited while waiting for the perform
+            dispatchPrecondition(condition: .notOnQueue(queueForRealmStore))
```
hboon, line 41: This line crashes if it's not running on `queueForRealmStore`. Where do we enforce that `performSync(:)` (i.e. its caller) always runs on either `.main` or `queueForRealmStore `?

AlphaWallet/Core/Types/RealmStore.swift
```diff
@@ -31,12 +32,18 @@ class RealmStore {
         thread.start()
     }
 
-    func performSync(_ callback: @escaping (Realm) -> Void) {
+    func performSync(_ block: @escaping (Realm) -> Void) {
```
hboon, line 35: I don't understand the original `RealmStore.performSync(_:)` before the PR. Does it work like that:

```
func performSync(block)
    if currentThread == .mainThread
        run(block, .mainThread + mainThreadRealm)
    else
        run(block, .backgroundThread + backgroundThreadRealm)
```

Why is it optionally switching between 2 realm objects based on the caller of `performSync(:)`? Why not just one, say on .main (or background) and then always go to it?

AlphaWallet/Core/Types/RunLoopThread.swift
```diff
@@ -43,14 +43,14 @@ class RunLoopThread: Thread {
         }
     }
 
-    func performSync(_ block: @escaping () -> Swift.Void) {
```
oa-s, line 46: `performSync` - get called for all calls related to realm, when we fetch or write data to realm, all operations call on run loop thread. (maybe function name isn't good)

AlphaWallet/Core/Types/RealmStore.swift
oa-s: > Does it work like that:

yes, pretty like you described

AlphaWallet/Core/Types/RealmStore.swift
oa-s: the `else`  part could be called on any non main thread,  and we use 
```           
dispatchPrecondition(condition: .notOnQueue(queueForRealmStore))
queueForRealmStore.sync {
```
to make sure the no one other is trying to call in same time

hboon (approved):

PR #4896, opened 2022-06-22
Display network icon, etc when confirmation transaction #4499

### LINKED ISSUE #4499: Send ETH transaction's actionsheet shown from dapp browser should show network etc like it does when we send from Wallet tab
Visit https://web3.hboon.repl.co (on testnet! This sends 0.0001 ETH to me).

Screenshot 2022-05-04 at 2 52 56 PM

vs

Screenshot 2022-05-04 at 2 54 20 PM

oa-s:
Closes #4499

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -825,6 +825,15 @@
 		87300E8B2851C9FE009DE4B3 /* AcceptDeepLinkViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87300E8A2851C9FE009DE4B3 /* AcceptDeepLinkViewModel.swift */; };
 		87300E9228573461009DE4B3 /* ConfigureImageStorage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87300E9128573461009DE4B3 /* ConfigureImageStorage.swift */; };
 		872A98662862EC0C00196EA3 /* base64_image_example.txt in Resources */ = {isa = PBXBuildFile; fileRef = 872A98652862EC0C00196EA3 /* base64_image_example.txt */; };
+		872A987128645EF100196EA3 /* SectionProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987028645EF100196EA3 /* SectionProtocol.swift */; };
+		872A987328645F1D00196EA3 /* SendFungiblesTransactionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987228645F1D00196EA3 /* SendFungiblesTransactionViewModel.swift */; };
+		872A987528645F6800196EA3 /* DappOrWalletConnectTransactionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987428645F6800196EA3 /* DappOrWalletConnectTransactionViewModel.swift */; };
+		872A987728645F9A00196EA3 /* TokenScriptTransactionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987628645F9A00196EA3 /* TokenScriptTransactionViewModel.swift */; };
+		872A987928645FC700196EA3 /* SendNftTransactionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987828645FC700196EA3 /* SendNftTransactionViewModel.swift */; };
+		872A987B28645FF100196EA3 /* ClaimPaidErc875MagicLinkViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 872A987A28645FF100196EA3 /* ClaimPaidErc875MagicLinkViewModel.swift */; };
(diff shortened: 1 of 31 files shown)


hboon (changes requested):
Screenshot 2022-06-23 at 11 27 02 AM

This adds Network, still needs to add:

- [ ] Balance
- [ ] Recipient

oa-s:
@hboon it looks like there is missing recipient, 

```
2022-06-23 10:01:26.316131+0300 AlphaWallet[10731:985597] [Browser] dapp action: signTransaction(AlphaWallet.UnconfirmedTransaction(transactionType: AlphaWallet.TransactionType.dapp(TokenObject {
	primaryKey = 0x000000000000000000000000000000000000003c-3;
	chainId = 3;
	contract = 0x000000000000000000000000000000000000003c;
	name = Ropsten;
	symbol = ETH;
	decimals = 18;
	value = 0;
	isDisabled = 0;
	rawType = ether;
	shouldDisplay = 1;
	sortIndex = (null);
	balance = List<TokenBalance> <0x10a71f630> (
	
	);
	_info = TokenInfoObject {
		uid = 0x000000000000000000000000000000000000003c-3;
		coinGeckoId = (null);
		imageUrl = (null);
	};
}, AlphaWallet.DAppRequester(title: Optional("replit"), url: Optional(https://web3.hboon.repl.co/))), value: 10000000000000, recipient: nil, contract: Optional(ethereumAddress: 0xbbce83173d5c1D122AE64856b4Af0D5AE07Fa362), data: Optional(0 bytes), gasLimit: Optional(21000), tokenId: nil, tokenIdsAndValues: nil, gasPrice: nil, nonce: nil, v: nil, r: nil, s: nil, expiry: nil, indices: nil)) from: userContentController(_:didReceive:)
```

is it ok that wallet address located as `contract` field?

hboon:
> it looks like there is missing recipient,

Oh, that's a problem then. I think the last statement of this function in `DappAction` should be modified:

```
private static func makeUnconfirmedTransaction(_ object: [String: DappCommandObjectValue], server: RPCServer, transactionType: TransactionType) -> UnconfirmedTransaction {
    //…
    let data = Data(_hex: object["data"]?.value ?? "0x")
    return UnconfirmedTransaction(
        transactionType: transactionType,
        value: value,
        recipient: nil,
        contract: to,
        data: data,
        gasLimit: gasLimit,
        gasPrice: gasPrice,
        nonce: nonce
    )
```

if `data` is empty or "0x", it should be:

```
    return UnconfirmedTransaction(
        //…
        recipient: to,
        contract: nil,
```

else:

```
    return UnconfirmedTransaction(
        //…
        recipient: nil,
        contract: to,
```

and then change the actionsheet code accordingly. So this case we are targetting should hit the `if`. Can you see if that logic above makes sense?

And so because we aren't building the call-data ourselves it means that if it was say an ERC20 transaction, we will go into the `else` flow (`recipient: nil, contract: to`) and then not be able to display the function call, but we'll leave that for another day.

hboon:
(who knew adding a network icon digs up other things :P)

hboon:
Also the PR title and commit message doesn't capture this:

> Send ETH transaction's actionsheet shown from dapp browser should show network etc like it does when we send from Wallet tab

oa-s:
> Visit https://web3.hboon.repl.co/ (on testnet! This sends 0.0001 ETH to me).

When we decode `value: 10000000000000` it appears as 0 in `BigInt`, so we see it like `amount: 0 eth` in action sheet, is something missing with this number?

hboon:
> > Visit https://web3.hboon.repl.co/ (on testnet! This sends 0.0001 ETH to me).
> 
> When we decode `value: 10000000000000` it appears as 0 in `BigInt`, so we see it like `amount: 0 eth` in action sheet, is something missing with this number?

It should be displayed as "0.00001" instead. It's strange. It appears to be like this:

send amount | shown in actionsheet
-|-
0.00001 | 0.00000 //This one's wrong
0.0001 | 0.001
1 | 1
1.2 | 1.2

See if you can fork this to try https://replit.com/@hboon/web3#script.js

oa-s:
> 

0.00001 | 0.00000 //This one's wrong
-- | --

That is because of our `EtherNumberFormatter` shot number formatter. it drops numbers with less than 4 chars

oa-s:
updating it with `.full` number formatted helps, but it might drake displaying for other

oa-s:
i wonder how it works, when i type, value: as `0.00001`, it works fine and action sheet appears on the screen, but when i type number `1`, it doesn't.

hboon:
> i wonder how it works, when i type, value: as `0.00001`, it works fine and action sheet appears on the screen, but when i type number `1`, it doesn't.

Maybe they go through different formatter objects?

Have to be careful with this PR then. Could you test with the original (i.e. `master`) prior to the PR? If it works the same in terms of amount parsing/formatting, then let's clean up (if necessary) and merge this first and fix the formatting/conversion else where. Better merge sooner than later if there's no impact.

oa-s:
Formatting work as before this pr, there was no changes with it. That is because we crop the number with first 4 digits for formatted value

oa-s:
so there is no changes related to formatting in this PR

hboon:
Ok. Let’s get this merge soonish then and look at the formatting/parsing problem separately.

PR #4924, opened 2022-06-30
Make settings view model subscribable, add reload view model when wallet name or ens resolved, fixes assigning resolved ens name to wrong cell, for settings screen #4903

### LINKED ISSUE #4903: A Temporary Displaying Issue on Settings
Have you noticed a displaying issue loading Settings when you open AlphaWallet? For a moment the page does not displayed SELECT ACTIVE Networks and instead displayed repeatedly CHANGE/ADD WALLET. 

image

oa-s:
Closes #4903


AlphaWallet/Settings/Coordinators/SettingsCoordinator.swift
```diff
@@ -41,7 +41,8 @@ class SettingsCoordinator: Coordinator {
 	var coordinators: [Coordinator] = []
 
 	lazy var rootViewController: SettingsViewController = {
-		let controller = SettingsViewController(config: config, keystore: keystore, account: account, analyticsCoordinator: analyticsCoordinator, domainResolutionService: domainResolutionService)
+        let viewModel = SettingsViewModel(account: account, keystore: keystore, config: config, analyticsCoordinator: analyticsCoordinator, domainResolutionService: domainResolutionService)
+		let controller = SettingsViewController(viewModel: viewModel)
```
hboon, line 45: Indentation seems to be off here.

AlphaWallet/Settings/ViewControllers/SettingsViewController.swift
```diff
@@ -1,7 +1,7 @@
 // Copyright © 2018 Stormbird PTE. LTD.
 
 import UIKit
-import PromiseKit
+import Combine
```
hboon, line 4: There seems to be quite a bit of unrelated changes in the PR. Are all of them needed to fix it?

Can you also rename the PR title and commit? The original issue title doesn't describe well what was the problem.

hboon (changes requested):

AlphaWallet/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -203,3 +375,22 @@ enum SettingsSection {
         }
     }
 }
+
+extension SettingsSection: Hashable {
+    static func == (lhs: SettingsSection, rhs: SettingsSection) -> Bool {
```
hboon, line 380: Think this is not necessary and can be auto-synthesized. (also the `default:` can introduce bugs when we add new cases and the compiler can't remind us to update the `==` function). Probably the other `==` function(s) too?

hboon (approved):
Merge when ready

hboon:
Should also mention fixing the display/refresh bug in commit :)

PR #4930, opened 2022-07-01
Update with full number formatter for transaction confirmation view model, Support negative numbers formatting for `EtherNumberFormatter`

### LINKED ISSUE #4923: Fix parsing (maybe it is just the formatting display?) of send amount from ether.js (not sure about web3.js)
Visit https://web3.hboon.repl.co/ (This sends 0.0001 ETH to me). Don't have to confirm the transaction.

Refer: https://github.com/AlphaWallet/alpha-wallet-ios/pull/4896#issuecomment-1166901989

Just noticed that although it says `0.00000 ETH` (wrong), it says `0.01 USD`. So, maybe the ETH amount is actually correct, just not formatted correctly for display?

Screenshot 2022-06-30 at 2 10 03 PM

oa-s:
Closes #4923

### DIFF
--- a/AlphaWallet/Foundation/EtherNumberFormatter.swift
+++ b/AlphaWallet/Foundation/EtherNumberFormatter.swift
@@ -1,6 +1,6 @@
 // Copyright SIX DAY LLC. All rights reserved.
 import BigInt
-import Foundation
+import Foundation 
 
 extension EtherNumberFormatter {
 
@@ -130,15 +130,13 @@ final class EtherNumberFormatter {
         precondition(minimumFractionDigits >= 0)
         precondition(maximumFractionDigits >= 0)
 
-        let dividend = BigInt(10).power(decimals)
-        let (integerPart, remainder) = number.quotientAndRemainder(dividingBy: dividend)
-        let integerString = self.integerString(from: integerPart)
-        let fractionalString = self.fractionalString(from: BigInt(sign: .plus, magnitude: remainder.magnitude), decimals: decimals)
-        if fractionalString.isEmpty {
-            return integerString
+        let formatted = formatToPrecision(number.magnitude, decimals: decimals)
+        switch number.sign {
+        case .plus:
+            return formatted
+        case .minus:
+            return "-" + formatted
         }
-
-        return "\(integerString)\(decimalSeparator)\(fractionalString)"
     }
 
     private func integerString(from: BigInt) -> String {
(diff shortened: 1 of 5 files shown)


hboon (commented):
Can you describe where supporting negative numbers formatting would make a difference? Can it be seen in a particular screen?

Also I'm still seeing "0.0000 = 0.01 USD" in the actionsheet when hitting https://web3.hboon.repl.co. Is it because the value is too small? If that is so, can we display as many decimals in the `Amount` row as we need to display the value? This is (I think) the only place in the app where we should not drop decimals.

oa-s:
> Can you describe where supporting negative numbers formatting would make a difference? Can it be seen in a particular screen? 
There was some kind of behaviour when wea re trying to format negative number, like: `-0.01234234` and returned result was: `0.001` and not `-0.001`

oa-s:
> Also I'm still seeing "0.0000 = 0.01 USD" in the actionsheet when hitting https://web3.hboon.repl.co/. Is it because the value is too small? If that is so, can we display as many decimals in the Amount row as we need to display the value? This is (I think) the only place in the app where we should not drop decimals.

yes, right, that is because the value is to small. I mentioned in some comment to use full number formatter for this case.

oa-s:
btw, what about displaying `a new balance`? for example when we got `0` balance and trying to send some amount, a new balance going to be with minus sign,  for small amounts it will be like: `-0.0000`

hboon:
> yes, right, that is because the value is to small. I mentioned in some comment to use full number formatter for this case.

OK. Let's do that. Is it possible to drop trailing zeros beyond the decimal spaces we want (currently 4)? I suppose it does that? i.e "0.0000001" but not "0.0000001000"

> btw, what about displaying a new balance? for example when we got 0 balance and trying to send some amount, a new balance going to be with minus sign, for small amounts it will be like: -0.0000

Is this referring to the first row in the screenshot https://github.com/AlphaWallet/alpha-wallet-ios/issues/4923#issue-1289646546. i.e the value "New: 0.3595 ETH"?

In that case, yeah, it should not be negative, the smallest it should go should be `0.0000`.

> There was some kind of behaviour when wea re trying to format negative number, like: -0.01234234 and returned result was: 0.001 and not -0.001

So this PR doesn't close #4923, right? :) If so, want to update the first comment?

oa-s:
> Is this referring to the first row in the screenshot https://github.com/AlphaWallet/alpha-wallet-ios/issues/4923#issue-1289646546. i.e the value "New: 0.3595 ETH"?

Yes

hboon:
Remember:

> So this PR doesn't close https://github.com/AlphaWallet/alpha-wallet-ios/issues/4923, right? :) If so, want to update the first comment?

Ok. Buzz me if you want me to test or anything :)

oa-s:
> So this PR doesn't close https://github.com/AlphaWallet/alpha-wallet-ios/issues/4923, right? :) If so, want to update the first comment?

it does, here what we got after pr:
IMG_0011

hboon (approved):
Great, thanks. 

> 0,00001  BNB

But I thought we changed it quite a while ago so that the decimal separator is always "." and thousands separator is always "," no matter which locale is used?

If reverting back to locale-specific separator (so "," for you) is because of this PR, can you fix it and then merge? Otherwise let's merge and fix it to always use "." for decimal and "," for thousands separately.

oa-s:
> But I thought we changed it quite a while ago so that the decimal separator is always "." and thousands separator is always "," no matter which locale is used?

in this pr it doesn't change seprator logic, it shouldn't , i will check

Issue #4931, opened 2022-07-02
Do not add 1gwei to `2000wei` gas price estimate for PHI network


Phinetwork:
Gas fees are much too high for phi network. transactions can process fast with 2000 wei . 1 gwei is very high for chain and makes it very expensive for users. can we please adjust gas to 2000 wei and not 1 gwei thank you

hboon:
@Phinetwork in case, shouldn't eth_gasPrice return `1` instead of `2000` (`0x7d0`)?

```
curl -H 'accept: application/json' -H 'content-type: application/json' -H 'user-agent: AlphaWallet/395 CFNetwork/1237 Darwin/20.5.0' -H 'accept-language: en-us' --data-binary ' {"method":"eth_gasPrice","params":[],"id":1,"jsonrpc":"2.0"}' --compressed "https://rpc1.phi.network" -s

{"jsonrpc":"2.0","id":1,"result":"0x7d0"}
```

Phinetwork:
not sure what you mean. I Just know that chain will process fast with transaction cost of 2000 wei 1gwei is very expensive for the tokenomics. but 2000 wei works to process fast and is appropriate for tokenomics.

hboon:
I'm sorry I mixed up the 2 gas prices (`2000` and `1`) when I was writing this comment: https://github.com/AlphaWallet/alpha-wallet-ios/issues/4931#issuecomment-1173310044.

Will fix this. There will be a limitation in the UI though, the gas price used will show up as "~0 PHI". Our gas UI is tailored for `gwei` :)

Phinetwork:
thats ok if gas price shows up at 0 gwei would perfer that they have user pay so high gas. will make it so users can practically use with wei so its ok we appreciate the update

PR #4951, opened 2022-07-08
Fix: do not add 1gwei to PHI's gas price estimate (because it is already very low: 2000wei as of writing) #4931

hboon:
Fixes: #4931

### DIFF
--- a/AlphaWallet/Settings/Types/RPCServers.swift
+++ b/AlphaWallet/Settings/Types/RPCServers.swift
@@ -349,6 +349,16 @@ enum RPCServer: Hashable, CaseIterable {
         }
     }
 
+    var shouldAddBufferWhenEstimatingGasPrice: Bool {
+        switch self {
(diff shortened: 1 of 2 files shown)


Phinetwork:
> Fixes: #4931

Are we able to merge this PR?

oa-s (approved):

Phinetwork:
Can we merge this PR?

Phinetwork:
Is this merge in the newest update because when go to send still defaults at 1 gwei. Just curious if this merge is already live or will be live in next update

hboon:
@Phinetwork hi! it's been merged into master, but it's not live yet. The release from last week was a release to try something out so it has minimal changes. The next one would include this.

Phinetwork:
> @Phinetwork hi! it's been merged into master, but it's not live yet. The release from last week was a release to try something out so it has minimal changes. The next one would include this.

Awesome thank you for the update appreciate it and your help. 🙏 are we also in the process of getting phi price api added.

PR #4958, opened 2022-07-10
Add “mail³” to the dapps list

qiudeqiu:
If bug:
Fixes #

If not a bug:
Closes #

1-liner summary:

Anything particular to test:


hboon:
@qiudeqiu thanks. Would you help squash them into a single commit and so we can merge?

qiudeqiu:
@hboon Thanks! I have squashed them into a single commit.

hboon (approved):
Thanks for contributing and helping to squash. We are merging this.

Unfortunately the app store forbade us from listing dapps, so this list is not currently used. It is still maintained in this repo, but in the meantime this will not be visible in the app at the moment.

PR #4967, opened 2022-07-12
Re-architecting to have a "data pipeline" to support upcoming Wallet API #4918

### LINKED ISSUE #4918: Re-architecting to have a "data pipeline" to support upcoming Wallet API
Basic idea of pipeline:

```
┌────────────────┐    ┌────────────────┐    ┌────────────────┐              ┌────────────────┐
│                │    │ Call Infura +  │    │                │              │                │
│                │    │  Etherscan +   │    │                │              │                │
│                │    │  OpenSea and   │    │                │              │   Wallet and   │
│Call Wallet API │───▶│     other      │───▶│ Call CoinGecko │──Display────▶│ Activity tabs  │
│                │    │  RPC/explorer  │    │                │              │                │
│                │    │      APIs      │    │                │              │                │
│                │    │                │    │                │              │                │
└────────────────┘    └────────────────┘    └────────────────┘              └────────────────┘
```

Especially when the Wallet API isn't ready yet, we'll still be pulling the same data from the same data sources (Infura, Etherscan, OpenSea, CoinGecko, etc), but we'll structure it so that the flow is more uni-directional with the output of the pipeline being the data that is ready to be displayed directly in the Wallet and Activity tabs.

And when the Wallet API is ready, it'll still be the same flow, except we'll only use Infura/Etherscan for private chains (for forks) or for chains that the Wallet API does not support. So the 2nd box above is just backfilling the Wallet API's data. [^1]

Then following that, we do the usual calls for CoinGecko (again if the Wallet API includes those, we can drop it) to supplement with ticker data, etc.

One way to look at it is at the end, we have 2 big JSON [^2], both arrays — 1 for the Wallet tab and another for the Activity tab. They just need to be mapped to model (model-view) objects and displayed on screen. So most if not all of the data processing logic will be moved out from the current coordinators/controllers/models.

And when there are changes to the data (for example we refresh CoinGecko data), we run through the pipeline again, using cached data if necessary and then end up with the 2 big JSON and refresh the entire Wallet tab and Activity tabs.

[^1] The step to call Infura/Etherscan still needs to be around for private chains (forks) or chains that the Wallet API doesn't support.
[^2] It might not be JSON, but it's the simplest way to look at it and possibly the easiest way to manage it…

oa-s:
Closes #4918


hboon:
@oa-s do you want me to look at anything in particular?

oa-s:
Yes please, could you check if this imp, of arch, fits our plan. This is still wil, might be missed parts.

hboon:
Where are good starting points to look at? :)

oa-s:
The top line is `TokensProcessingPipeline`

oa-s:
i will split this pr on smallest parts

hboon:
> TokensProcessingPipeline

Oh shucks. The first keyword I searched for was "pipeline" and I didn't see it.

> i will split this pr on smallest parts

Thanks! Buzz me again for this after we go through the smaller refactorings then.

oa-s:
@hboon pr is ready for review

AlphaWallet/Core/Types/Tokens/TokensProcessingPipeline.swift
```diff
@@ -0,0 +1,234 @@
...
+            .store(in: &cancelable)
+    }
+
+    private func startTickersHandling() {
+        //NOTE: To don't block start method, and apply delay to fetch tickers, inital only
+        Publishers.Merge(Just(tokensService.tokens).delay(for: .seconds(2), scheduler: RunLoop.main), tokensService.newTokens)
+            .removeDuplicates()
+            .eraseToAnyPublisher()
+            .sink { [coinTickersFetcher, tokensService] tokens in
+                let nativeCryptoForAllChains = RPCServer.allCases.map { MultipleChainsTokensDataStore.functional.etherToken(forServer: $0) }
```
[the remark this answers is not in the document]
oa-s: nope, that came from some of pr before, there was one case (not remember actually where) there was an issue with loading tickers for native crypto.

oa-s:
> For diffable source, can we make it so that walletSummary, blockieImage, etc all becomes flattened in the future? So that this line (and similar lines):

i guess yes, that could be possible

AlphaWallet/Core/Types/Tokens/NativeTokenSourceAdapter.swift
```diff
@@ -0,0 +1,138 @@
...
+//  NativeTokenSourceAdapter.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 08.07.2022.
+//
+
+import Foundation
+import Combine
+
+class NativeTokenSourceAdapter: TokenSourceAdapter {
```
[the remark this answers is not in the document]
oa-s: yep right, maybe `ClientSideTokenSourceProvider` will be better

AlphaWallet/Core/Types/Tokens/TokensProcessingPipeline.swift
```diff
@@ -0,0 +1,234 @@
...
+    func mark(token: TokenIdentifiable, isHidden: Bool) {
+        tokensService.mark(token: token, isHidden: isHidden)
+    }
+
+    func refresh() {
+        tokensService.refresh()
+    }
+    
+    func start() {
+        cancelable.cancellAll()
```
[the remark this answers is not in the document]
hboon: should -> typo. I meant to say "shouldn't"

hboon:
> > For diffable source, can we make it so that walletSummary, blockieImage, etc all becomes flattened in the future? So that this line (and similar lines):
> 
> i guess yes, that could be possible

Yes, I would love to see how fast the app launches with cached data and how fast it refreshes when it fetches data, in the future :)

AlphaWallet/Core/Types/Tokens/TokensProcessingPipeline.swift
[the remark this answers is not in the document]
oa-s: it came, from prev impl, when start could be called 2 times

AlphaWallet/Core/Types/Tokens/AlphaWalletTokensService.swift
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    func mark(token: TokenIdentifiable, isHidden: Bool) {
+        let primaryKey = TokenObject.generatePrimaryKey(fromContract: token.contractAddress, server: token.server)
+        tokensDataStore.updateToken(primaryKey: primaryKey, action: .isHidden(isHidden))
+    }
+
+    func token(for contract: AlphaWallet.Address) -> Token? {
+        //NOTE: same as below
+        return tokensDataStore.token(forContract: contract)
```
oa-s, line 70: @hboon what do you think about these two methods, do we need to get token directly from `tokensDataStore ` to route a call to right chain related provider?

hboon: I'm not sure, what's the alternative? But either way, `tokens(for:)` should use the same mechanism, right?

```
func tokens(for servers: [RPCServer]) -> [Token] {
        return tokensDataStore.enabledTokens(for: servers)
    }
```

oa-s: yes

hboon: Ok then. Do what looks right to you. We can tweak it later if necessary

oa-s:
@hboon can you please check commit 53b021f39428d8a2ef24090e4524552fda5e5858, added removing duplicates to reduce updates

AlphaWallet/Core/Types/Tokens/TokensProcessingPipeline.swift
```diff
@@ -0,0 +1,234 @@
...
+                }
+            }).map { [tokensService] _ in tokensService.token(for: contract, server: server) }
+            .receive(on: RunLoop.main)
+            .eraseToAnyPublisher()
+
+        return Publishers.Merge4(tokensService.tokenPublisher(for: contract, server: server), whenTickersHasChanged, whenSignatureOrBodyChanged, whenEventHasChanged)
+            .map { $0.flatMap { TokenViewModel(token: $0) } }
+            .map { [weak self] in self?.applyTicker(token: $0) }
+            .map { [weak self] in self?.applyTokenScriptOverrides(token: $0) }
+            .removeDuplicates(by: { return $0?.hashValue == $1?.hashValue })
```
hboon, line 124: Why do we need to call and compare `.hasValue` explicitly? Is it because we override `==` to compare contract addresses only, but we want to check the balance etc here?

oa-s: this is needed especially when we resolve coin tickers, result tokenViewModelPublisher might be triggered even if token related ticker id hasn't changed, so wat to avoid such events, the same for tokens collection events there might be too many events, even if we apply diffable data source, sorting, filtering, is going to be applied to snapshot even collection elements haven't changed.

oa-s: yes, as we override == operator, we might need make full comparison, for all fields

AlphaWallet/Core/CoinTicker/CoinGecko/CoinGeckoTickersFetcher.swift
```diff
@@ -44,6 +44,15 @@ final class CoinGeckoTickersFetcher: CoinTickersFetcher {
         promises.removeAll()
     }
 
+    func addOrUpdateTestsOnly(ticker: CoinTicker?, for token: TokenMappedToTicker) {
```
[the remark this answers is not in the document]
oa-s: optional to remove value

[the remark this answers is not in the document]
hboon: But it defaults to an empty dictionary so it never overrides key=`token` with nil, right?

PR #5029, opened 2022-07-20
[Refactor] Replace mockable FakeKeystore with real FakeEtherKeystore #5028

### LINKED ISSUE #5028: [Refactor] Replace mockable FakeKeystore with real FakeEtherKeystore

oa-s:
Closes #5028


hboon:
To review? Remember the `wip` commit.

oa-s:
> BTW, do you think we should keep the codeowner configuration which sets a default reviewer? Or does having it make it worse?

i don't know, up to you :)

hboon:
> > BTW, do you think we should keep the codeowner configuration which sets a default reviewer? Or does having it make it worse?
> 
> i don't know, up to you :)

Ok, Created #5033. Let's see if the PR is done correctly and if it helps :)

hboon (approved):

PR #5047, opened 2022-07-25
Delete TokenBalance from Realm when remove from list, fix test quite #5046

### LINKED ISSUE #5040: Check test suite
Tests seems to be broken. But I'm not sure why. Can you help to check by running locally @oa-s ?

This is for one the PRs:

https://app.travis-ci.com/github/AlphaWallet/alpha-wallet-ios/builds/253463201

But seems to happen on `master` (d6448bd84b0cf10a7f389111a98fd15418b8af37) too. The errors seems to be different for iOS 15.x and 14.x. 15.x seems to work better.

### LINKED ISSUE #5046: Delete TokenBalance from Realm when remove from list

### LINKED ISSUE #5048: Crash when running entire test suite. 11 Pro iOS 15.4 simulator Raw
Seems to be crashing randomly in Combine code.

https://gist.github.com/hboon/b5f761c17c091b8f0a77d9b77e72bdc1

oa-s:
Closes #5046, Closes #5048, Closes #5040

### DIFF
--- a/AlphaWallet/Tokens/Types/TokensDataStore.swift
+++ b/AlphaWallet/Tokens/Types/TokensDataStore.swift
@@ -26,6 +26,7 @@ protocol TokensDataStore: NSObjectProtocol {
     func token(forContract contract: AlphaWallet.Address, server: RPCServer) -> Token?
     func add(hiddenContracts: [AddressAndRPCServer])
     func deleteTestsOnly(tokens: [Token])
+    func tokenBalancesTestsOnly() -> [TokenBalanceValue]
     func add(tokenUpdates updates: [TokenUpdate])
     @discardableResult func addCustom(tokens: [ERCToken], shouldUpdateBalance: Bool) -> [Token]
     @discardableResult func updateToken(primaryKey: String, action: TokenUpdateAction) -> Bool?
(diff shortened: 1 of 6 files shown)


AlphaWallet/Tokens/Types/TokensDataStore.swift
```diff
@@ -577,21 +586,18 @@ class MultipleChainsTokensDataStore: NSObject, TokensDataStore {
...
         //NOTE: remove old balances if something has changed
-        let balancesToDelete = token.balance
-            .filter { !balance.rawValue.contains($0.balance) }
-            .compactMap { token.balance.index(of: $0) }
+        let balancesToDelete = Array(token.balance
+            .filter { !balance.rawValue.contains($0.balance) })
 
         if !balancesToAdd.isEmpty || !balancesToDelete.isEmpty {
-            for index in balancesToDelete {
-                token.balance.remove(at: index)
```
hboon, line 593: The old version doesn't seem to work correctly because when it deletes the size (and hence the indices) are modified.

But I don't understand how both the old and new version of deletion works though. Why do we delete the `TokenBalance` instance from `token.balance` array if it matches?

hboon (changes requested):
Test `testTokenBalanceDeletion()` fails on Travis https://app.travis-ci.com/github/AlphaWallet/alpha-wallet-ios/jobs/577625115 but seems to work for me locally when only running `TokenObjectTest`.

AlphaWallet/Tokens/Types/TokensDataStore.swift
oa-s: we can suppose that function `updateNonFungibleBalance` calls with valid balances values, so to avoid replacing values that not changed
we look for balances that we are going to delete, (that don't contain in `balance:` parameter), that might be old balances. To remove values we loopp through list, to resolve object index in list, as `List<>` has method `remove(at:` that accepts objects index. Then just map throught indecies to delete object. Out of this unchanged balances stays same, new balances will be added to list, old balances removed from list. There only one issue, that `remove(at:` method doesn't remove object from Realm, so it stays in realm.

hboon: Oh I see!

hboon (approved):

PR #5094, opened 2022-08-02
Fix: Klaytn NFT token balance wrongly displayed as 0 #5093

hboon:
Fixes: #5093


oa-s (approved):

hboon:
@oa-s do you know why we seemed to be able to detect the Klaytn balance earlier? Is this the only way it was picked up?

oa-s:
i guess they changed response field, because previously it was as `value` field.

PR #5109, opened 2022-08-03
Make swap tokens screen being dismissed by swipe #5108

### LINKED ISSUE #5108: Make swap tokens screen being dismissed by swipe

oa-s:
Closes #5108


hboon (approved):
Why did it not work previously?

oa-s:
> Why did it not work previously?

when back button has overridden, swipe doesn't work

PR #5118, opened 2022-08-05
Read API keys from environment variables if available

hboon:
For Travis since we'll only set these environment variables there.


hboon:
I have no idea why it fails on Travis. Going to try to merge and see what happens. Or do you see why it might fail @oa-s ?

oa-s (approved):

PR #5138, opened 2022-08-11
Fix ENS resolution slow or broken in Send fungible screen in certain cases

### LINKED ISSUE #5137: Activity indicator under address textbox sometimes spins forever
A few cases:

* Tap an EIP681 link like ethereum:0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359?value=2.014e18
* or paste "hweeboon" and type "." then "e" then "t" (so it becomes "hweeboon.et")

Screenshot 2022-08-11 at 12 28 19 PM

hboon:
Fixes #5137

### DIFF
--- a/AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
+++ b/AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
@@ -37,14 +37,14 @@ extension DomainResolutionService: DomainResolutionServiceType {
         }
 
         return Just(value)
-            .receive(on: Config.backgroundQueue)
+            //We want to be sure it's on main
+            .receive(on: RunLoop.main)
             .setFailureType(to: SmartContractError.self)
             .flatMap { [getEnsAddressResolver] value in
                 getEnsAddressResolver.getENSAddressFromResolver(for: value)
             }.catch { [unstoppableDomainsV2Resolver] _ -> AnyPublisher<AlphaWallet.Address, PromiseError> in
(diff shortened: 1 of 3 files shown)


AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
```diff
@@ -37,7 +37,7 @@ extension DomainResolutionService: DomainResolutionServiceType {
         }
 
         return Just(value)
-            .receive(on: Config.backgroundQueue)
```
oa-s, line 40: Why do we need to move to main queue here?

oa-s (approved):

AlphaWallet/Core/UnstoppableDomains/DomainResolutionService.swift
hboon: Running it on background and then delivering it in the main thread introduces a predicable ~8 second delay before the background work starts. I don't know why. But seems a good idea to move it completely to the main thread. Most of the time is just spent waiting for the network. I *think* the way to go is just always do things on the main thread, very few select things need to be on background threads. It just complicates things.

I'm going to modify the code a little before force-pushing though. In the PR, it now calls `.receive(on: Runloop.main)` at the start and end of the chain, it should just need 1 call.

oa-s: ok, got it.  There is a call in some of inner functions `address.nameHash` and its could be good to avoid make it on main thread, might block ui

PR #5156, opened 2022-08-12
Fixes updating ui in background #5155

### LINKED ISSUE #5155: Updating ui in background

oa-s:
Fixes #5155


oa-s:
Screenshot 2022-08-12 at 09 02 57

hboon:
> Screenshot 2022-08-12 at 09 02 57

@oa-s why do you think this happens? Prior to #5156, the `receive(to:)` was caller with `.main` earlier on the chain, doesn't that move everything after that to be on `.main`? The code that called `sink{}` in the screenshot above is already on `main` right? What causes the `sink`'s closure arg to be running in the background?

oa-s:
its because of order of `receive(to:)` we receive only ` Just(address)` on main thread, the rest of operators might return data on non main queue.

PR #5158, opened 2022-08-14
Add ApeNow to the iOS dapps list

apenowdev:
Looking to add ApeNow to the dapps list!


hboon (approved):

hboon:
@apenowdev Thanks for contributing. We are merging this.

Unfortunately the app store forbade us from listing dapps, so this list is not currently used. It is still maintained in this repo, but in the meantime this will not be visible in the app at the moment.

PR #5171, opened 2022-08-16
[Refactor] Edit transaction configuration view #5170

### LINKED ISSUE #5170: [Refactor] Edit transaction configuration view
[Refactoring as part of https://github.com/AlphaWallet/alpha-wallet-ios/issues/4138

oa-s:
Closes #5170


hboon (commented):
Does this PR force the warning to not appear if it's the default configuration (i.e. calculated with `eth_gasPrice`? If so, we shouldn't hide it because the network itself might be congested.

oa-s:
oh, i see, will update prs title

PR #5172, opened 2022-08-16
EIP1559 implementation #4138

### LINKED ISSUE #4138: Support EIP1559
1. I used the Estimate fee calculation from this code: https://github.com/zsfelfoldi/feehistory, also ported to Kotlin here: https://github.com/komputing/KEthereum/blob/master/eip1559_feeOracle/src/main/kotlin/org/kethereum/eip1559_fee_oracle/EIP1559FeeOracle.kt
#### Note that we may want to override this calculation with an oracle later, such as the BlockNative Gas Oracle where available. I have compared the results of the calculation with the values appearing in Etherscan of the newly written blocks - it matches quite well. ####

2. This calculation uses the ```eth_feeHistory``` node call results and returns an array of 16 values from the [0]: Rapid to [15]: Slow
```
eth_feeHistory (eg:{"jsonrpc":"2.0","method":"eth_feeHistory","params":["0x64", "latest",[25, 75]],"id":1})  returns:
public class FeeHistory
{
    public String[] baseFeePerGas;
    public double[] gasUsedRatio;
    public String oldestBlock;
    public String[][] reward;
}
```
The elements of the array returned from the calculation looks like this:
```
public class EIP1559FeeOracleResult
{
    public final BigInteger maxFeePerGas;
    public final BigInteger maxPriorityFeePerGas;
}
```
(On the Android implementation I also recorded the baseFee in case this is needed for the UI).

3. Not all chains support EIP1559! You will need to distinguish; this is as simple as the node not recognising the opcode ```eth_feeHistory``` and returning an error. If this happens (eg Artis chain) you will need to fall back to Legacy style. On Android, the compromise I used was not to allow changing the gas speed for those chains. This will be expanded back to allow selection as this means you can't cancel or speedup transactions on those chains.
4. If you have implemented a wallet sweep action; this is only possible using Legacy transaction style since the exact amount of gas used is not known with 1559 until the transaction is written.

The design choices I used for Android:
- Maintain the legacy system for chains that don't support EIP1559 (including Artis, Optimism).
- Select the [0] element for Rapid, [15] for Slow, and calculate the intermediate two points for Standard and Fast.
- Use the same UI as we currently use to display Rapid -> Slow selection, but add 'Max Fee' and 'Priority Fee', similar to this: https://www.blocknative.com/gas-estimator
- Convert Gas Price slider to Max Fee slider.
- Add additional slider for Priority Fee. Minimum is 0, max is 4. (TODO: use eth_maxPriorityFeePerGas instead of 4, check this is what this call is for).
- Sometimes when testing lesser used chains like Gnosis chain the eth_feeHistory call returns a 'garbage' result where max gas would be returned at 0x7, ie 0.00000007 Gwei, which is clearly not correct. To deal with this, use a minimum of 1 Gwei Max Fee and minimum 1 Gwei Priority Fee.
- Tested on Mainnet and Testnet chains results seem good. Interestingly Polygon doesn't have much variance between rapid and slow. This corresponds with the 1559 fees used on the chain by written transactions.

As you might imagine, this corresponds to a lot of additional infrastructure. Hopefully knowing all the information before starting it makes your implementation a little smoother.

oa-s:
Closes #4138

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/Eip1559FeeOracle.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/Eip1559FeeOracle.swift
@@ -1,0 +1,41 @@
+//
+//  Eip1559FeeOracle.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 16.08.2022.
+//
+
+import Foundation
+import BigInt
+import Combine
+
+class Eip1559FeeOracle {
+    private let suggestEip1559: SuggestEip1559
+    private let blockchainProvider: BlockchainProvider
+
+    init(blockchainProvider: BlockchainProvider) {
+        self.blockchainProvider = blockchainProvider
+        self.suggestEip1559 = SuggestEip1559(blockchainProvider: blockchainProvider)
+    }
+
+    func eip1559FeeEstimates(blockCount: Int = 100, block: BlockParameter = .latest, rewardPercentile: [Int] = []) async throws -> Eip1559FeeEstimates {
+        do {
+            let feeHistory = try await blockchainProvider.feeHistory(blockCount: blockCount, block: block, rewardPercentile: rewardPercentile).first
+            let priorityFee = try await suggestEip1559.suggestPriorityFee(feeHistory: feeHistory)
+            let elems = try await suggestEip1559.calculateResult(priorityFee: priorityFee, feeHistory: feeHistory)
+
+            var estimates: [GasSpeed: Eip1559FeeOracleResult] = [:]
+
+            let third = elems.count / 3
+
+            estimates[.rapid] = elems[0]
+            estimates[.fast] = elems[third]
+            estimates[.standard] = elems[third * 2]
+            estimates[.slow] = elems[elems.count - 1]
+
+            return Eip1559FeeEstimates(estimates: estimates)
+        } catch {
+            throw SessionTaskError(error: error)
+        }
+    }
+}
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
@@ -1,0 +1,199 @@
+//
+//  SuggestEip1559.swift
+//  AlphaWallet
+//  https://github.com/komputing/KEthereum/blob/master/eip1559_feeOracle/src/main/kotlin/org/kethereum/eip1559_fee_oracle/EIP1559FeeOracle.kt
+//  Created by Vladyslav Shepitko on 16.08.2022.
+//
+
+import Foundation
+import BigInt
+import AlphaWalletLogger
+
+actor SuggestEip1559 {
+    /// priority fee offered when there are no recent transactions
+    private let fallbackPriorityFee: Decimal = 2
+
+    /// effective reward value to be selected from each individual block
+    private let rewardPercentile = [10]
+
+    /// suggested priority fee to be selected from sorted individual block reward percentiles
+    private let rewardBlockPercentile = 40
+
+    /// highest timeFactor index in the returned list of suggestion
+    private let maxTimeFactor: Int = 15
+
+    /// sampled percentile range of exponentially weighted baseFee history
+    private let sampleMinPercentile: Double = 10
+
+    private let sampleMaxPercentile: Double = 30
+
+    /// extra priority fee offered in case of expected baseFee rise
+    private let extraPriorityFeeRatio = Decimal(0.25)
+
+    let blockchainProvider: BlockchainProvider
+
+    init(blockchainProvider: BlockchainProvider) {
+        self.blockchainProvider = blockchainProvider
+    }
+
+    func suggestPriorityFee(feeHistory: FeeHistory) async throws -> Decimal {
+        try await suggestPriorityFee(firstBlock: feeHistory.oldestBlock, gasUsedRatio: feeHistory.gasUsedRatio)
+    }
+
+    private func suggestPriorityFee(firstBlock: Int, gasUsedRatio: [Double]) async throws -> Decimal {
+        var ptr = gasUsedRatio.count - 1
+        var needBlocks = 5
+        var rewards: [Double] = []
+        while needBlocks > 0 && ptr >= 0 {
+            let blockCount = maxBlockCount(gasUsedRatio: gasUsedRatio, ptr, needBlocks)
+            if blockCount > 0 {
+                let lastBlock = firstBlock + Int(ptr)
+
+                let feeHistory = try await blockchainProvider.feeHistory(
+                    blockCount: blockCount,
+                    block: .blockNumber(value: lastBlock),
+                    rewardPercentile: rewardPercentile).first
+
+                let rewardSize = feeHistory.reward.count
+                guard !feeHistory.reward.isEmpty else { throw EIP1559Error.rewardsNotFound }
+                for reward in feeHistory.reward {
+                    let value = reward[0]
+                    rewards += [value]
+                }
+                if rewardSize < blockCount {
+                    break
+                }
+                needBlocks -= blockCount
+            }
+            ptr -= blockCount + 1
+        }
+
+        if rewards.isEmpty {
+            return fallbackPriorityFee
+        }
+        rewards.sort()
+
+        let index = Int(floor(Double((rewards.count - 1) * rewardBlockPercentile) / 100.0))
+
+        return Decimal(rewards[index])
+    }
+
+    enum EIP1559Error: Error {
+        case baseFeeIsEmpty
+        case firstBlockMissing
+        case failureToBuildTimedEstimate
+        case indexedRewardNotFound
+        case rewardsNotFound
+        case indexedBaseFeeNotFound
+        case minBaseFeeNotFound
+    }
+
+    func calculateResult(priorityFee: Decimal, feeHistory: FeeHistory) throws -> [Int: Eip1559FeeOracleResult] {
+        var baseFee = feeHistory.baseFeePerGas
+
+        guard !baseFee.isEmpty else { throw EIP1559Error.baseFeeIsEmpty }
+
+        let lastBaseFee = baseFee[baseFee.count - 1] * 1.125
+        baseFee[baseFee.count - 1] = lastBaseFee
+        for i in feeHistory.gasUsedRatio.indices.reversed() where feeHistory.gasUsedRatio[i] > 0.9 {
+            baseFee[i] = baseFee[i + 1]
+        }
+
+        let order: [Int] = Array(0 ... feeHistory.gasUsedRatio.count).indices.sorted {
+            guard let first = baseFee[safe: $0] else { return false }
+            guard let second = baseFee[safe: $1] else { return true }
+            return first < second
+        }
+
+        var maxBaseFee: Decimal = .zero
+        var result: [Int: Eip1559FeeOracleResult] = [:]
+
+        for timeFactor in (0...maxTimeFactor).reversed() {
+            var bf = try predictMinBaseFee(baseFee: baseFee, order: order, timeFactor: timeFactor)
+            var t = priorityFee
+            bf = max(bf, maxBaseFee)
+
+            if bf > maxBaseFee {
+                maxBaseFee = bf
+            } else {
+                // If a narrower time window yields a lower base fee suggestion than a wider window then we are probably in a price dip.
+                // In this case getting included with a low priority fee is not guaranteed; instead we use the higher base fee suggestion
+                // and also offer extra priority fee to increase the chance of getting included in the base fee dip.
+                t += (maxBaseFee - bf) * extraPriorityFeeRatio
+            }
+
+            infoLog("[Eip1559] estimate: \(timeFactor) maxFeePerGas: \(bf + t), maxPriorityFeePerGas: \(t)")
+
+            result[timeFactor] = Eip1559FeeOracleResult(
+                maxFeePerGas: (bf + t).toBigUInt(units: .gwei)!,
+                maxPriorityFeePerGas: t.toBigUInt(units: .gwei)!)
+        }
+
+        return result
+    }
+
+    private func predictMinBaseFee(baseFee: [Double], order: [Int], timeFactor: Int) throws -> Decimal {
+        assert(!baseFee.isEmpty)
+
+        let timeFactor = Double(timeFactor)
+        if timeFactor < 1e-6 {
+            guard let value = baseFee.last else { throw EIP1559Error.minBaseFeeNotFound }
+            return Decimal(value)
+        }
+
+        let pendingWeight = (1 - exp(-1 / timeFactor)) / (1 - exp(-Double(baseFee.count) / timeFactor))
+        var sumWeight: Double = 0.0
+        var result: Double = .zero
+        var samplingCurveLast: Double = 0.0
+
+        for each in order {
+            sumWeight += pendingWeight * exp(Double(each - baseFee.count + 1) / timeFactor)
+            let samplingCurveValue = samplingCurve(percentile: sumWeight * 100.0)
+
+            result += (samplingCurveValue - samplingCurveLast) * baseFee[each]
+
+            if samplingCurveValue >= 1 {
+                return Decimal(result)
+            }
+            samplingCurveLast = samplingCurveValue
+        }
+
+        return Decimal(result)
+    }
+
+    private func samplingCurve(percentile: Double) -> Double {
+        if percentile <= sampleMinPercentile {
+            return 0.0
+        } else if percentile >= sampleMaxPercentile {
+            return 1.0
+        } else {
+            return Double((1 - cos((percentile - sampleMinPercentile) * 2 * Double.pi / (sampleMaxPercentile - sampleMinPercentile))) / 2)
+        }
+    }
+
+    private func maxBlockCount(gasUsedRatio: [Double], _ _ptr: Int, _ _needBlocks: Int) -> Int {
+        var blockCount = 0
+        var ptr = _ptr
+        var needBlocks = _needBlocks
+        while needBlocks > 0 && ptr >= 0 {
+            if gasUsedRatio[ptr] == 0.0 || gasUsedRatio[ptr] > 0.9 {
+                break
+            }
+            ptr -= 1
+            needBlocks -= 1
+            blockCount += 1
+        }
+        return blockCount
+    }
+}
+
+public struct Eip1559FeeOracleResult: Equatable {
+    public let maxFeePerGas: BigUInt
+    public let maxPriorityFeePerGas: BigUInt
+}
+
+public extension Int {
+    var hexEncoded: String {
+        return String(self, radix: 16).add0x
+    }
+}
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -722,6 +722,8 @@
 		87AE2736273D701A00E184D4 /* WalletConnectV2Types.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87AE2735273D701A00E184D4 /* WalletConnectV2Types.swift */; };
 		87B1ACD928C0D45B0072A5E2 /* chains.zip in Resources */ = {isa = PBXBuildFile; fileRef = 87B1ACD828C0D45B0072A5E2 /* chains.zip */; };
 		87B1ACDB28C0D55D0072A5E2 /* tokens.json in Resources */ = {isa = PBXBuildFile; fileRef = 87B1ACDA28C0D55D0072A5E2 /* tokens.json */; };
+		87B1ACEA28C0FC4F0072A5E2 /* Eip1559GasSpeedViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B1ACE828C0FC4F0072A5E2 /* Eip1559GasSpeedViewModel.swift */; };
+		87B1AD3228C0FE9E0072A5E2 /* LegacyGasSpeedViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B1AD3128C0FE9E0072A5E2 /* LegacyGasSpeedViewModel.swift */; };
 		87B1AD3728C1FE030072A5E2 /* KeychainStorage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B1AD3628C1FE030072A5E2 /* KeychainStorage.swift */; };
 		87B1B5EE28C870D60072A5E2 /* MixpanelService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B1B5EC28C870D60072A5E2 /* MixpanelService.swift */; };
 		87B1B5EF28C870D60072A5E2 /* AnalyticsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87B1B5ED28C870D60072A5E2 /* AnalyticsService.swift */; };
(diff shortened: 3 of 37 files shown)


oa-s:
@hboon need your help to check if function `EIP1559FeeOracle.eip1559FeeEstimates()` works as needed.

hboon:
@oa-s you need to push your web3swift fork.

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
```diff
@@ -0,0 +1,209 @@
...
+            }
+            ptr -= 1
+            needBlocks -= 1
+            blockCount += 1
+        }
+        return blockCount
+    }
+}
+
+public struct EIP1559FeeOracleResult {
```
hboon, line 191: Probably should rename this as not only used for "oracle" result

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
```diff
@@ -0,0 +1,209 @@
...
+            result[timeFactor] = EIP1559FeeOracleResult(maxFeePerGas: bfbi + tbi, maxPriorityFeePerGas: tbi, baseFee: bfbi)
+        }
+
+        return result
+    }
+
+    private func predictMinBaseFee(baseFee: [BigInt], order: [Int], timeFactor: Double) -> BigInt {
+        assert(!baseFee.isEmpty)
+
+        if timeFactor < 1e-6 {
```
hboon, line 138: This is to check for zero, right? `.isZero` then?

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
```diff
@@ -0,0 +1,209 @@
...
+            let blockCount = maxBlockCount(gasUsedRatio: gasUsedRatio, ptr, needBlocks)
+            if blockCount > 0 {
+                // feeHistory API call with reward percentile specified is expensive and therefore is only requested for a few non-full recent blocks.
+                let lastBlock = firstBlock + BigInt(ptr)
+
+                print("[GasEstimator] lastBlock: \(lastBlock.hexEncoded) for ptr: \(ptr), c: \(BigInt(lastBlock.hexEncoded.drop0x, radix: 16))")
+
+                let feeHistory = try gasService.feeHistory(blockCount, lastBlock: lastBlock.hexEncoded, rewardPercentile: rewardPercentile).wait()
+                let rewardSize = feeHistory.reward?.count ?? 0
+                if rewardSize > 0 {
```
hboon, line 50: Think this `> 0` isn't necessary?

hboon (commented):
- [ ] Do we check or hardcode if a chain supports EIP1559?
- [ ] I tried this by doing an JSON/RPC call with `eth_feeHistory` with `0x64` and hardcoding a `FeeHistory` struct, so I could have gotten it wrong:

The estimates don't look correct. 

Etherscan (https://etherscan.io/gastracker) vs ours:

Screenshot 2022-08-17 at 12 19 36 PM

^ Compare the total 5 gwei on Etherscan's vs our max fee. 2 Gwei is too low

Our MetaMasks:

Screenshot 2022-08-17 at 12 19 51 PM

^ The 5 gwei base fee is much closer to Etherscan's

AlphaWallet/Tokens/Logic/EIP1559FeeOracle.swift
```diff
@@ -0,0 +1,48 @@
...
+    private let server: RPCServer
+    private lazy var gasEstimator: GetFeeHistory = .init(server: server)
+
+    init(server: RPCServer) {
+        self.server = server
+    }
+
+    func eip1559FeeEstimates() -> Promise<EIP1559FeeEstimates> {
+        return gasEstimator.feeHistory(100, lastBlock: "latest", rewardPercentile: [])
+            //NOTE: on .global() as suggestPriorityFee performs sync request call
```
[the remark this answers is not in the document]
oa-s: the function `suggestPriorityFee` makes blocking call of function `feeHistory(blockCount`, so on main thread its going to be blocked

AlphaWallet/Tokens/Logic/GetFeeHistory.swift
```diff
@@ -0,0 +1,32 @@
...
+        self.server = server
+    }
+
+    public func feeHistory(_ blockCount: Int, lastBlock: String = "latest", rewardPercentile: [Int]) -> Promise<FeeHistory> {
+        print("[GasEstimator] feeHistory for blockCount: \(blockCount.toString()), lastBlock: \(lastBlock), rewardPercentile: \(rewardPercentile)")
+        guard let web3 = try? getCachedWeb3(forServer: server, timeout: 6) else {
+            return Promise(error: Web3Error(description: "Error creating web3 for: \(server.rpcURL) + \(server.web3Network)"))
+        }
+
+        let promise: Promise<FeeHistory> = firstly {
```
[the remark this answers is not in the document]
oa-s: missed that case, will update code

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
oa-s: used the same name as in android impl

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
oa-s: moved android implementation of  `SuggestEIP1559` into swift

oa-s:
> Do we check or hardcode if a chain supports EIP1559?
nope, in case chain when doesnt support EIP1559 the fee history request will return an error

AlphaWallet/Tokens/Logic/EIP1559FeeOracle.swift
```diff
@@ -0,0 +1,48 @@
...
+                }
+            }
+            .compactMap(on: .main, { elems -> EIP1559FeeEstimates? in
+                guard !elems.isEmpty else { return nil }
+
+                var estimates: [TransactionConfigurationType: EIP1559FeeOracleResult] = [:]
+                estimates[.rapid] = elems[0].value
+                estimates[.fast] = elems[5].value
+                estimates[.standard] = elems[10].value
+                estimates[.slow] = elems[elems.count - 1].value
```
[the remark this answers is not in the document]
oa-s: it returns an array of 15 elems, otherwise function failures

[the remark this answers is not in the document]
hboon: Right, but I mean `elems.last` is clearer than `elems[elems.count - 1]`.

AlphaWallet/Tokens/Logic/SuggestEIP1559.swift
oa-s: the is for loop right after this check `for index in 0 ..< rewardSize {`  in case if there is 0 in rewardSize, it will cause a crash.

hboon: If `rewardSize` is 0, it will skip the for-loop, right?

oa-s: Yep

oa-s:
continue after https://github.com/AlphaWallet/alpha-wallet-ios/pull/5246

hboon:
@oa-s don't forget this

oa-s:
> should finish up this

here we got an issue with calculating estimate values, have no idea how to fix it

hboon:
Would you help rebase and point out where it is? Can help to see too, if you like

oa-s:
> Would you help rebase and point out where it is? Can help to see too, if you like

yes, sure

oa-s:
@hboon bump

hboon:
@oa-s ah. Should I look at any particular place?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEIP1559.swift
```diff
@@ -0,0 +1,284 @@
...
+        case baseFeeIsEmpty
+        case firstBlockMissing
+        case failureToBuildTimedEstimate
+        case indexedRewardNotFound
+        case rewardsNotFound
+        case indexedBaseFeeNotFound
+        case minBaseFeeNotFound
+    }
+
+    func calculateResult(priorityFee: Double, feeHistory: FeeHistory, gasService: GetFeeHistory) throws -> [Int: EIP1559FeeOracleResult] {
```
oa-s, line 85: @hboon looks like this function returns incorrect result

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/EIP1559FeeOracle.swift
```diff
@@ -0,0 +1,65 @@
...
+                    print("[GasEstimator] \(each.key) - baseFee: \(each.value.baseFee), maxFeePerGas: \(each.value.maxFeePerGas), maxPriorityFeePerGas: \(each.value.maxPriorityFeePerGas)")
+                }
+            }.compactMap(on: queue, { elems -> EIP1559FeeEstimates? in
+                guard !elems.isEmpty else { return nil }
+
+                var estimates: [TransactionConfigurationType: EIP1559FeeOracleResult] = [:]
+
+                //array of 15 elems split on 4 chunks for rapid take first elem, and for the rest the last elem in chuncked array
+                let chunked = elems.chunked(into: TransactionConfigurationType.sortedEip1559FastestFirst.count)
+                for each in zip(chunked, TransactionConfigurationType.sortedEip1559FastestFirst) {
```
oa-s, line 50: and might be also here

hboon:
@oa-s okie, I'll take a look. Thanks!

hboon:
@oa-s sorry, stuck with some other work, will come back to this :)

oa-s:
@hboon i will push after rebase, fifixed signing tx, and some other issues, still have issue with estimation coputation

oa-s:
> @oa-s thanks. Sorry, haven't come to look at this yet. I will, but I will let you know before I do in case you work on this.



> @oa-s thanks. Sorry, haven't come to look at this yet. I will, but I will let you know before I do in case you work on this.

yes, sure

hboon:
@oa-s what else is left for this PR?

oa-s:
tweak changing fees when sending a tx

oa-s:
+ handling fees updates

oa-s:
@hboon could u please take a look at editing gas fee, not quite sure how to done it

hboon:
@oa-s OK, I'll check the next 2 days. Working on something at the moment. Can you help paste a permalink to show where to look (if you know already)?

hboon:
Getting a warning after I rebase master (not pushed here):

> [R.swift] project.pbxproj is internally inconsistent.

Probably a bad merge with the project file

hboon:
Pushed 2 commits that fixes the project file. They can be squashed into your commit(s)

hboon:
Just notes:

Sent an EIP-1559 transaction on Goerli https://goerli.etherscan.io/tx/0x6fbe4f273f7c0ef419e8e86ca7f3c6e5d38f85bbe06cb5a4dd7a1fb03bbdcaab

Was successful and gas price = 284.x gwei, much higher than a transaction from 2 days ago https://goerli.etherscan.io/tx/0x598d15adc5cd894ab82b2c42cfca2516d1f39a365cea1110c83bf2e2dc4ba9b1

But seems inline with a transaction around the same period (281 gwei) https://goerli.etherscan.io/tx/0xe17bb1df31f254db3d1336dd5312ebfc5b49b8b1160852330498610b99507099

oa-s:
I am finishing with applying updates to ui, will test it more, and we will be able to finish it

oa-s:
What about estimates, do they seem correct? Sometimes i was able to catch zero estimates, on mainnet. Looks like that happened because of zero base fee. Not sure on what step it setted to zero

hboon:
- [ ] Looks like `TransactionConfigurator` isn't destroyed after `TransactionConfirmationCoordinator` is destroyed by cancelling the transaction confirmation actionsheet

- [ ] With EIP-1559 transactions, we still need the old "Gas Limit". So 2 sliders for pre-EIP-1559 and 3 sliders for EIP-1559

Screenshot 2023-04-07 at 7 24 26 PM

hboon:
> What about estimates, do they seem correct? Sometimes i was able to catch zero estimates, on mainnet. Looks like that happened because of zero base fee. Not sure on what step it setted to zero

I'll play with it a bit more and look at the estimation function. No idea how it's supposed to work, but I'll check

oa-s:
Ok, I will check

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/Eip1559FeeOracle.swift
```diff
@@ -0,0 +1,49 @@
...
+    private let blockchainProvider: BlockchainProvider
+
+    init(blockchainProvider: BlockchainProvider) {
+        self.blockchainProvider = blockchainProvider
+        self.suggestEip1559 = SuggestEip1559(blockchainProvider: blockchainProvider)
+    }
+
+    func eip1559FeeEstimates(blockCount: Int = 100, block: BlockParameter = .latest, rewardPercentile: [Int] = []) async throws -> GasEstimates {
+        do {
+            let feeHistory = try await blockchainProvider.feeHistory(blockCount: blockCount, block: block, rewardPercentile: rewardPercentile).first
```
hboon, line 23: Seems like these 2 lines both call `BlockchainProvider.feeHistory(blockCount:block:rewardPercentile:)`, once directly and once indirectly. Seems like we only need 1 call?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/Eip1559FeeOracle.swift
```diff
@@ -0,0 +1,49 @@
...
+            for each in GasSpeed.sortedEip1559FastestFirst {
+                guard let estimate = estimates[each], each != .standard else { continue }
+                if estimate == estimates[.standard] {
+                    estimates[each] = nil
+                }
+            }
+
+            return EIP1559FeeEstimates(estimates: estimates)
+        } catch {
+            print("[GasEstimator] \(error)")
```
hboon, line 45: Remember to replace this with `infoLog()` and friends or "xxx"

hboon:
> If you have implemented a wallet sweep action; this is only possible using Legacy transaction style since the exact amount of gas used is not known with 1559 until the transaction is written.

- [ ] How do we handle this? This is referring to the "All Funds" button

hboon:
- [ ] Good to mention and link to https://github.com/komputing/KEthereum/blob/master/eip1559_feeOracle/src/main/kotlin/org/kethereum/eip1559_fee_oracle/EIP1559FeeOracle.kt in the header for `SuggestEip1559`.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
```diff
@@ -0,0 +1,220 @@
...
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 16.08.2022.
+//
+
+import Foundation
+import BigInt
+import PromiseKit
+
+actor SuggestEip1559 {
```
hboon, line 12: Is there a need for this to be an `actor`?

AlphaWallet/Transfer/ViewModels/TransactionConfirmation/CancelTransactionViewModel.swift
```diff
@@ -210,6 +210,10 @@ extension ConfigureTransactionError: LocalizedError {
             return R.string.localizable.configureTransactionErrorGasPriceTooLow()
         case .leaveNonceEmpty:
             return R.string.localizable.configureTransactionErrorLeaveNonceEmpty()
+        case .feePerGasInvalid:
+            return "FeePerGas Invalid"
```
hboon, line 214: Don't forget to localize

hboon (commented):
Sometimes we end up with the priority fee for Rapid being lower than Fast. That shouldn't be possible:

Screenshot 2023-04-10 at 10 18 49 AM

Also, the Max Fee for Rapid is sometimes quite a lot higher than Fast. Is that because there's been a surge in feeds for the last few blocks?

Screenshot 2023-04-10 at 10 21 29 AM

When I tap on Custom and then Save, it shows this error "FeePerGas Invalid":

Screenshot 2023-04-10 at 10 28 13 AM

Also the "Priority Fee" slider goes from 0-4 as discrete Int. So if the user goes here and save, I suppose the Priority Fee would be overridden as 0. Should we change it to a floating point instead?

oa-s:
merge https://github.com/AlphaWallet/alpha-wallet-ios/pull/6647 first, need to resolve future conflicts

oa-s:
> Also the "Priority Fee" slider goes from 0-4 as discrete Int. So if the user goes here and save, I suppose the Priority Fee would be overridden as 0. Should we change it to a floating point instead?

i am trying to use floating point instead of int, but get issue with convertation entered value back to BigInt, 
in case if use type a number `1.23` for example, BigInt still will be 1, any clues how to handle such case?

oa-s:
> Sometimes we end up with the priority fee for Rapid being lower than Fast. That shouldn't be possible:
 can we facing such issue on Android?

hboon:
> > Also the "Priority Fee" slider goes from 0-4 as discrete Int. So if the user goes here and save, I suppose the Priority Fee would be overridden as 0. Should we change it to a floating point instead?
> 
> i am trying to use floating point instead of int, but get issue with convertation entered value back to BigInt, in case if use type a number `1.23` for example, BigInt still will be 1, any clues how to handle such case?

This:

> BigInt still will be 1

Is the unit for priority fee when stored here as BigInt in gwei? `1.23` is only truncated to `1` if we store it as gwei. So we should store it as wei instead (the smallest unit for Ethereum). I think, in general, whenever we store Ethereum amounts in BigInt/BigUInt, they should match the lowest "unit" in Ethereum then we're always safe in this aspect. We only convert them to `gwei` etc when we convert to string or floating point for calculating/storage.

hboon:
> Sometimes we end up with the priority fee for Rapid being lower than Fast. That shouldn't be possible:
> can we facing such issue on Android?

No idea about this one on Android. I can look into the computation if you want. But I was thinking you ported it from the Kotlin version, so I only stared at the UI :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
```diff
@@ -0,0 +1,220 @@
...
+        print("[GasEstimator] order: \(order)")
+
+        var maxBaseFee: Decimal = .zero
+        var result: [Int: EIP1559FeeOracleResult] = [:]
+
+        print("[GasEstimator] baseFee.1: \(baseFee)")
+        print("[GasEstimator] timeFactors: \(Array(Array(0 ..< maxTimeFactor).reversed()))")
+
+        for timeFactor in Array(Array(0 ..< maxTimeFactor).reversed()) {
+            var bf = try predictMinBaseFee(baseFee: baseFee, order: order, timeFactor: timeFactor)
```
hboon, line 131: It feels like there might be a flaw in the computation for `predictMinBaseFee()`. As you go down the list and make calls to it, it generates values in a curve that isn't always increasing (or decreasing, depending on which direction you look at it), so that makes the final estimates not strictly increasing/decreasing too. Maybe it's because the gas usage goes up and down and it doesn't take that into account correctly?

Barring studying and fixing the algorithm (or our implementation of it, which seems to follow the Kotlin version linked to), I think a simple patch is to just store the value of `bf` in each loop and make sure it doesn't go up (or go down, I didn't check which direction)

```
var bf = try predictMinBaseFee(baseFee: baseFee, order: order, timeFactor: timeFactor)
```

eg. 

```
var largestBf = 0 //TODO: need to check if largest or smallest and a sensible default or nil
for timeFactor in (0...maxTimeFactor).reversed() {
    var bf = try predictMinBaseFee(baseFee: baseFee, order: order, timeFactor: timeFactor)
    bf = max(bf, largestBf) //TODO or figure if it should be: min(bf, largestBf)
```

hboon (commented):
Hope that helps?

oa-s:
> Hope that helps? will check, thanks

oa-s:
updated code with latest changes, numbers looks good (not sure)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
```diff
@@ -121,24 +121,24 @@ actor SuggestEip1559 {
 
         print("[GasEstimator] order: \(order)")
 
-        var maxBaseFee: Decimal = .zero
+        var largestBf: Decimal = .zero
```
hboon, line 124: When I read this diff, I realised that maybe this is actually the same variable and your change was how it was supposed to work… even the old name makes sense now and why I wrote a comment about why the variable was unused (but forgot to post) 😆

hboon: @oa-s Do you want to rename it back but otherwise include your changes related to this variable? Easier in case we compare to Kotlin next time

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
```diff
@@ -0,0 +1,220 @@
...
+            if bf > maxBaseFee {
+                maxBaseFee = bf
+            } else {
+                // If a narrower time window yields a lower base fee suggestion than a wider window then we are probably in a price dip.
+                // In this case getting included with a low priority fee is not guaranteed; instead we use the higher base fee suggestion
+                // and also offer extra priority fee to increase the chance of getting included in the base fee dip.
+                t += (maxBaseFee - bf) * extraPriorityFeeRatio
+            }
+
+            print("[GasEstimator] maxFeePerGas: \(bf + t), maxPriorityFeePerGas: \(t), baseFee: \(bf))")
```
hboon, line 144: Don't forget to change to `infoLog()` and friends

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
```diff
@@ -0,0 +1,220 @@
...
+        var maxBaseFee: Decimal = .zero
+        var result: [Int: EIP1559FeeOracleResult] = [:]
+
+        print("[GasEstimator] baseFee.1: \(baseFee)")
+        print("[GasEstimator] timeFactors: \((0...maxTimeFactor).reversed())")
+
+        for timeFactor in (0...maxTimeFactor).reversed() {
+            var bf = try predictMinBaseFee(baseFee: baseFee, order: order, timeFactor: timeFactor)
+            var t = priorityFee
+            bf = max(bf, maxBaseFee)
```
hboon, line 133: This will cause the next line`if bf > maxBaseFee` always be false, right? Think have to tweak it a bit so the if-else is there and the `else` sometimes executes.

Also not sure if it should be `max()` or `min()` (and the `maxBaseFee` name accordingly). Can you help verify again?

hboon (commented):
Otherwise looks ok to me

oa-s:
@hboon do we have any validation rules for `priorityFee`? except safe range `0...4`

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
hboon: > if bf > maxBaseFee it looks like no, i am able to enter here

Oh, you are right.

> updating code with min function returns incorrect values for me

And yes, looks like bf ought to increase (or stay unchanged) with each iteration in the `for-each`, so max is correct.

Thanks for checking :)

hboon:
@oa-s we can look at this in a separate PR. I thought I saw this already done, but it applies for pre-EIP-1559 too

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
hboon: > @hboon do we have any validation rules for `priorityFee`? except safe range `0...4`

I don't know of any.

hboon: But because of the max() this is always 0, right?

```
(maxBaseFee - bf) * extraPriorityFeeRatio
```

So `t` is the same for all the estimates.

hboon: Anyway, if you'd like, we can merge this and keep the Feature flag disabled, then work out the estimation? Then less work rebasing to keep the branch up to date.

oa-s:
>  How do we handle this? This is referring to the "All Funds" button
nope, not for eip1559

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/eip1559/SuggestEip1559.swift
oa-s: > So t is the same for all the estimates.

yes

oa-s:
should be good at this point, e.g. tests seems broken in #6693

hboon:
> tests seems broken in https://github.com/AlphaWallet/alpha-wallet-ios/pull/6693

Fixed :P

@oa-s ok. Let's merge

hboon (approved):

hboon:
Added #6698

PR #5190, opened 2022-08-19
Issue with native swap #5043

### LINKED ISSUE #5043: Issue with native swap
- [x] If I choose MATIC Polygon -> WETH, it says "Swap Pair Not Found".
- [x] MATIC Polygon -> USDC is better, and the Fees/Current Price/Minimum Received appears, but the To amount for USDC remains 0 and "Minimum Received" is "0.0000 USDC" too

If I choose ETH Ethereum -> USDC,

- [x] the fees says "0.1050 ETH". That sounds too high to be correct (my ETH balance is 0.1065, not sure if it's related)
- [x] "Current Price" shows "1 ETH = 0.0000 USDC"

Does the spillage and transaction deadline settings work?

oa-s:
Closes #5043


hboon:
@oa-s does the spillage and transaction deadline settings work (I didn't test those)

hboon (approved):
Seems to work for me when I tried USDC -> ETH. Shall we merge this and have another PR to enable it?

oa-s:
> does the spillage and transaction deadline settings work (I didn't test those)

spillage does, but transaction deadline doesnt

oa-s:
> Seems to work for me when I tried USDC -> ETH. Shall we merge this and have another PR to enable it?

i will double check it, and in case everything is good yes, we can

hboon:
> > does the spillage and transaction deadline settings work (I didn't test those)
> 
> spillage does, but transaction deadline doesnt

Should remove it then.

oa-s:
@hboon do we have api documentation, maybe there is description of `transaction deadline` parameter?

oa-s:
out of their docs there is no transaction deadline in param

https://docs.li.fi/products/more-integration-options/li.fi-api/requesting-a-quote/token-transfer

hboon:
Remove it then :)

oa-s:
@hboon here we still have an issue with `totalFee` calculation, looks like its incorrect as well as `currentPrice`.
i suppose for `totalFee` we should use `swapQuote.estimate.feeCosts` models from api response.

oa-s:
here is a description of api response:
https://docs.li.fi/products/more-integration-options/li.fi-api/requesting-a-quote/token-transfer

oa-s:
in response we got an empty array `swapQuote.estimate.feeCosts`, not sure, but maybe because of some of parameter.

oa-s:
@hboon just checked an interface from lifi, looks a bit different from our:
https://docs.li.fi/products/more-integration-options/li.fi-embeddable

https://transferto.xyz/swap?fromChain=eth&fromToken=0x0000000000000000000000000000000000000000&toChain=eth&toToken=0x41f7b8b9b897276b7aae926a9016935280b44e97&fromAmount=1

hboon:
They seemed to have changed their UI recently. Does the elements still match?

oa-s:
i haven't seen prev design, so can't answer

oa-s:
but looks like they use `gas cost` instead of `fees`

oa-s:
i am not quite sure how to use `feeCosts` and `gasCost`, for now i am able to receive `gasCost` with some data, looks like it describes step in swap

hboon:
@oa-s OK. Would you help list out what you want me to check/verify if there's any? :) Or if you are just posting as you find things and figure it out, that's fine with me too.

oa-s:
sure, will update once have some

oa-s:
> No. That’s for cross-chain bridging right? We don’t support it yet. But once swaps works in production for a while, we’ll look at implanting bridging too

ok, i got it

hboon:
> @hboon any ideas how we can get right value of `currentPrice` we need rate `1 ETH == X USDT`?

I suppose you already figured that out? :)

oa-s:
> > @hboon any ideas how we can get right value of `currentPrice` we need rate `1 ETH == X USDT`?
> 
> I suppose you already figured that out? :)

i found the reason of incorrect calculation, fixed it out

PR #5205, opened 2022-08-23
Add dark mode to AccountsViewController

eviltofu:
Will remove commit for appearance -> automatic before merging.

image image


hboon (approved):

hboon:
> Will remove commit for appearance -> automatic before merging.

Better if you don't include that commit in the PR; at least for simpler ones. I sometimes switch between `Dark` and `Light` when testing out PRs related to dark mode any way. And I can then merge the PR as-is right after review.

eviltofu:
> > Will remove commit for appearance -> automatic before merging.
> 
> Better if you don't include that commit in the PR; at least for simpler ones. I sometimes switch between `Dark` and `Light` when testing out PRs related to dark mode any way. And I can then merge the PR as-is right after review.

You will have to change the Appearance -> Automatic thing yourself then. I will do so for the next PR.

PR #5206, opened 2022-08-24
Tweak log and logging levels

hboon:
Tweaked our logging so:

* verboseLog() — like debug, but not want to show it by default (since we configure our loggers to `.debug`) because it's too verbose/noisy. This includes logging in `deinit{}` and things that happen too frequently, yet not useful (eg. skipping TokenScript signature verification)
* debugLog() — like info, but more for debugging purposes. This is our default logging level for console, so must not be so excessive that the console is unreadable
* infoLog() — should be the most common. Refreshes, etc. This is our default logging level for log files (currently unused, but we'll let user attach them to support emails in the future)
* warnLog() - unexpected errors that are unhandled, mostly. So we can look out for these and fix/handle them
* errorLog() — not used

cc @eviltofu


hboon:
Force-pushed after tweaking commit message

oa-s (approved):

PR #5235, opened 2022-08-29
Tweak slippage value for swap #5233

### LINKED ISSUE #5233: "Unable To Build Swap Unsigned Transaction From Swap Provider"
https://discord.com/channels/548357107267928064/801305005286096906/1013325032467988501

IMG_8665 png

oa-s:
Closes #5233

### DIFF
--- a/AlphaWallet/Common/Views/TextField.swift
+++ b/AlphaWallet/Common/Views/TextField.swift
@@ -144,7 +144,7 @@ class TextField: UIControl {
     }
 
     private (set) lazy var heightConstraint: NSLayoutConstraint = {
-        heightAnchor.constraint(equalToConstant: ScreenChecker().isNarrowScreen ? 30 : 50)
+        heightAnchor.constraint(equalToConstant: DataEntry.Metric.TextField.Default.height)
     }()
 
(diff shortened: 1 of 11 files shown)


oa-s:
We might need to update ui with displaying some warning when user types slippage value, too heigh value might cause `Unable To Build Swap Unsigned Transaction From Swap Provider` error

oa-s:
> I’ll think of something, can we figure out a maximum slippage value to check against before posting the transaction?

comment from Max: 

`The slippage is supposed to be a floating point number. While 0.05 equals 5% slippage, so I am sure you want to pass 0.01 to get 1% slippage. We will block values passed higher than 0.1 to avoid that an unrealistic high slippage is passed by accident.`

hboon:
> not sure what message display to user

I guess we can't change it. All we can do is hope to catch things like the slippage thresholds ourselves before submitting it and hoping they show better error messages. We should feedback to them as we find them so they will improve the messages though.

oa-s:
@hboon is there something we need to change in this pr?

hboon:
@oa-s not that I know of.  Let's get it merged in and we can test+tweak again?

PR #5238, opened 2022-08-30
Group logic components in core folder #5234

### LINKED ISSUE #5234: Make a build so that the artifacts (jars for Android, frameworks for iOS) and their dependencies can be included in another project

```
import SDK
```

Goals
---
- [ ] Check feasibility

oa-s:
Closes #5234


oa-s:
Too many files, might be missed some, no functional changes, just files grouping

hboon (approved):
@oa-s OK. Too big to review. Let's merge it as soon as it build and passes tests (at least locally).

oa-s:
> @oa-s OK. Very big indeed. Let's merge it as soon as it builds and passes tests (at least locally).

ok, i will be ready at least when remove all dependencies from resources (for now strings have left), so we could easily move files to another folder. Apply public interfaces and importing we could make in another pr

hboon:
@oa-s OK, merge ASAP because you might have conflicts from other PRs.

oa-s:
squashed prs, once trevis will finish without errors will merge pr

PR #5246, opened 2022-08-31
remove dependency from resources #5234

### LINKED ISSUE #5234: Make a build so that the artifacts (jars for Android, frameworks for iOS) and their dependencies can be included in another project

oa-s:
Closes #5234


hboon:
Looks like the files must be referred to multiple times in the project, according to Travis?

hboon (approved):
Merge when ready.

oa-s:
> AlphaWalletFoundation

firstly it was `AlphaWallet` but we can't use this name as app extensions and tests can't resolve right link to the app, as app and pod has the same name.

oa-s:
> Merge when ready.

will do that after making interface public, because without it there is over **5k** error

oa-s:
also what about the name for lib? any suggestions, for now when there is no added files it couble be easier to change it. also we need to think about AlphaWallet core, maybe remove it or rename (which is preffered) or left as it is

hboon:
> will do that after making interface public, because without it there is over 5k error

😆

hboon:
> also what about the name for lib? any suggestions, for now when there is no added files it couble be easier to change it. also we need to think about AlphaWallet core, maybe remove it or rename (which is preffered) or left as it is

No idea at the moment. Don't worry about it. Use foundation first. Renaming files if we don't change the contents is still tracked in the history so only the project file will have some changes if/when we rename the pod.

PR #5248, opened 2022-09-01
Fix: chain ID for Palm both mainnet and testnet was displayed as negative number

hboon:
Before|After
-|-
Screenshot 2022-09-01 at 12 16 22 PM|Screenshot 2022-09-01 at 12 15 35 PM


hboon:
Force-pushed after rebase

hboon:

hboon:

hboon:
@eviltofu in case you missed this

PR #5253, opened 2022-09-02
[DEV] dont show swap button in tab bar even when swap feature is enabled unless in debug mode

hboon:
So it doesn't appear in TestFlights or accidentally in app store builds


oa-s (approved):

hboon:

PR #5259, opened 2022-09-02
Remove result pod dependency #5255

### LINKED ISSUE #5255: Foundation refactoring (long running)
- [x] move dependency from KeychainSwift, and SamKaychain to main app, update with protocol definition
- [x] remove dependency from Result pod, update with native Result
- [x] remove warnings
- [ ] Hide some methods under internal level
- [ ] Remove unuesd dependencies

oa-s:
Closes #5255


hboon (approved):
Test fails

PR #5260, opened 2022-09-02
Fix apollo build phase

hboon:
Change the `cd` command so `API.swift` is generated in the right place, from:

```
cd "${SRCROOT}"
```

to:

```
cd "${SCHEMA_PATH}"
```


hboon:
@oa-s seem if it's correct?

oa-s (approved):

PR #5262, opened 2022-09-02
Fix error message with Klaytn KAS RPC node for eth_call: "err: insufficient balance of the fee payer to pay for gas (supplied gas 9223372036854775807) 

### LINKED ISSUE #5261: Klaytn KAS eth_call returns because it doesn't like `gasPrice` passed as 0x0
```
{
  "jsonrpc": "2.0",
  "id": 2,
  "error": {
    "code": -32000,
    "message": "err: insufficient balance of the fee payer to pay for gas (supplied gas 9223372036854775807)"
  }
}
```

hboon:
Fixes #5261

The fix is special cased to Klaytn nodes because we don't know if other RPC nodes rely on gasPrice: 0x0 being around…


oa-s (approved):

hboon:

PR #5267, opened 2022-09-05
Added dark mode to WellDoneViewController

eviltofu:
image 
Light mode

image
Dark mode


eviltofu:
The image needs to be updated for dark mode.

hboon:
> The image needs to be updated for dark mode.

So we have the small version of the icon  for dark mode, but not bigger version(s)? Can you help make a list of all these in a new issue? Then we can collect them and update at one go.

hboon (approved):

hboon:
> > The image needs to be updated for dark mode.
> 
> So we have the small version of the icon for dark mode, but not bigger version(s)? Can you help make a list of all these in a new issue? Then we can collect them and update at one go.

@eviltofu in case you missed this

eviltofu:
Going to see if there are any dark mode logos I can reuse.

PR #5271, opened 2022-09-05
[DEV] Speed up building of Cocoapods

### LINKED ISSUE #5270: Speed up development build time
Especially with `AlphaWalletFoundation` taking ~40s, total app taking ~60s for no changes.

hboon:
Closes #5270

Before the PR, compiling AlphaWalletFoundation (477 swift files) as part of building the app without changes takes 41.4 seconds, so project takes 60s

After the PR,

If we set SWIFT_OPTIMIZATION_LEVEL to -0none for AlphaWalletFoundation, time taken for the pod goes down to ~3s

If we set SWIFT_OPTIMIZATION_LEVEL to -0none for all pods (debug builds on), time taken for the entire app goes down to ~12-20s (I don't know why there's such a big variance), compared to ~20s before the introduction of AlphaWalletFoundation

Not the scientific with many runs, but it ought to be significantly faster :)


oa-s (approved):

oa-s (approved):

hboon:
Force-pushed after rebase

PR #5290, opened 2022-09-06
update links to foundation pod library, add example project, remove local modules folder #5234

### LINKED ISSUE #5234: Make a build so that the artifacts (jars for Android, frameworks for iOS) and their dependencies can be included in another project

oa-s:
Closes #5234


hboon:
Wow. I’ll take a look tomorrow. But could you help summarise what this is about? Just in case I don’t catch it and miss a cycle.

oa-s:
PR updated pod resources urls for AlphaWallet/Core/Foundation/ENS/OpenSea that were moved to a new public repo https://github.com/AlphaWallet/AlphaWalletFoundation

no functional changes in any of pods source files.

AlphaWalletFoundation repo contains the next pods:
>   pod 'AlphaWalletFoundation'
>   pod 'AlphaWalletAddress'
>   pod 'AlphaWalletCore'
>   pod 'AlphaWalletGoBack'
>   pod 'AlphaWalletENS'
>   pod 'AlphaWalletOpenSea'

+ example project (example project could be updated with more examples, for now there is just lib usage)

hboon:
> example project (example project could be updated with more examples, for now there is just lib usage)

Where is the example project? Is it still the AlphaWallet app? Can you help create a separate repo for the project? Cleaner and easier to understand (so that new repo should have a few files only). And there's a good chance it can evolve separately — if we stop work on that, it wouldn't affect the main repo

> PR updated pod resources urls for AlphaWallet/Core/Foundation/ENS/OpenSea that were moved to a new public repo https://github.com/AlphaWallet/AlphaWalletFoundation

Is this because the example project is in the same repo and replaces AlphaWallet?

I don't understand this change:

From:

```
pod 'AlphaWalletAddress', :path => 'modules/AlphaWalletAddress'
   pod 'AlphaWalletCore', :path => 'modules/AlphaWalletCore'
#…
```

to:

```
pod 'AlphaWalletAddress', :git => 'https://github.com/AlphaWallet/AlphaWalletFoundation.git', :commit => '8807643db7d709297ac185f2c5f13c7121c1817c'
   pod 'AlphaWalletCore', :git => 'https://github.com/AlphaWallet/AlphaWalletFoundation.git', :commit => '8807643db7d709297ac185f2c5f13c7121c1817c'
#…
```

Is it because we are trying to repurpose the AlphaWallet repo for the POC driver project?

---

I thought we would do this:

* Continue to refer to module/AlphaWalletFoundation in the AlphaWallet repo's `Podfile` like we do for module/AlphaWalletCore etc. So we don't to create https://github.com/AlphaWallet/AlphaWalletFoundation.git
* Create a new repo which contains an iOS app that has a `Podfile` that refers to `https://github.com/AlphaWallet/alpha-wallet-ios/module/AlphaWalletFoundation` (or whichever pods) (we don't have to worry about pod version)
    * eg. pod 'AlphaWalletFoundation', :git=>'https://github.com/AlphaWallet/alpha-wallet-ios/module/AlphaWalletFoundation', :commit => 'xxx' #maybe this is not possible and that's why we have https://github.com/AlphaWallet/AlphaWalletFoundation.git ?
    * This new repo includes that few lines of pseudo that access the wallet functionality

This new repo would demonstrate how an app would integrate with our UI-less SDK (actually we want a library to wrap our library, but I forgot, but I think it's close, so we can try that later)

Is this what you are trying to do?

oa-s:
> Where is the example project? Is it still the AlphaWallet app?

Example project is located in alphaWallet foundation repo, just like any other pod, has its own example project.
link: https://github.com/AlphaWallet/AlphaWalletFoundation/tree/master/Example

hboon:
> Where is the example project? Is it still the AlphaWallet app?
> Example project is located in alphaWallet foundation repo, just like any other pod, has its own example project.
> link: https://github.com/AlphaWallet/AlphaWalletFoundation/tree/master/Example

Oh I see now. One thing not good about this is we need to move AlphaWalletFoundation into a separate repo https://github.com/AlphaWallet/AlphaWalletFoundation. It makes development more tedious (because we need to modify AlphaWalletFoundation quite often).

How about this, can we create a new pod `AlphaWalletSDK`. And make a https://github.com/AlphaWallet/AlphaWalletSDK repo. Remove https://github.com/AlphaWallet/AlphaWalletFoundation. So:

* Our AlphaWallet app repo gets restored to how it was before this PR
* The new AlphaWalletSDK pod has dependencies on the modules in AlphaWallet repo, including AlphaWalletFoundation (and maybe AlphaWalletFoundation)
* AlphaWalletSDK includes the sample project (similar to how this PR currently does it with the example in AlphaWalletFoundation)

Is this possible?

oa-s:
i got your idea, but i not quite sure if it makes sence to make pod for AlphaWalletSDK.
i case when we add dependency for `AlphaWalletSDK` to AlphaWalletFoundation/Core/Ens/OpenSea will we be able to see types, in added project. e.g
our target project `X`, 
we add `pod AlphaWalletSDK` (sdk actually have no files, just have dependency from alphaWallet)
and if we do like import AlphaWalletSDK, will be be able to actual see types

hboon:
You mean 3rdPartyX -> AlphaWalletSDK -> AlphaWalletFoundation

Anything public in AlphaWalletFoundation is accessible by 3rdPartyX? That's OK. We'll just not document them :)

If so, the benefit over 3rdPartyX -> AlphaWalletFoundation is:

* we can publish AlphaWalletSDK properly as a CocoaPod (and maybe SPM in the far future) when it's ready
    * 3rd party can just watch that repo and not have much noise
    * Documentation is also clearer and targeted for them
* our development over in AlphaWallet app repo (which contains AlphaWalletFoundation) is not affected

Makes sense?

oa-s:
> Makes sense?

looks ok to me, lets try it

PR #5301, opened 2022-09-08
move pod specs for `foundation, ens, core, opensea, goBack` libraries #5234

### LINKED ISSUE #5234: Make a build so that the artifacts (jars for Android, frameworks for iOS) and their dependencies can be included in another project

oa-s:
Closes #5234


oa-s:
pr allows pods usage with link like:

`pod 'AlphaWalletCore', :git => 'https://github.com/AlphaWallet/alpha-wallet-ios.git'`

oa-s:
we got example project that is using linked pods https://github.com/AlphaWallet/AlphaWalletExample

hboon (commented):
I'm not sure this is good, it moves all the `.podspec` out and somewhat pollutes the project root directory; but we can go with it since it still doesn't affect day to day development. Interesting approach!

But for the `AlphaWalletExample`, is it still possible to have an additional `AlphaWalletSDK` either in the `alpha-wallet-ios` repo  or a separate repo so that `AlphaWalletExample` only needs to import 1 (AlphaWalletSDK) or 2 pods? Then we don't need to expose so much implementation details to `AlphaWalletExample` and also much less easy to break those projects using AlphaWallet as an SDK.

AlphaWalletAddress.podspec
```diff
@@ -20,7 +20,7 @@ Pod::Spec.new do |s|
   s.swift_version    = '4.2'
   s.platform         = :ios, "13.0"
   s.source           = { :git => 'git@github.com:AlphaWallet/alpha-wallet-ios.git', :tag => "#{s.version}" }
-  s.source_files     = 'AlphaWalletAddress/**/*.{h,m,swift}'
+  s.source_files     = 'Modules/AlphaWalletAddress/AlphaWalletAddress/**/*.{h,m,swift}'
```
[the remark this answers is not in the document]
oa-s: > Can you make these modules (lowercase "m") instead of Modules? I think the latter will break for case-sensitive file systems.

it should be possible, i'l try

oa-s:
> I'm not sure this is good, it moves all the .podspec out and somewhat pollutes the project root directory; 

i didn't find any other way to link pod, that is licated in child directory.

oa-s:
> But for the AlphaWalletExample, is it still possible to have an additional AlphaWalletSDK either in the alpha-wallet-ios repo or a separate repo so that AlphaWalletExample only needs to import 1 (AlphaWalletSDK) or 2 pods? 

i am not quite sure about types visibility, i mean when our pod `AlphaWalletSDK` has dependency from `core, foundation, ...` we are not able to see types in Example project, if we import only `AlphaWalletSDK`

oa-s:
@hboon do i need to close this pr, or we will go with it?

hboon:
> @hboon do i need to close this pr, or we will go with it?

Let's leave it open first, depending on how/if we do AlphaWalletSDK as a separate repo.

hboon:
> i am not quite sure about types visibility, i mean when our pod AlphaWalletSDK has dependency from core, foundation, ... we are not able to see types in Example project, if we import only AlphaWalletSDK

This sounds good actually, for encapsulation — So it'll build, we'll just to either "re-export" the times in AlphaWalletSDK, maybe through `typealias` or some class wrappers? Is that correct?

Podfile
```diff
@@ -33,12 +33,12 @@ target 'AlphaWallet' do
   pod 'PaperTrailLumberjack/Swift'
   pod 'Charts'
   pod 'CocoaLumberjack', '3.7.0'
-  pod 'AlphaWalletAddress', :path => 'modules/AlphaWalletAddress'
-  pod 'AlphaWalletCore', :path => 'modules/AlphaWalletCore'
-  pod 'AlphaWalletGoBack', :path => 'modules/AlphaWalletGoBack'
-  pod 'AlphaWalletENS', :path => 'modules/AlphaWalletENS'
-  pod 'AlphaWalletOpenSea', :path => 'modules/AlphaWalletOpenSea'
-  pod 'AlphaWalletFoundation', :path => 'modules/AlphaWalletFoundation'
+  pod 'AlphaWalletAddress', :path => '../alpha-wallet-ios'
```
hboon, line 36: Will this work too: `:path => '.'` work?

hboon (approved):

Podfile
oa-s: will check

oa-s: yes, it works!, updating with using `'.'`

PR #5307, opened 2022-09-09
Move kingfisher dependency back to aw project #5306

### LINKED ISSUE #5306: Move some types back to aw project

oa-s:
Closes #5306


hboon (approved):
This is necessary because they belong to the app, but not the AlphaWallet logic?

hboon:
If so, better to mention the "why" in the commit message and PR title

oa-s:
> This is necessary because they belong to the app, but not the AlphaWallet logic?

yes, there some types, those actually related to the app, and not its functionality that was moved to the pod

PR #5324, opened 2022-09-10
update string comparison for ticker ids with x.compare(y, options: .caseInsensitive) == .orderedSame #5323

### LINKED ISSUE #5323: Fix crash on getting localized string for ticker id

oa-s:
Fixes #5323


hboon (commented):
Is this to fix the Crashlytic crash report earlier about tickerId ? Isn't it basically the same code?

oa-s:
i am not quite sure, maybe there underhood some logic for localizedLowercased property, that causes a crash. Just want to check, maybe it could help.

hboon (approved):

Issue #5330, opened 2022-09-12
Wrong error message when it should be "You already added this address to wallets" when importing a wallet again


hboon:
<img width="487" alt="Screenshot 2022-09-12 at 12 03 13 PM" src="https://user-images.githubusercontent.com/56189/189586042-6183421b-eb43-4d1f-b788-a5661de631cc.png">

PR #5331, opened 2022-09-12
Fix: Wrong error message when it should be "You already added this address to wallets" when importing a wallet again 

hboon:
Fixes: #5330. Includes a test. Hard to catch it otherwise.

@oa-s This seems like it's related to us adding localized strings in the app, outside of AlphaWalletFoundation and it not loading correctly. Do you think there are other cases like this?

Before PR|After PR
-|-
Screenshot 2022-09-12 at 12 03 13 PM|Screenshot 2022-09-12 at 2 00 29 PM


hboon:
Hold merging. I must have mixed up the branches

hboon:
Superseded by #5332 (this PR was on the wrong branch)

PR #5359, opened 2022-09-13
Replace passing OpenSea with NftProvider, hide default implementation of init method #5358

### LINKED ISSUE #5358: Replace passing OpenSea with NftProvider, hide default implementation of init method

oa-s:
Closes #5358


hboon (approved):
👍

PR #5393, opened 2022-09-15
[Refactoring] Group `web3swift` types in own `Web3` type, remove some imports of `web3swift`

oa-s:


hboon (approved):
Would be good to go even further to have a separate inhale pod that wraps web3swift as separate PR.

PR #5399, opened 2022-09-16
Display PHI NETWORK Price #4890

### LINKED ISSUE #4890: Display PHI NETWORK Price
Need to add PHI Network Price API

here is the api example https://price.phi.network/api/ticker?filter=WPHI

pulls data from subgraph https://graph.phiswap.com/subgraphs/name/exchange/graphql

from analytics here https://info.phiswap.com

### LINKED ISSUE #5433: Check if need to apply PHI v1 price to PHI v2
* After #5429 and #5399

oa-s:
Closes #4890, Closes #5433


modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Phi/HpiNetworkProvider.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  HpiNetworkProvider.swift
```
hboon, line 2: Hpi -> Phi :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/BaseCoinTickersFetcher.swift
```diff
@@ -0,0 +1,172 @@
...
+
+        storage.addOrUpdate(tickers: tickers)
+    }
+
+    public func fetchTickers(for tokens: [TokenMappedToTicker], force: Bool = false) {
+        let targetTokensToFetchTickers = tokens.filter {
+            if promises[$0] != nil {
+                return false
+            } else {
+               return force || hasExpiredTickersLifeTimeSinceLastUpdate(for: $0)
```
hboon, line 52: I always see the price displayed $0. Maybe the check in this line is faulty?

hboon (changes requested):

modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/BaseCoinTickersFetcher.swift
oa-s: oh, i see, will check it. for me it was an empty db of tickers, so might be missed this case.

oa-s: @hboon can u please share shared.realm file in dm. want to check if ticker exists

hboon: Sent

oa-s: thanks, have u check latest pr on your side, i checked realm, it has ticker for phi native coin, so it should applear, works for me.

hboon: No graphs appear for me.

Can you help write a Realm query that I can type in Realm Studio so I can trace which object should match? eg. `name = 'PHI'` ?

oa-s: oh you meant that there is no chart for phi token? there is no api call for retrieving prices history, so it doesn't work, ticker price only

oa-s: missed to mention this in pr description

hboon: Oh. Ok. But my price is always $0. Maybe it’s expected when balance = 0? I forgot to check with a non-zero balance

hboon: Ah I see what I was not finding now, we should show the current unit price for PHI (or any chain) as long as we have it, even if we don't have historical data.

From:

Screenshot 2022-09-21 at 11 22 37 AM

This is OK, but tap on the "Value: 3,131…" label and it should show the market price even if we don't have the historical data. Maybe just hide the appreciation/depreciation if it's not available? eg. "Market Price: 1.00 USD". Compare:

PHI's without current price|Gnosis (correct, with price)
-|-
Screenshot 2022-09-21 at 11 22 38 AM|Screenshot 2022-09-21 at 11 26 01 AM

oa-s: try latest commits, its fixed there

hboon: Ah, looks good.

hboon (approved):

PR #5429, opened 2022-09-22
Add `RPCServerWithEnhancedSupport` to dramatically reduce number of files and changes required to add most new chains

hboon:
Contains 2 commits:

* 1st commit contains the changes for this PR
* 2nd commit is to show how many files and changes are needed to add a basic chain (4 files, but still many lines of changes) — will be removed before merging

With this PR, when we need to add a new chain that doesn't have more fancy features and exceptions, it should only touch 4 files instead of 20 (!) files. I've also noticed that since we moved lots of code out into pods, including `AlphaWalletFoundation`, when we add a new case to `RPCServer` and build, it no longer shows all the places which needs to be changed (i.e. with errors) like it used to. We have to change and build, repeat a few times to sweep through all the files, making it even slower.

@oa-s not sure if this approach is good. Thoughts?

Edit: have to change 4 files, not 2 files :)

### DIFF
--- a/AlphaWallet/Common/Services/LocalNotificationService.swift
+++ b/AlphaWallet/Common/Services/LocalNotificationService.swift
@@ -22,10 +22,10 @@ class LocalNotificationService: ScheduledNotificationService {
         let notificationCenter = UNUserNotificationCenter.current()
         let content = UNMutableNotificationContent()
         //TODO support other mainnets too
-        switch server {
+        switch server.serverWithEnhancedSupport {
         case .main, .xDai:
             content.body = R.string.localizable.transactionsReceivedEther(amount, server.symbol)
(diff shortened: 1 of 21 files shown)


Phinetwork:
Please be sure the v2 chain is chain id 144 & explorer phiscan.com the chain id and explorer is different than
V1 chain. Just want to make sure this is clear.

Phinetwork:
> 
> >Please be sure the v2 chain is chain id 144 & explorer phiscan.com the chain id and explorer is different than ...
> 
> @Phinetwork yeah. Thanks for the heads-up. The PHI 2 changes here are just temporary placeholders first. 

Ok great just wanted to make sure. Appreciate your help @hboon your awesome.

hboon:
Force-pushed so second commit (which belongs to a separate PR) completes PHI 2 integration

oa-s (approved):

hboon:
Force-pushed after rebase master and removing the PHI 2 commit. Going to merge.

PR #5430, opened 2022-09-23
Swap upgrade: allow users to filter allowed exchanges + select from multiple quotes #5418

### LINKED ISSUE #5418: Swap upgrade: allow users to filter allowed exchanges + select from multiple quotes
Two major upgrades:
1. Allow users to select from exchanges via Settings. The available exchanges are available via `/tools`. In Android, we copied and saved locally + modified to add the exchange website/URL. We then display and link out to this URL to provide the user some more information, as some exchanges are not as popular.

2. Provide quotes from these selected exchanges via [`/advanced/routes`](https://apidocs.li.fi/reference/post_advanced-routes). Note that this is a `POST` request

After user selects a quote, you can use the exchange `key` as parameter to the `/quote` endpoint (`&allowExchanges=`)

----
Please check out Android PR to see UX.
Disclaimer: UX is not final - need Tomek's input/approval
Link: https://github.com/AlphaWallet/alpha-wallet-android/pull/2835

oa-s:
Closes #5418


hboon:
@oa-s I need to check something else so can't look into this at the moment, if you think it's ok, let's merge it in first, since it's still disabled in prod.

PR #5438, opened 2022-09-24
remove unused remark from locale files - fix issue #5415

### LINKED ISSUE #5415: Verify if we don't check for min 5 characters, if we aren't checking, remove the remark on the screen
image

Mac3n:
Closes #5415

1-liner summary:
I had checked for limit validation on the name and there was no validation so that we could remove the unused remark.


hboon (approved):

hboon:
@Mac3n thank you!

PR #5445, opened 2022-09-26
Switch to using Infura for IPFS gateway so files pinned with Infura do not need to wait 45-60s before they are available via IPFS (we were accessing via ipfs.io)

### LINKED ISSUE #5443: TokenScript files pinned to IPFS via Infura is taking a long time (50-60s instead of 1s) to download via ipfs.io the first time
While it might generally not matter if the performance is just worse for the first hit, it's a problem for us because we pin via Infura and access the URL almost immediately. So changing it to access via Infura IPFS gateway is a big usability improvement

hboon:
Closes #5443


oa-s (approved):

hboon:
Force-pushed after rebaase master and fixing tests

PR #5463, opened 2022-09-28
move removing realm file on apps launch #5460

### LINKED ISSUE #5460: Adding a wallet after removing it in the same session doesn't save it
Tested with watching wallets, but probably affects all types. Tested on simulator.

1. Note number of watched wallets N
2. Watch wallet 0xa392DD3B76Ae38EB8d382Ec4087627fc1850a532
3. Note number of watched wallets N+1
4. Look at Ream folders/file created. Includes 0xa392
5. Switch to another wallet
6. Delete 0xa392
7. Look at Ream folders/file created. 0xa392 has been removed
8. Watch wallet 0xa392DD3B76Ae38EB8d382Ec4087627fc1850a532 again
9. Look at Ream folders/file created. 0xa392 is not created
10. Restart app

Expected:
---
11. Note number of watched wallets N+1

Observed:
---
11. Note number of watched wallets N

oa-s:
Fixes #5460

### DIFF
--- a/AlphaWallet/AppCoordinator.swift
+++ b/AlphaWallet/AppCoordinator.swift
@@ -154,16 +154,14 @@ class AppCoordinator: NSObject, Coordinator {
     private func bindWalletAddressesStore() {
         walletAddressesStore
             .didRemoveWalletPublisher
-            .sink { [weak self] account in
-                guard let `self` = self else { return }
-
+            .sink { [config, analytics, keystore, legacyFileBasedKeystore] account in
                 //TODO: pass ref
                 FileWalletStorage().addOrUpdate(name: nil, for: account.address)
-                PromptBackupCoordinator(keystore: self.keystore, wallet: account, config: self.config, analytics: self.analytics).deleteWallet()
-                TransactionsTracker.resetFetchingState(account: account, config: self.config)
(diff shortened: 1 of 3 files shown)


oa-s:
Removing Realm files not supported, and might cause unexpected behaviour, its recommended to do it on apps launch.
https://stackoverflow.com/questions/45683173/realm-crash-when-invalidating-while-updating

updated code base with removing files on apps start.

hboon (changes requested):
I think it doesn't handle this:

1. Launch app
2. Add wallet W
3. Delete wallet W
4. Add wallet W
5. Optionally switch wallet away from W
6. Re-launch app

Expected
---
7. Wallet W not around

Observed
---
7. Wallet W should be around

Removing W from `DatabaseMigration.walletsToDeleted` in step 4 might fix it? (but we still have to be careful with keys etc)

But this means that when a user does step 4 (add wallet W, again), it is basically an undo functionality right? Can you think of any security implication?

oa-s:
now it should be ok

hboon (approved):

hboon:
@oa-s could you help modify the commit message to be "Fix: Adding a wallet after removing it in the same session doesn't save it" before merging? Better to explain the "why" than "what"

PR #5468, opened 2022-09-29
update retry logic for `getTokenType(_: )` function #5472

### LINKED ISSUE #5465: Mark deprecated Testnets
Refer to [#2854](https://github.com/AlphaWallet/alpha-wallet-android/issues/2854)

### LINKED ISSUE #5472: update retry logic for getTokenType(_: ) function

oa-s:
Closes  #5472

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenProviderType.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenProviderType.swift
@@ -42,7 +42,7 @@ public class TokenProvider: TokenProviderType {
         let queue = queue
 
         return firstly {
-            attempt(maximumRetryCount: numberOfTimesToRetryFetchContractData) {
+            attempt(maximumRetryCount: numberOfTimesToRetryFetchContractData, shouldOnlyRetryIf: TokenProvider.shouldRetry(error:)) {
                 GetNativeCryptoCurrencyBalance(forServer: server, analytics: self.analytics, queue: queue)
                     .getBalance(for: address)
             }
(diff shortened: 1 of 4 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenProviderType.swift
```diff
@@ -146,6 +146,27 @@ public class TokenProvider: TokenProviderType {
         }
     }
 
+    fileprivate static func shouldRetry(_ error: Error) -> Bool {
```
hboon, line 149: Should be `shouldRetryError(_ error: Error)` or `shouldRetry(error: Error)`. More idiomatic?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenProviderType.swift
```diff
@@ -146,6 +146,27 @@ public class TokenProvider: TokenProviderType {
...
+        if let web3Error = error as? Web3.Web3Error {
```
hboon, line 150: Do you want to see if `RpcNodeRetryableRequestError` can be checked here? I'm not sure if `RpcNodeRetryableRequestError` *is* always correct — i.e. if all the enum-cases are indeed retryable (but that's the idea) — though, so better to double check it.

hboon (commented):
Remove to remove the reference to #5465 :)

hboon:
In a future PR, we can probably stop checks for ERC875 and hardcode contracts for them since it's unlikely there will be new contracts based on it (which goes with the ERC165 hash `0x4f452b9a`/isStormBirdContract)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenProviderType.swift
oa-s: not sure about `RpcNodeRetryableRequestError ` here, as its checkes only for APIKit errors and here we can't get it.

hboon: Hmm.. in that case, how about stub out the body of `shouldRetry()` first , return `true` with a `TODO` and merge? I think we might have to go through and be careful with what is retried or not. We are already retrying anyway, but if we code the error types in, we might think that those are the expected ones.

PR #5501, opened 2022-10-04
Add Buy ETH button to bottom of Wallet tab #5398

### LINKED ISSUE #5398: Add Buy ETH button to bottom of Wallet tab
* Only if Ethereum mainnet is enabled
* Works like the "Buy" button when we tap "Buy" after tapping the ETH token > … > Buy

Something like Android's:

oa-s:
Closes #5398

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
@@ -188,6 +188,10 @@ extension UIPasteboard {
 
 extension TokensCoordinator: TokensViewControllerDelegate {
 
+    func buyCryptoSelected(in viewController: UIViewController) {
+        delegate?.buyCrypto(wallet: wallet, server: .main, viewController: viewController, source: .walletTab)
+    }
+
     func whereAreMyTokensSelected(in viewController: UIViewController) {
         delegate?.whereAreMyTokensSelected(in: self)
     }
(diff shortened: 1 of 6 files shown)


AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
```diff
@@ -188,6 +188,12 @@ extension UIPasteboard {
 
 extension TokensCoordinator: TokensViewControllerDelegate {
 
+    func buyCryptoSelected(in viewController: UIViewController) {
+        guard config.enabledServers.contains(.main) else { return }
```
hboon, line 192: Shall we move this condition to control whether the buy button is shown?

hboon (commented):
It clips off the bottom of the table — in the screenshot, there is a GNO token below the xDai token:

Screenshot 2022-10-05 at 11 13 09 AM

AlphaWallet/Tokens/Coordinators/TokensCoordinator.swift
oa-s: checking

oa-s:
@hboon  updated pr, looks ok now

hboon:
Just noticed 1 thing, the background colors of the actionsheet buttons seems a bit strange. But maybe it was always there with the Buy Crypto buttons and not because of this PR?

Screenshot 2022-10-05 at 7 36 08 PM

oa-s:
not sure that its pr changes, guess it was before

hboon:
Yeah. Thought might be it. Let’s  fix it separately.

PR #5510, opened 2022-10-05
Added dark mode to Swap Option on main wallet tab

eviltofu:
Main Wallet Tab > ... > Swap


hboon:

hboon (approved):
@eviltofu In the future, would you layout the screenshots like this in the future? Much easier to read (and it works on mobile):

Before PR|After PR
-|-
 |

PR #5525, opened 2022-10-07
Improve/fix ERC1155 NFT auto-detection

### LINKED ISSUE #5462: Fix NFT auto-detection for Polygon

hboon:
Closes #5462

A chain with short blocktimes and limit on eventlog fetching window like Polygon (~2 seconds + 3500 window) means that fetching events starting from block 0 could take a long time before reaching the current block. As of writing it could take > 6 days of the app running for it to catch up for a fresh install. So ERC1155s will not show up.

With this PR, we always do these 2 steps in a refresh cycle:

A. Fetch start from the latest block (capped by the maximum eventlog window if there is one)
B. then, catch up with the newest window that we haven't fetched

So while we might not detect all the tokens immediately (if there is a eventlog window cap), we will usually be able to fetch recent tokens fairly soon and catch up with older ones. Note that ERC1155 fetching has always been 2-step:

1. Fetch contracts + token IDs
2. Fetch balance for each token ID

So as long as we know the contract + token ID (which is improved with this PR), we will get the balance for that contract + token ID combo (separately).

In the future, we might be more aggressive with step B, leaving step A to just fetch 1 window with each refresh

---

@oa-s I've modified a little how the internal queue is used in `Erc1155TokenIdsFetcher` and I think it should be a good reference. There's this:

```
extension Erc1155TokenIdsFetcher.functional {
    //This is only for development purposes to keep the PromiseKit `Resolver`(s) from being deallocated when they aren't resolved so PromiseKit don't show a warning and create noise and confusion
    private static var fetchEventsPromiseKitResolversKeptForDevelopmentFeatureFlagOnly: [Resolver<[Erc1155TransferEvent]>] = .init()

    private static let queue: DispatchQueue = .global(qos: .utility)
```

We do all work explictly on `.main` in the file and only few select things in `queue` (which is `.utility`). The `queue` is managed by `Erc1155TokenIdsFetcher` only, not passed in and not accessible from outside. We *could* expand it so that a module (possible few files within an inhouse pod) shares the same queue, but same thing, never passed around, and not accessible from outside. Output from functions are always dispatched on `.main` to the callers outside.

This seems like a good balance of speed and tidiness? What do you think @oa-s?

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -321,6 +321,7 @@
 		5E7C79DC5BDC391C983B6CAC /* ChainIdTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C78DB6F47344EBBC2567F /* ChainIdTests.swift */; };
 		5E7C79E9A6A3DFB7FA680752 /* DappViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C793CDFA907BFDFECB6CB /* DappViewCell.swift */; };
 		5E7C79EB8A6EC89F6F033268 /* CreateInitialWalletViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E7C74A5B5B9D8AD0BD913C1 /* CreateInitialWalletViewController.swift */; };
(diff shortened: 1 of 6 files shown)


oa-s (approved):

oa-s:
looks ok to me if it does't block main thread

PR #5536, opened 2022-10-11
[Refactoring] remove browser unsed params, rename placeholders

oa-s:


hboon:

AlphaWallet/Browser/Views/DappBrowserNavigationBar.swift
```diff
@@ -101,7 +101,8 @@ final class DappBrowserNavigationBar: UINavigationBar {
         backButton.tintColor = Configuration.Color.Semantic.navigationbarPrimaryFont
         backButton.adjustsImageWhenHighlighted = true
         backButton.setImage(R.image.toolbarBack(), for: .normal)
-
+        backButton.isHidden = true
```
hboon, line 104: When would we reset `isHidden`? :)

PR #5559, opened 2022-10-12
Add analytics param for publisher version of api kitsession.request

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Extensions/Publishers/Session+Publishers.swift
```diff
@@ -12,13 +12,26 @@ import Combine
...
+                if let e = e as? RpcNodeRetryableRequestError {
+                    logRpcNodeError(e, analytics: analytics)
+                }
                 return e is RpcNodeRetryableRequestError
-            })
-            .eraseToAnyPublisher()
+            }).eraseToAnyPublisher()
+    }
+
+    private static func logRpcNodeError(_ rpcNodeError: RpcNodeRetryableRequestError, analytics: AnalyticsLogger) {
```
hboon, line 26: Is the implementation of `logRpcNodeError(_:,analytics:) `the same as the one in `Session+PromiseKit.swift`? Then we can call it instead?

hboon (approved):

PR #5563, opened 2022-10-12
Fix bad access crash for apollo #5562

### LINKED ISSUE #5562: Fix bad access crash for apollo

oa-s:
Closes #5562


hboon (approved):
Didn't check, presume it's working :)

PR #5564, opened 2022-10-12
Enable Crashlytics for tests, and disable crash collection for tests #5420

### LINKED ISSUE #5420: `The default Firebase app has not yet been configured. FirebaseApp.configure()` still happening when running tests
Was this what #5417 was trying to fix? :)

oa-s:
Closes #5420


AlphaWallet/Common/Services/FirebaseReportService.swift
```diff
@@ -15,14 +15,16 @@ extension AlphaWallet {
...
+                if isAlphaWallet() {
                     FirebaseApp.configure(options: options)
                 }
             }
             return FirebaseCrashlyticsReporter()
         }()
 
-        private init() { }
+        private init() {
+            crashlytics.setCrashlyticsCollectionEnabled(!isRunningTests())
```
hboon, line 26: Not sure it works correctly because the function's comment is:

```
 * The value does not apply until the next run of the app. If you want to disable data
 * collection without rebooting, add the FirebaseCrashlyticsCollectionEnabled flag to your app's
 * Info.plist.
```

If it's too much of a hassle, let's just leave it. It's only printing the warning when running tests right?

oa-s: Updated pr with plist for tests, and disabled collection in plist

AlphaWallet/GoogleService-Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>FirebaseCrashlyticsCollectionEnabled</key>
```
hboon, line 5: Does this file have to be updated? Because I keep a copy of the prod version for submission.

hboon (approved):

AlphaWallet/GoogleService-Info.plist
oa-s: yes, i guess we don't need to upload it

oa-s: in case for tests, what file do we need to use?

hboon: For tests locally and Travis? I think the one you added `AlphaWallet/GoogleService-InfoTests.plist` looks good. Was just wondering why you updated `AlphaWallet/GoogleService-Info.plist` too (because then it might mean I need to update the prod copy)

oa-s: oh i see, updated it with manually setting FirebaseCrashlytics Collection Enabled, to avoid misunderstanding.

PR #5573, opened 2022-10-13
switch wallet connect v1 spm to alphaWallet branch #5570

### LINKED ISSUE #5570: Can't connect using WalletConnect for minting with DevCon ticket attestation
1. Visit the attestation URL in macOS Safari
2. In the same session, click "NFTs" in the top navigation bar
3. Click "Mint Arbitrum One NFT"
4. Dot click "Mint NFT"
5. Click "WALLET CONNECT" (it doesn't seem to matter which chain is enabled at this point because the actionsheet to start the session doesn't appear at all for me)
6. Wait for the session confirmation actionsheet to appear

Don't click "Mint NFT" at any time :)

It times out when I tried it a few times. Also have tried this:

A. It works most of the time on Android AlphaWallet. Fails a few times
B. It works most of the time on another wallet app. Fails a few times
C. It works https://test.walletconnect.org the one time I tried (and accidentally mint 😭 a token on Polygon)
D. AlphaWallet iOS works if I connect to http://example.walletconnect.org
E. It doesn't work even if I comment out the timeout and wait

oa-s:
Closes #5570


hboon (approved):
Commit message should mention to handle`chainId` being both string and int.

PR #5591, opened 2022-10-17
Apply promise caching until it get resolved #5579

### LINKED ISSUE #5579: Evaluate if `CachedERC1155ContractDictionary` and its usage is optimal and improve if necessary
A few considerations:

* Does it cache only results we are sure of (i.e not `false` due to network connectivity issues)
* Should we include `RPCServer` as the key, so it's address+chain

Followups:

- [ ] Rename it so it can be used for other token types
- [ ] Use it for other token types

oa-s:
Closes #5579


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetErc20Balance.swift
```diff
@@ -4,24 +4,41 @@ import Foundation
...
-                guard let balanceOfUnknownType = balanceResult["0"], let balance = BigInt(String(describing: balanceOfUnknownType)) else {
-                    throw CastError(actualValue: balanceResult["0"], expectedType: BigInt.self)
-                }
-                return balance
-            })
+    func getBalance(for address: AlphaWallet.Address, contract: AlphaWallet.Address) -> Promise<BigInt> {
+        firstly {
+            .value(contract)
+        }.then(on: queue, { [weak self, queue, server] contract -> Promise<BigInt> in
+            if let promise = self?.inFlightPromises[contract] {
```
hboon, line 22: Can't cache with just `contract`, I think it needs at least `contract+address/wallet` as the key. Similar for at least some of the rest?

hboon (changes requested):

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetErc20Balance.swift
oa-s: oh, yes right, updating

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/CallSmartContractFunction.swift
```diff
@@ -88,13 +88,13 @@ fileprivate var smartContractCallsCache = AtomicDictionary<String, (promise: Pro
 private let callSmartContractQueue = DispatchQueue(label: "com.callSmartContractQueue.updateQueue")
 //`shouldDelayIfCached` is a hack for TokenScript views
 //TODO should trap 429 from RPC node
-public func callSmartContract(withServer server: RPCServer, contract: AlphaWallet.Address, functionName: String, abiString: String, parameters: [AnyObject] = [], shouldDelayIfCached: Bool = false, queue: DispatchQueue? = nil) -> Promise<[String: Any]> {
+public func callSmartContract(withServer server: RPCServer, contract: AlphaWallet.Address, functionName: String, abiString: String, parameters: [AnyObject] = [], shouldDelayIfCached: Bool = false, queue: DispatchQueue? = nil, useCashed: Bool = true) -> Promise<[String: Any]> {
```
hboon, line 91: useCashed -> useCache

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/CallSmartContractFunction.swift
```diff
@@ -88,13 +88,13 @@ fileprivate var smartContractCallsCache = AtomicDictionary<String, (promise: Pro
...
     let timeout: TimeInterval = 60
     //We must include the ABI string in the key because the order of elements in a dictionary when serialized in the string is not ordered. Parameters (which is ordered) should ensure it's the same function
     let cacheKey = "\(contract).\(functionName) \(parameters) \(server.chainID)"
     let ttlForCache: TimeInterval = 10
     let now = Date()
-    if let (cachedPromise, cacheTimestamp) = smartContractCallsCache[cacheKey] {
+    if let (cachedPromise, cacheTimestamp) = smartContractCallsCache[cacheKey], useCashed {
```
hboon, line 97: Faster to flip the conditions order so `useCashed` comes before `let`? Faster. But is it less readable?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractName.swift
```diff
@@ -3,21 +3,42 @@
...
+        firstly {
+            .value(contract)
+        }.then(on: queue, { [weak self, queue, server] contract -> Promise<String> in
+            let key = contract.eip55String.hashValue
+            if let promise = self?.inFlightPromises[key] {
+                return promise
+            } else {
+                let functionName = "name"
+                let promise = attempt(maximumRetryCount: 2, shouldOnlyRetryIf: TokenProvider.shouldRetry(error:)) {
+                    callSmartContract(withServer: server, contract: contract, functionName: functionName, abiString: Web3.Utils.erc20ABI, useCashed: false)
```
hboon, line 27: But shouldn't we cache `name`, `decimals` even more so than balance?

hboon: Do you think it's better to keep the useCache/useCashed change in a separate PR? Easier to decide. I'm wondering if the function arg should default to `false` instead of `true` (I know before the PR, the lack of the arg means it is implicitly defaulting to `true`)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/Erc1155BalanceFetcher.swift
```diff
@@ -11,27 +11,44 @@ import AlphaWalletWeb3
...
-        return firstly {
-            callSmartContract(withServer: server, contract: contract, functionName: "balanceOfBatch", abiString: AlphaWallet.Ethereum.ABI.erc1155String, parameters: [addresses, tokenIds] as [AnyObject])
-        }.map { result in
-            guard let balances = result["0"] as? [BigUInt], balances.count == tokenIds.count else {
-                throw CastError(actualValue: result["0"], expectedType: [BigUInt].self)
+    func fetch(contract: AlphaWallet.Address, tokenIds: Set<BigInt>) -> Promise<[BigInt: BigUInt]> {
+        firstly {
+            .value(contract)
+        }.then(on: queue, { [weak self, queue, address, server] contract -> Promise<[BigInt: BigUInt]> in
+            let key = "\(contract.eip55String)-\(tokenIds.hashValue)".hashValue
```
hboon, line 29: What's the reason to use the hash as the key instead of the original string? I thought there's a few reasons not to:

* hashing a string key is already done by the dictionary anyway and the keys are short (so they hash fast)
* string keys are more inspectable/debuggable

hboon (commented):
I checked about half the PR. Presume it's similar :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractName.swift
oa-s: > But shouldn't we cache `name`, `decimals` even more so than balance?

do you mean to cache actual values, and not only promises?

oa-s: > Do you think it's better to keep the useCache/useCashed change in a separate PR? Easier to decide. I'm wondering if the function arg should default to `false` instead of `true` (I know before the PR, the lack of the arg means it is implicitly defaulting to `true`)

made it as `true`, as its behaviour by default for function `callSmartContract`, cached promise value might return same result when we do retrying, when retry function called in less than 10s range

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/CallSmartContractFunction.swift
oa-s: will do

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/Erc1155BalanceFetcher.swift
oa-s: yes, i agree with you, made it as `int` because it returns same length, so it might be optimized

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractName.swift
hboon: >> But shouldn't we cache name, decimals even more so than balance?
> do you mean to cache actual values, and not only promises?

No, I mean we should cache name and decimals (and symbols). Doesn't matter if it's with values or promises. But it's not cached here with `useCashed: false`.

hboon: Oh, do you mean setting it to `true` for name/decimals would cache a bad value? If so something is wrong with the caching? (A) If it's a smart contract execution error (for example `symbols` isn't available for ERC1155), then it should/can be cached, but (B) if it's a RPC node timeout or API key error, it shouldn't be cached.

A can be cached, controlled by the `useCashed` flag, B should never be cached.

Is this what you are referring to?

oa-s: it's a bit different of what i meant. when we call `attempt` function for promise retrying we actually get result of closure param for `attempt` function that returns us a promise. In case of `GetContractDecimals` class we will get  promise from call `callSmartContract(` function. In short period of time this function returns same `promise`. since we don't release cached promises in `callSmartContract`  when we get failed result for `callSmartContract` it still cached and the next call of attempt will return same failed promise when it should create a new promise.

hboon: Oh, then it seems like the laying of `attempt()` + `callSmartContract()` needs to be flipped so caching is done correctly (for the result of `attempt()`). Otherwise caching is broken?

hboon: Or it should just not cache in `callSmartContract()` in the case (B) above.

hboon: > Or it should just not cache in callSmartContract() in the case (B) above.

This seems a bit better.

oa-s: but i am not sure about calling function `callSmartContract()`, caching for promises should be there, so that is why there is `useCache` param value.

oa-s: ok, yes, lets leave it without `useCache` flag, and use it as it is for now

PR #5600, opened 2022-10-18
Fixed Dark Mode for Background In ImportWalletViewController

eviltofu:
In main wallet tab screen, tap the upper left corner icon. Then tap the plus symbol on the upper right corner. Select Import. Tap on the SEED PHRASE text view.

Before fix | After fix
-|-
Simulator Screen Shot - iPhone 13 - 2022-10-18 at 12 17 08 copy | Simulator Screen Shot - iPhone 13 - 2022-10-18 at 12 23 48 copy


hboon:
Looks OK. Waiting for a review?

hboon (approved):

PR #5610, opened 2022-10-19
[DEV] Remove Cocoapods caching with Travis because we are using development pods for our inhouse modules

hboon:
If caching is enabled, adding a file to an existing module doesn't modify `Podfile.lock` and breaks the build on Travis


hboon:
See how long it takes first

hboon:
3 minutes more. Not so bad

PR #5613, opened 2022-10-19
Added dark mode to navigationController of Activity tab

eviltofu:
Tap activity on tab bar at bottom of screen.

Before commit | After commit
-|-
before copy | after copy


hboon (commented):
Can you fix the commit message and PR title? Just "Add dark mode to Activity tab". Prefer that to class names which changes more often

eviltofu:
> Can you fix the commit message and PR title? Just "Add dark mode to Activity tab". Prefer that to class names which changes more often

Updated. @hboon

hboon (approved):

hboon:
@eviltofu Why does it not affect Settings tab though?

PR #5615, opened 2022-10-19
Added dark mode to Transaction view

eviltofu:
Tap the activity tab, tap an entry listed, tap the icon in the upper right corner.
0 | 1 | 2
-|-|-
0 copy | 1 copy | 2 copy
.png)

Before commit | After commit
-|-
before | after


hboon:
You can refer to that screen as "Transaction view"

eviltofu:
Updated @hboon

hboon (approved):

PR #5641, opened 2022-10-21
Added dark mode to export keystore json

eviltofu:
Tap `Settings > Advanced > Export Keystore JSON`

Before | After
-|-
before-1|after-1
before-2|after-2
before-3|after-3

### DIFF
--- a/AlphaWallet/Common/Types/Configuration.swift
+++ b/AlphaWallet/Common/Types/Configuration.swift
@@ -233,7 +233,7 @@ struct Configuration {
             }
 
             static let textFieldIcon = UIColor { trait in
-                return colorFrom(trait: trait, lightColor: R.color.alabaster()!, darkColor: R.color.venus()!)
+                return colorFrom(trait: trait, lightColor: R.color.venus()!, darkColor: R.color.alabaster()!)
             }
 
             static let textFieldBackground = UIColor { trait in
--- a/AlphaWallet/Common/Views/ButtonsBar/ButtonsBarBackgroundView.swift
+++ b/AlphaWallet/Common/Views/ButtonsBar/ButtonsBarBackgroundView.swift
@@ -12,7 +12,7 @@ class ButtonsBarBackgroundView: UIView {
     private let buttonsBar: ButtonsBarViewType
     private let separatorLine: UIView = {
         let view = UIView()
-        view.backgroundColor = R.color.mike()
+        view.backgroundColor = Configuration.Color.Semantic.tableViewSeparator
         view.translatesAutoresizingMaskIntoConstraints = false
         return view
     }()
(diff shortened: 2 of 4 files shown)


AlphaWallet/Common/Views/ButtonsBar/ButtonsBarBackgroundView.swift
```diff
@@ -12,7 +12,7 @@ class ButtonsBarBackgroundView: UIView {
     private let buttonsBar: ButtonsBarViewType
     private let separatorLine: UIView = {
         let view = UIView()
-        view.backgroundColor = R.color.mike()
+        view.backgroundColor = Configuration.Color.Semantic.tableViewSeparator
```
hboon, line 15: This changes the color for Light mode right? Is it correct?

AlphaWallet/Common/Types/Configuration.swift
```diff
@@ -233,7 +233,7 @@ struct Configuration {
             }
 
             static let textFieldIcon = UIColor { trait in
-                return colorFrom(trait: trait, lightColor: R.color.alabaster()!, darkColor: R.color.venus()!)
+                return colorFrom(trait: trait, lightColor: R.color.venus()!, darkColor: R.color.alabaster()!)
```
hboon, line 236: Why switch it around?

eviltofu: It was a mistake in the first place. If not switched around, it would be the same colour as the text field background.

AlphaWallet/Common/Views/ButtonsBar/ButtonsBarBackgroundView.swift
eviltofu: Not for light mode but I made it so that all separators in the app have the same colour.

hboon: But:

```
static let tableViewSeparator = UIColor { trait in
    return colorFrom(trait: trait, lightColor: R.color.mercury()!, darkColor: R.color.venus()!)
}
```

Doesn't it mean that in light mode, it used to be `mike()` and is now `mecury()`?

AlphaWallet/Common/Types/Configuration.swift
hboon: Not sure how this is used. Is this relevant to this PR? or should it be a separate one? If it's a separate one, can you make a before and after screenshot to show what this does?

eviltofu: Before switching | After switching
-|-
before|after

hboon:
@oa-s would you please help to review? Thanks.

oa-s (approved):

oa-s:
@eviltofu fix conflicts and merge when ready

eviltofu:
Is this PR merge necessary now? I just ran the current master 014a8fcd9e84c6f8b3532264558479517e09c2a9 and this screen now behaves under dark mode.

PR #5697, opened 2022-10-28
simplify TokenOrContract #5624

### LINKED ISSUE #5624: Improve `TokenOrContract` in `TokensDataStore.swift`
This:

```
public enum TokenOrContract {
    case nonFungibleToken(ERCToken)
    case token(Token)
    case delegateContracts([AddressAndRPCServer])
    case deletedContracts([AddressAndRPCServer])
    /// We re-use the existing balance value to avoid the `Wallets` tab showing that token (if it already exist) as `balance = 0` momentarily
    case fungibleTokenComplete(name: String, symbol: String, decimals: Int, contract: AlphaWallet.Address, server: RPCServer, onlyIfThereIsABalance: Bool)
    case none
}

//…
```

A few things:

* `TokenOrContract` seems to be just used for importing/autodetecting, so maybe it can be moved to another file or its own file. The name might need some updating to reflect this
* The `.token` case might not belong here, seems to be overlapped with `nonFungibleToken` and `fungibleTokenComplete`
* `fungibleTokenComplete` might need to be renamed to remove the "Complete" or alternatively rename all the other cases, but that would be redundant
* The associated values for `nonFungibleToken` and `fungibleTokenComplete` is inconsistent — one uses an object to hold the values, the other, the raw values

oa-s:
Closes #5624

### DIFF
--- a/AlphaWallet/Tokens/Coordinators/NewTokenCoordinator.swift
+++ b/AlphaWallet/Tokens/Coordinators/NewTokenCoordinator.swift
@@ -96,7 +96,7 @@ extension NewTokenCoordinator: NewTokenViewControllerDelegate {
         delegate?.didClose(in: self)
     }
 
-    func didAddToken(token: ERCToken, in viewController: NewTokenViewController) {
+    func didAddToken(token: ErcToken, in viewController: NewTokenViewController) {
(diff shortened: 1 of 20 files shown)


hboon (approved):
Didn't test.

PR #5728, opened 2022-11-02
Update transaction screen with being able to self reload when transaction has changed #5726

### LINKED ISSUE #5726: Update transaction screen with being able to self resload when transaction has changed

### LINKED ISSUE #5774: Transparent navigation bar in dark mode for transaction view
Screenshot 2022-11-09 at 15 33 08

oa-s:
Closes #5726, Closes #5774


hboon (approved):

AlphaWallet/Transactions/ViewModels/TransactionDetailsViewModel.swift
```diff
@@ -4,164 +4,190 @@ import BigInt
...
-    }
+        self.analytics = analytics
+    }
+
+    func transform(input: TransactionDetailsViewModelInput) -> TransactionDetailsViewModelOutput {
+        let transactionRow = transactionsService.transactionPublisher(for: transactionRow.id, server: transactionRow.server)
+            .compactMap { $0 }
+            .receive(on: RunLoop.main)
+            .map { transaction -> TransactionRow in
+                if transaction.localizedOperations.count > 1 {
```
oa-s, line 55: @hboon do you think this can brake a screen as transaction row can have `.item(transaction, operation)` type.  so out of here we can't figure out right transaction row for transaction. TransactionRow builds based on localized operations, and thay can be changed when transaction has updated

hboon: Not exactly sure of the scenario, but a transaction should only change state when it's pending -> confirmed, so they should be at most a handful of pending transactions and they should all be at the top, sorted by nonce. Does that help? If not, can you describe more?

oa-s: so in best case transaction row will stay the same, that willn't cause issues

PR #5735, opened 2022-11-04
Updated Makefile to allow user to specify iOS 14 and 15 subversions, and iPhone device on simulator, and clean the source code by version

eviltofu:
### Selection of device/os combo in tests

Allows user to specify the iPhone device and iOS subversions for iOS 14 and iOS 15. The default options are iPhone 12 Pro, iOS 14.5, iOS 15.2, and the latest iOS 16 (16.1). 

If you want to use a different iPhone device, use:
`iphone_target="iPhone X" make test`

If you want to use a different iPhone device and iOS subversion, use:
`iphone_target="iPhone 13" test14_target=14.2 make test14`

In order to make these changes permanent, you would need to export the variables (probably via your shell login script).
`export iphone_target="iPhone 12"`
`export test15_target=15.7`

If the device, os combination does not exist in your Xcode setup, `xcodebuild` will dump a list of installed device iOS combinations. You would need to download the specific device/os combo in Xcode `Window > Devices and Simulators`.

### New clean options
The old `make clean` only cleaned up the source code for iPhone 12 and the latest iOS. The newer option cleans up all the source code from iOS 14, 15, and 16.

`make source_clean, make source_clean14, make source_clean15, make source_clean_all` allows the user to just clean the source code without removing the gem and pod files.

`make remove_installed_dir` removes the pod and gem files without removing any of the compiled code.

### Information display
The selected iPhone device, iOS 14 and iOS 15 subversions are displayed when the user types `make`.

### Package resolution
`make build` has been added in order to fix the problem of fixing the package resolution issue when the SwiftPackage Manager packages have been updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,22 +6,32 @@ vendor_path = ./vendor/bundle
 beautify_cmd = ./Pods/xcbeautify/xcbeautify
 curl_cmd = /usr/bin/curl
 compress_cmd = ./scripts/compress_chains
+iphone_destination ?= iPhone 12 Pro
+test14_target ?= 14.5
+test15_target ?= 15.2
 
 all: target
-	@echo "Please specify a target. Please use 'make target' to show targets."
+	@echo
+	@echo "Please specify a target."
+	@echo "iPhone destination is set to $(iphone_destination)."
 
 target:
-	@echo "install_bundle     : install the correct version of bundle."
-	@echo "install_gems       : install all the required gems."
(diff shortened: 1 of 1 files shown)


hboon:
@eviltofu let's just close this. Let's just strive for supporting the latest iOS/SDK version (after it's live for a while) and secondary support for 1 version older. But not really need to complicate the dev tool/process until we need more.

PR #5740, opened 2022-11-07
Wrap into struct representation encodeABI() smart contract functions calls. #5739

### LINKED ISSUE #5739: Wrap into struct representation encodeABI() smart contract functions calls.
make single responsibility structs like:
```
public struct Erc875Transfer: ContractMethod {
    let contractAddress: AlphaWallet.Address
    let recipient: AlphaWallet.Address
    let indices: [UInt16]

    public init(contractAddress: AlphaWallet.Address, recipient: AlphaWallet.Address, indices: [UInt16]) {
        self.contractAddress = contractAddress
        self.recipient = recipient
        self.indices = indices
    }

    public func encodedABI() throws -> Data {
        let parameters: [Any] = [recipient, indices.map({ BigUInt($0) })]
        let arrayType: ABIType
        if contractAddress.isLegacy875Contract {
            arrayType = ABIType.uint(bits: 16)
        } else {
            arrayType = ABIType.uint(bits: 256)
        }
        let functionEncoder = Function(name: "transfer", parameters: [.address, .dynamicArray(arrayType)])
        let encoder = ABIEncoder()
        try encoder.encode(function: functionEncoder, arguments: parameters)

        return encoder.data
    }
}
```

oa-s:
Closes #5739


hboon (approved):
Nice! Maybe break up `ContractMethod.swift` into multiple smaller files? And move them into a pod?

Do you intend to keep `web3swift` to more or less matching upstream? So we can either move this into `web3swift` (if we want to heavily customize web3swift), or we can create a new pod for web3-related stuff?

PS: Typo in "reprosentation"

PR #5771, opened 2022-11-09
Fix crash on app launch Thread 1: Simultaneous accesses to 0x11a4e5158, but modification requires exclusive access #5770

### LINKED ISSUE #5770: Fix crash on app launch `Thread 1: Simultaneous accesses to 0x11a4e5158, but modification requires exclusive access`

oa-s:
Fixes #5770


hboon (approved):
Why does it fix the crash?

oa-s:
tried few times, works for me

hboon:
Hmm.. value vs. reference semantics seems to be really hard to get right sometimes. I remember there was a bug where behavior just silently breaks when we swap `class` and `struct`…

PR #5780, opened 2022-11-10
[Refactor] Move credentials/APIs to single file

hboon:
@oa-s I thought there was a reason to keep them separate, but I can't remember now. Do you?


oa-s:
Have no idea, I know we apply it for some extension for rpc server. To move to separate file. To avoid crash. But for creds

oa-s (approved):

PR #5790, opened 2022-11-10
send analytics event `Passcode / Face ID enabled/disabled` #5755

### LINKED ISSUE #5755: Log to see how commonly Passcode / Face ID is enabled
~~Create Android issue (standardize name)~~ (forgot Android doesn't have this functionality)

oa-s:
Closes #5755

### DIFF
--- a/AlphaWallet/Settings/ViewModels/SettingsViewModel.swift
+++ b/AlphaWallet/Settings/ViewModels/SettingsViewModel.swift
@@ -118,7 +118,8 @@ final class SettingsViewModel {
                 }
             case .help, .tokenStandard, .version, .wallet: return nil
             }
-        }.compactMap { [lock] isOn -> Void? in
+        }.compactMap { [lock, analytics] isOn -> Void? in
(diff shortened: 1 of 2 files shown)


AlphaWallet/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -120,7 +120,11 @@ final class SettingsViewModel {
                 }
             case .help, .tokenStandard, .version, .wallet: return nil
             }
-        }.compactMap { [lock] isOn -> Void? in
+        }.compactMap { [lock, analytics] isOn -> Void? in
+            analytics.log(action: Analytics.Action.enableDisablePasscode, properties: [
```
hboon, line 124: It should be logged as a user property with `setUser(property:, value:)` instead

hboon: Probably add to a new enum-case to `UserProperties`

hboon (approved):

PR #5793, opened 2022-11-10
Set `next` bool bar button for `inputAccessoryButtonType: InputAccess

### LINKED ISSUE #5784: Tapping Done button in Custom gas ("Set Speed") screen does nothing so we can't set custom gas for transactions

oa-s:
Fixes #5784

### DIFF
--- a/AlphaWallet/Common/Views/TextField.swift
+++ b/AlphaWallet/Common/Views/TextField.swift
@@ -131,7 +131,7 @@ class TextField: UIControl {
             case .done:
                 textField.inputAccessoryView = UIToolbar.doneToolbarButton(#selector(doneButtonTapped), self)
             case .next:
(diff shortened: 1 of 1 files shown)


oa-s:
looks like we have to check `Done` `Next` buttons, not always work properly, we could use `IQKeyboardManager` really helps with keyboard.

hboon:
Let's try `IQKeyboardManagerSwift` then

oa-s:
@hboon while adding `IQKeyboardManager` i found that it can't resolve next response (when next button selected) for complex view hierarchy. This resoles by setting `self.view` for `UIViewController`.So we have to inherit out view controllers from newvelly added one `ViewController`. is that acceptable, or we need look for an another solution? its related to screens with multiple inputs.

hboon:
Hmm.. How about if we do it (focus next and scroll) manually? If we need it for more screens, we can find a more general approach? It looks like this might be the only screen in the app that is tall with textboxes

PR #5798, opened 2022-11-11
fix switching chains in walletconnect doesnt appear to work, tapping a chain does nothing  5797

hboon:


hboon:
Skip review to get it into release

PR #5805, opened 2022-11-14
WalletConnect switching UI #5799

### LINKED ISSUE #5799: Check (and if necessary fix) WalletConnect switch sessions's view controller(s) popping
Follow up to quick fix in #5798

https://github.com/AlphaWallet/alpha-wallet-ios/blob/c62c61b54d67b34121f28605a6e8212f52310615/AlphaWallet/WalletConnect/Coordinator/WalletConnectSessionCoordinator.swift#L63

Need to take care that it works from at least 3 points:

* Wallet tab when there's only 1 session
* Wallet tab when there's more than 1 session
* Settings > WalletConnect

oa-s:
Closes #5799


hboon (approved):
WalletConnect switching UI: So much better. Thanks!

PR #5807, opened 2022-11-15
Error when Send small amounts of value using USD View on PHI Network #5803

### LINKED ISSUE #5803: Error when Send small amounts of value using USD View on PHI Network
When in send section and change to USD from PHI and enter $100 and hit send it will send 0 phi because the amount is .00004 and that populates as zero can we please fix this so amounts smaller than $1000 populates when $1000 is used it works that is .0004 phi. But amounts like $50 will send a transaction as 0 phi when it's really .00002 PHI image

This is the transaction that was sent using $100 usd you can see how it actually sent 0 phi rather than .00004 phi 

https://phiscan.com/tx/0x0bd4ae807850c67bc72c9695d1effe82fff46430c4dd75f632e81ba249cacb1b

oa-s:
Closes #5803


oa-s:
was able to reproduce this issue when user's locale had decimalSeparator different from `EN_GB` locale that is using by default for number formatter.

hboon:
@oa-s can you give an example of USD to enter for PHI and ETH to try this out?

Maybe the amount I Used for ETH is too small:

Screenshot 2022-11-16 at 12 20 11 PM

hboon:
And would be good to have a test case for this (the parser, or formatter)

Phinetwork:
> @oa-s can you give an example of USD to enter for PHI and ETH to try this out?
> 
> 
> 
> Maybe the amount I Used for ETH is too small:
> 
> 
> 
> Screenshot 2022-11-16 at 12 20 11 PM
> 
> 

Anything below $1000 USD in PHI does not show. It seems to be 3 digit USD amounts on PHI network that show as zero amount in PHI when enter as as USD the conversion value is not using enough decimals.

Phinetwork:
image

Phinetwork:
> image

Sends like this on chain 

https://phiscan.com/tx/0x0bd4ae807850c67bc72c9695d1effe82fff46430c4dd75f632e81ba249cacb1b

Zero PHI was sent but value was $100 usd on wallet.

Phinetwork:
Also since new update phi20 tokens are being sent to contract address rather than destination wallet this needs to be fixed as it causes major issues in sending tokens they get lost at the contract address.

hboon (changes requested):
@oa-s with 18e9f8c8bc5b48df987433cd07db7aad3847f80d, I can still reproduce on Ethereum mainnet like this:

1. Enter ETH = 0.00001
2. Tap ETH to switch to USD. It shows $0.01
3. Tap Send

Expected:

4. Actionsheet shows 0.00001 ETH = 0.01 USD

Observed:

4. Actionsheet shows 0 ETH = 0 USD

Screenshot 2022-11-17 at 11 10 13 AM



This might be a quick way to test and hunt down the issue:

1. Enter USD $0.01
2. Tap USD to switch to ETH
3. Should non-0 ETH unless it goes beyond 18 decimals

oa-s:
yes, i facing the same, the amount is too small and as we use formating to maps to 0

oa-s:
@hboon what strategy could we use in such case?

hboon:
Does this make sense:

We always store the actual values (or the converted to values) without rounding them down to 4 (or 5) decimals. We only round them when displaying in most parts of the app, but never in the confirmation actionsheet

If the user converts from $X -> Y ETH, we can round it when we display in the textbox, so sometimes it will be 0, but this is not the issue in my last screenshot right?

If this isn't right, could help help give a specific (made up is fine) example of the amounts involved at each step and maybe we can see where it doesn't look right? But basically we should store absolute values and then at each place display either the absolute value or the rounded down value.

hboon:
Superseded by #5827

PR #5815, opened 2022-11-17
Remove PHI and PHI2 from built-in chains after internal review

hboon:


oa-s (approved):

Phinetwork:
Why would you remove this? After weve promoted your wallet? This is the worse mistake you could ever make. I thought this was an open-source decentralized wallet. Y'all are the same as these centralized scammy apps. You don't care about crypto you care about centralized Vc back pieces of trash that steal from people & have zero value. You just removed the most high performance chain on your app. This is by far the silliest decision.

Phinetwork:
> Why would you remove this? After weve promoted your wallet? This is the worse mistake you could ever make. I thought this was an open-source decentralized wallet. Y'all are the same as these centralized scammy apps. You don't care about crypto you care about centralized Vc back pieces of trash that steal from people & have zero value. You just removed the most high performance chain on your app. This is by far the silliest decision. 

You have nothing to say for yourself @hboon Victor Zhang is going to destroy this wallet. He's a moron.

PR #5826, opened 2022-11-18
[Tests] Fix broken tests

hboon:
@oa-s `KeystoreKey(password: passphrase, key: privateKey)` seems to have become slower? So strange.


oa-s:
Yes with latest version of iOS it became slower

PR #5827, opened 2022-11-18
Fix sending small ether amount #5803

### LINKED ISSUE #5762: (Fractional?) Amount in confirmation actionsheet doesn't match ETH transfer amount
Try sending some ETH to another wallet. Tested on iPhone 14 Pro hardware.

IMG_53B181B961F7-1

### LINKED ISSUE #5765: Replace String with BigInt when sending nativeCryptocurrency or erc20

### LINKED ISSUE #5803: Error when Send small amounts of value using USD View on PHI Network

### LINKED ISSUE #5910: SendViewModel.swift has white space violation
[SendViewModel.swift:528:1 ](https://github.com/AlphaWallet/alpha-wallet-ios/blob/9c0ede48ed8dff9e7a618769c10e3584d07cd9eb/AlphaWallet/Transfer/ViewModels/SendViewModel.swift#L528) Vertical Whitespace Violation: Limit vertical whitespace to a single empty line. Currently 2. (vertical_whitespace)

oa-s:
Closes  #5803, Closes #5765, Closes #5762, Closes #5910

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -786,6 +786,7 @@
 		87ED8FAB2488E4610005C69B /* SendViewSectionHeaderViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87ED8FAA2488E4610005C69B /* SendViewSectionHeaderViewModel.swift */; };
 		87ED8FAD2488EA610005C69B /* SupportViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87ED8FAC2488EA610005C69B /* SupportViewController.swift */; };
 		87ED8FAF2488EA9C0005C69B /* SupportViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87ED8FAE2488EA9C0005C69B /* SupportViewModel.swift */; };
+		87EFF9CC2941D4F500D3FE30 /* TransactionTypeFromQrCodeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87EFF9CB2941D4F500D3FE30 /* TransactionTypeFromQrCodeTests.swift */; };
 		87F15AEB27F1EA4E00EB9787 /* SwapQuoteDetailsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87F15AEA27F1EA4E00EB9787 /* SwapQuoteDetailsView.swift */; };
 		87F15AED27F1EA6900EB9787 /* SwapQuoteDetailsViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87F15AEC27F1EA6900EB9787 /* SwapQuoteDetailsViewModel.swift */; };
 		87F15AEF27F1EA8B00EB9787 /* SwapQuoteFieldView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 87F15AEE27F1EA8B00EB9787 /* SwapQuoteFieldView.swift */; };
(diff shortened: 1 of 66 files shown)


hboon (approved):

hboon:
@oa-s does this supersede #5807 ?

oa-s:
yes

oa-s:
@hboon  can u please review this pr?

hboon:
1. Tap Send
2. Tap ETH to switch to USD
3. Enter 0.2
4. Shows "~0.0001 ETH"
5. Tap USD to switch to ETH
6. Shows 0.0001 (great)
7. Edit to 0.00011

Expect
---
8. Show more than 0.2 USD (since 0.00011 > 0.0001)

Observed
---
8. Shows "~0.12 USD"

But if our computation is correct and just due to rounding, I think we can let this one through for now. Just wanted to make sure we aren't computing wrongly

hboon:
Also I needed to change: `return viewModel.fiatRawValue.value` to: `return viewModel.fiatRawValue.valueOptional` in order to build successfully.

hboon:
Sorry, hang on, looking at the code a bit

hboon:
@oa-s do you think if it helps to extract from `AmountTextFieldViewModel` the parts that due with the pair (Crypto <-> fiat), track `.notSet` (as well as the current exchange rate) etc into a standalone type? Then it's easier to understand both that new type and `AmountTextFieldViewModel,`and we can test that specifically with actual numbers. Preferably a dumb one, no publishers?

oa-s:
When we enter small amount of usd (`0.002` for example), we able to see rounded value of eth, `0.00001` for example, but actual its calculated value equals `0.00001958474636363748472` and so on. But when we switch back to crypto we expected to see `0.00001` and not `0.00001958474636363748472`, again when switching we do calculation of alternative value, usd in our case, and take as raw entered value `0.00001` multiplying it by some rate we will get another different from  `0.002` value. Any way we keep actual values, but displaying rounded. that is how we get zeros on confirmation when sending small usd values.

oa-s:
also i am facing an issue when transaction type get updated, (balance updated or something else), entered amount get changed. Fixing it now with moving updating amoun text field only when it has initial value, or all funds selected or qr code get scanned.

oa-s:
> @oa-s do you think if it helps to extract from `AmountTextFieldViewModel` the parts that due with the pair (Crypto <-> fiat), track `.notSet` (as well as the current exchange rate) etc into a standalone type? Then it's easier to understand both that new type and `AmountTextFieldViewModel,`and we can test that specifically with actual numbers. Preferably a dumb one, without `Combine`?

i will try to do that

oa-s:
@hboon  added `useOutputFormattingDebugFlag` to `AmountTextFieldViewModel` you can play with it.

we could show alternative amount for entered crypto/fiat value, and in case it different user will know that.

oa-s:
like here
Screenshot 2022-11-30 at 18 47 46

Screenshot 2022-11-30 at 18 48 21

hboon:
I can't tell `useOutputFormattingDebugFlag` what does. Can you help elaborate?

I noticed the green value (annotated) in the screenshot. But it appears regardless of the value of `useOutputFormattingDebugFlag`

Screenshot 2022-12-01 at 10 06 46 AM

Which one represents the actual value that is going to be used to craft the transaction?

oa-s:
when `useOutputFormattingDebugFlag` is true, that means we applying formatters for output numbers.

when crypo has selected we are goin to send value from text field
when fiat has selected we send green value, (calculated)

oa-s:
Updated pr, with increasing formaterrs fraction digits for amount text field only, still wip. but looks better, have to check more

oa-s:
looks like i have a solution:
- replace NSDecimalNumber for representing users entered amount with Double (because NSDecimalNumber failes check `((NSDecimalNumber(0.001) / NSDecimalNumber(0.02231)) * NSDecimalNumber(0.02231) == NSDecimalNumber(0.001))`)
- increase formatting fraction digits for amount only (confirmation screen, send screen) for fiat (**2** replace with **6**) crypto **4** replace with **8**

oa-s:
@hboon can you please check if it works on your side?

hboon:
> looks like i have a solution:
> 
> * replace NSDecimalNumber for representing users entered amount with Double (because NSDecimalNumber failes check `((NSDecimalNumber(0.001) / NSDecimalNumber(0.02231)) * NSDecimalNumber(0.02231) == NSDecimalNumber(0.001))`)
> * increase formatting fraction digits for amount only (confirmation screen, send screen) for fiat (**2** replace with **6**) crypto **4** replace with **8**

Which error does this fix? (sorry I might have forgotten). When would this computation be used? We don't usually have division unless it's for number of decimal places, right?

```
((NSDecimalNumber(0.001) / NSDecimalNumber(0.02231)) * NSDecimalNumber(0.02231) == NSDecimalNumber(0.001)))
```

oa-s:
issues fixed: 
- formating small amount (when send small fiat amount, and crypto to fiat rate is big)
- precission issue (when we send fiat selected value, i was facing an issue when entered amount like 0.001 became 0.009999 on confirmation screen). for some reasons because of NSDecimalNumber computation

hboon:
@oa-s hmm.. ok. Would you help to add a test case and let's get this merged.

PR #5847, opened 2022-11-22
Add missing copied to clipboard alerts on accounts screen #5846

### LINKED ISSUE #5846: Add missing copied to clipboard alerts on accounts screen

oa-s:
Closes #5846


AlphaWallet/Accounts/Coordinators/AccountsCoordinator.swift
```diff
@@ -182,18 +181,23 @@ class AccountsCoordinator: Coordinator {
...
             controller.addAction(copyAction)
+
             let cancelAction = UIAlertAction(title: R.string.localizable.cancel(), style: .cancel) { _ in }
             controller.addAction(cancelAction)
 
             navigationController.present(controller, animated: true)
         }
     }
 
+    private func showCopiedToClipboard(account: Wallet) {
```
hboon, line 196: Maybe `copyToClipboard(account:)` instead? Because it doesn't only show the alert, it does the copying too

hboon (approved):

PR #5851, opened 2022-11-22
Replace XmlHandler with struct definition #5840

### LINKED ISSUE #5840: Memory leak, replace XmlHandler with struct definition
leaving the app for some time on simulator causes memory usage grows over gigabytes

oa-s:
Closes  #5840


modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -759,7 +759,7 @@ final class ThreadSafe {
 }
 
 /// This class delegates all the functionality to a singleton of the actual XML parser. 1 for each contract. So we just parse the XML file 1 time only for each contract
-public class XMLHandler {
+public struct XMLHandler {
```
hboon, line 762: Why do we do this?

Also it should be harmless because of `XMLHandler`'s properties are reference types, right?

PR #5853, opened 2022-11-22
Replace BigInt with BigUInt #5820

### LINKED ISSUE #5820: Replace BigInt with BigUInt
most of values in the app can't be negative

oa-s:
Closes #5820


hboon (approved):
Didn't check the code, but run through smoke test, seems OK.

hboon:
BTW, good to explain the why in the commit, eg.:

> Replace BigInt with BigUInt when they are never negative #5820

PR #5855, opened 2022-11-23
Add switching between fiat currencies feature #5758

### LINKED ISSUE #5758: Add switching between fiat currencies feature

oa-s:
Closes #5758

### DIFF
--- a/AlphaWallet/Accounts/ViewModels/AccountViewModel.swift
+++ b/AlphaWallet/Accounts/ViewModels/AccountViewModel.swift
@@ -15,7 +15,7 @@ struct AccountViewModel {
             let style = NSMutableParagraphStyle()
             style.alignment = .right
 
-            return .init(string: accountRowViewModel.balance.valuePercentageChangeValue, attributes: [
+            return .init(string: accountRowViewModel.balance.changePercentageString, attributes: [
                 .font: Fonts.regular(size: 20),
                 .foregroundColor: accountRowViewModel.balance.valuePercentageChangeColor,
                 .paragraphStyle: style
@@ -66,7 +66,7 @@ extension BlockiesImage {
 
 extension WalletBalance {
     var valuePercentageChangeColor: UIColor {
-        return EthCurrencyHelper(ticker: nil).valueChangeValueColor(from: changeDouble)
+        return TickerHelper(ticker: nil).valueChangeValueColor(from: change?.amount)
     }
 }
 
--- a/AlphaWallet/Accounts/ViewModels/AccountsViewModel.swift
+++ b/AlphaWallet/Accounts/ViewModels/AccountsViewModel.swift
@@ -101,7 +101,8 @@ final class AccountsViewModel {
             .map { $0.map { self.buildAccountRowViewModel(wallet: $0) } }
             .flatMapLatest { $0.combineLatest() }
 
-        let walletsSummary = walletBalanceService.walletsSummary
+        let walletsSummary = input.willAppear
+            .flatMap { [walletBalanceService] _ in walletBalanceService.walletsSummary }
             .map { [config] in WalletSummaryViewModel(walletSummary: $0, config: config) }
 
         let viewModels = Publishers.CombineLatest(accountRowViewModels, walletsSummary)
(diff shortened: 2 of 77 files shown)


hboon (commented):
The charts don't appear but not sure if it's just this PR.

I didn't check, but is the currency default to USD for users who have not enabled the feature flag? Especially new installs

hboon (approved):
The charts don't appear but not sure if it's just this PR.

I didn't check, but is the currency default to USD for users who have not enabled the feature flag? Especially new installs

The send screen should probably be updated too? But we can leave it for a separate PR:

Screenshot 2022-11-30 at 11 01 52 AM

oa-s:
> I didn't check, but is the currency default to USD for users who have not enabled the feature flag? Especially new installs

i will double check but yes, default currency is USD

oa-s:
> The charts don't appear but not sure if it's just this PR.
> 
> I didn't check, but is the currency default to USD for users who have not enabled the feature flag? Especially new installs
> 
> The send screen should probably be updated too? But we can leave it for a separate PR:
> 
> Screenshot 2022-11-30 at 11 01 52 AM

oh yes, it didn't applied for amount text field

hboon:
Just in case it's ready, buzz me :)

oa-s:
> Just in case it's ready, buzz me :)

@hboon looks ok now

hboon:
Got a very strange crash at https://github.com/oa-s/alpha-wallet-ios/blob/52081f6e986dbd33c5c76d66cc42907585d95ffb/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/TokenScriptFileIndices.swift#L68

```
if contractsToFileNames[contract].isEmpty {
```

It crashed complaining that it's nil. It looks like the follow extension (which is in the same module) isn't loaded?

```
extension Optional where Wrapped: Collection {
    public var isEmpty: Bool {
        switch self {
        case .none:
            return true
        case .some(let value):
            return value.isEmpty
        }
    }
}
```

After I added some debug statements, I can't reproduce it anymore, even after removing the debug statements. I don't think it's related to this PR.

hboon (commented):
Remember to reset this:

```
public let shouldPretendIsRealWallet = true
```

Maybe not related to this PR? But "All Funds" show 3.0958, but the actionsheet total is 3.0956 ETH.

Screenshot 2022-12-02 at 3 20 57 PM

oa-s:
> Maybe not related to this PR? But "All Funds" show 3.0958, but the actionsheet total is 3.0956 ETH.

yes, its value - gas

oa-s:
> Got a very strange crash at https://github.com/oa-s/alpha-wallet-ios/blob/52081f6e986dbd33c5c76d66cc42907585d95ffb/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/TokenScriptFileIndices.swift#L68
> 
> ```
> if contractsToFileNames[contract].isEmpty {
> ```
> 
> It crashed complaining that it's nil. It looks like the follow extension (which is in the same module) isn't loaded?
> 
> ```
> extension Optional where Wrapped: Collection {
>     public var isEmpty: Bool {
>         switch self {
>         case .none:
>             return true
>         case .some(let value):
>             return value.isEmpty
>         }
>     }
> }
> ```
> 
> After I added some debug statements, I can't reproduce it anymore, even after removing the debug statements. I don't think it's related to this PR.

hmm, haven't seen that

oa-s:
maybe need to create a new issue for that, definatelly not related to currencies changing

hboon:
> > Maybe not related to this PR? But "All Funds" show 3.0958, but the actionsheet total is 3.0956 ETH.
> 
> yes, its value - gas

But shouldn't it show `3.0956` above "All Funds" then? Maybe `3.0958` is always how it worked? Never mind that.

hboon:
> maybe need to create a new issue for that, definatelly not related to currencies changing

Created #5899.

hboon:
@oa-s this one's not ready yet?

oa-s:
looks like its ready

oa-s:
only thing i can facing its when we got a lot of tokens it takes to long time to resolve tickers, especially for ticker ids resolving. Might need to check resolvation logic, in another pr.

hboon:
Maybe related to:

> the last isssue to fix is left (sometimes tokens don't get updated when a new tickers resolved)

After I switch currency and the UI reloads, showing the Wallet tab, I wait for ~1 minute and the fiat data isn't showing up (it was shown before I switched currency):

Screenshot 2022-12-15 at 9 49 01 AM

Pulling down to refresh shows this after ~2 secs:

Screenshot 2022-12-15 at 9 49 28 AM

hboon:
If you prefer to merge first, go ahead since there's a feature flag.

PR #5857, opened 2022-11-23
Add switch fiat currencies screen #5758

### LINKED ISSUE #5758: Add switching between fiat currencies feature

oa-s:
Closes #5758


AlphaWallet/Settings/ChangeCurrency/Coordinator/ChangeCurrencyCoordinator.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  ChangeCuurencyCoordinator.swift
+//  AlphaWallet
+//
+//  Created by Vladyslav Shepitko on 30.06.2020.
```
hboon, line 5: The header dates are wrong

hboon (commented):
Can you explain how this PR and #5855 work together or what they do?

I ran #5857 and when I change the currency from USD to Japanese Yen, Wallet tab, send currency option as well as the graphs are still in USD.

Then I tried #5855, but it crashed with:

> AlphaWalletFoundation/RealmStore.swift:106: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=10 "Migration is required due to the following errors:

Maybe I ran them in the wrong order?

oa-s:
#5857 is still wip , have to fix migration

oa-s:
#5855 only adds currencies screen, will tweak a little bit it, and continue with #5857

hboon:
Ah, ok. Let me know what to check then :)

oa-s:
@hboon pr is pretty good for review

PR #5888, opened 2022-12-01
Fix stucking transaction as pending #5874

### LINKED ISSUE #5874: Transaction that was confirmed quickly still marked and stuck as pending

oa-s:
Fixes #5874

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Activities/Types/ActivitiesFilterStrategy.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Activities/Types/ActivitiesFilterStrategy.swift
@@ -50,11 +50,12 @@ fileprivate extension ActivitiesFilterStrategy {
 }
 
 extension ActivitiesFilterStrategy.functional {
-    static func predicateForNativeCryptocurrencyTransactions() -> NSPredicate {
+    static func predicateForNativeCryptocurrencyTransactions(allowZeroValue: Bool = true) -> NSPredicate {
         let completed = TransactionState.predicate(for: .completed)
         let pending = TransactionState.predicate(for: .pending)
         let isInCompletedOrPandingState = NSCompoundPredicate(orPredicateWithSubpredicates: [completed, pending])
-        let valueNonEmpty = NSPredicate(format: "value != '' AND value != '0'")
+        //NOTE: Why actually we need value to be non zero?
+        let valueNonEmpty = NSPredicate(format: allowZeroValue ? "value != ''" : "value != '' AND value != '0'")
         let hasZeroOperations = NSPredicate(format: "localizedOperations.@count == 0")
 
         return NSCompoundPredicate(andPredicateWithSubpredicates: [isInCompletedOrPandingState, hasZeroOperations, valueNonEmpty])


AlphaWallet/Tokens/ViewModels/FungibleTokenDetailsViewModel.swift
```diff
@@ -46,7 +46,7 @@ final class FungibleTokenDetailsViewModel {
     func transform(input: FungibleTokenDetailsViewModelInput) -> FungibleTokenDetailsViewModelOutput {
         input.willAppear.flatMapLatest { [coinTickersFetcher, token] _ in
             coinTickersFetcher.fetchChartHistories(for: .init(token: token), force: false, periods: ChartHistoryPeriod.allCases)
-        }.print("xxx.chartHistories").assign(to: \.value, on: chartHistoriesSubject)
+        }.assign(to: \.value, on: chartHistoriesSubject)
```
hboon, line 49: Separate commit or PR? :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Activities/Types/ActivitiesFilterStrategy.swift
```diff
@@ -50,11 +50,12 @@ fileprivate extension ActivitiesFilterStrategy {
 }
 
 extension ActivitiesFilterStrategy.functional {
-    static func predicateForNativeCryptocurrencyTransactions() -> NSPredicate {
+    static func predicateForNativeCryptocurrencyTransactions(allowZeroValue: Bool = true) -> NSPredicate {
         let completed = TransactionState.predicate(for: .completed)
         let pending = TransactionState.predicate(for: .pending)
         let isInCompletedOrPandingState = NSCompoundPredicate(orPredicateWithSubpredicates: [completed, pending])
-        let valueNonEmpty = NSPredicate(format: "value != '' AND value != '0'")
+        //NOTE: Why actually we need value to be non zero?
```
hboon, line 57: Do you mean why a native crypto transaction can send 0 eth when it doesn't seem to do anything useful? I don't know the original reason why it's allowed, but it can be used for a few things:

* cancel a pending transaction for almost free (just gas). This is how we cancel pending transactions in AlphaWallet
* force the nonce to be updated. when we deploy a contract on EVMs, the contract's address is computed from the combination of the sender and the nonce. So this can be a useful trick to force the nonce to be the same on several chains, so you can have the same contract address on those chains (example, my USD-Boon token is always 0xb0b on Ethereum, Gnosis, etc or my smart contract wallet is always 0xc0c)
* to signal something on-chain for a certain protocol. eg. we both agree there's an address 0xb0b that we can send 0 ETH to from an address to say that we trust you. The 0xb0b might be computed.

If this answers your note in the code, would you help to update/remove it accordingly? :)

hboon: But I just tried on master a62a0baa500ebb6e94a872ba88c436936a18bcfd, sending 0 ETH on Goerli and it seems to update the Activity tab correctly. But maybe this time it picked it up with a full transaction history refresh?

AlphaWallet/Tokens/ViewModels/FungibleTokenDetailsViewModel.swift
oa-s: removed

modules/AlphaWalletFoundation/AlphaWalletFoundation/Activities/Types/ActivitiesFilterStrategy.swift
oa-s: for me it appeared in activities tab too, but not in activities for selected token, when send screen has poeed

hboon:
Close for now

PR #5889, opened 2022-12-01
Write a test that verifies send ERC20 token generates correct data, add functions for building unsigned transaction from transaction type #5818

### LINKED ISSUE #5818: Write a test that verifies send ERC20 token generates correct data
Best to check the `data` field of the transaction directly. Test should fail if #5812 is not applied

oa-s:
Closes #5818


hboon (approved):
Great, but need to fix the tests. It doesn't build

oa-s:
> Great, but need to fix the tests. It doesn't build

fixed

PR #5896, opened 2022-12-02
fix crash when accessing backing store

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/AssetDefinitionBackingStore.swift
```diff
@@ -2,7 +2,7 @@
 
 import Foundation
 
-public protocol AssetDefinitionBackingStore {
+public protocol AssetDefinitionBackingStore: AnyObject {
```
hboon, line 5: Huh. Why does this crash?

PR #5897, opened 2022-12-02
Support Sepolia #5480

### LINKED ISSUE #5480: Support Sepolia
The official Sepolia RPC listed here: https://sepolia.dev ie https://rpc.sepolia.dev

Isn't working, so use the .org variant instead, which does. It's an official but independent RPC (winner of https://twitter.com/TimBeiko/status/1536802812683689985):

https://rpc.sepolia.org/

Here is the fallback RPC:

https://rpc2.sepolia.org/

These are the only two reliable Sepolia RPCs I could find.

Also ensure the API endpoint for fetching TX is:

https://api-sepolia.etherscan.io/api?

oa-s:
Closes #5480


oa-s:
need to review api and rpc urls

hboon (commented):
Is this done?

oa-s:
Yes, it looks good for me

hboon (approved):
Didn't check. Trust you tested. Seems to require quite a few files to be modified though. I though the made it so that fewer files were required to support a new chain.

PR #5903, opened 2022-12-05
Change `fetchXML` result for official token script file to `.unmodified`, to make test pass #5900

### LINKED ISSUE #5900: testShouldNotCallCompletionBlockWithCacheCaseIfNotAlreadyCached fails
Test fails because `AssetDefinitionInMemoryBackingStore` always returns false for `isOfficial`. 

AssetDefinitionStoreTests.swift:22
```swift
    func testShouldNotCallCompletionBlockWithCacheCaseIfNotAlreadyCached() {
        let contractAddress = AlphaWallet.Address.make()
        let store = AssetDefinitionStore(backingStore: AssetDefinitionInMemoryBackingStore())
        let expectation = XCTestExpectation(description: "cached case should not be called")
        expectation.isInverted = true
        store.fetchXML(forContract: contractAddress, server: nil, useCacheAndFetch: true) { [weak self] result in
            guard self != nil else { return }
            switch result {
            case .cached:
                expectation.fulfill()
            case .updated, .unmodified, .error:
                break
            }
        }
        wait(for: [expectation], timeout: 0)
    }
```

AssetDefinitionStore.swift:211
```swift
        if !backingStore.isOfficial(contract: contract) {
            completionHandler?(.cached)
            return
        }
```

AssetDefinitionInMemoryBackingStore.swift:40
```swift
    public func isOfficial(contract: AlphaWallet.Address) -> Bool {
        return false
    }
```

oa-s:
Closes #5900


hboon (approved):

hboon:
@oa-s the commit mesage and PR title needs updating: it's "overridden", not "official token script file".

PR #5912, opened 2022-12-07
Fixed UIColorExtensionTests.swift

### LINKED ISSUE #5909: Test for color/UI fails
`UIColorExtensionTests.testLightDarkColorMode`

eviltofu:
Fixes #5909

### DIFF
--- a/AlphaWalletTests/Extensions/UIColorExtensionTests.swift
+++ b/AlphaWalletTests/Extensions/UIColorExtensionTests.swift
@@ -41,8 +41,8 @@ class UIColorExtensionTests: XCTestCase {
     }
 
     func testLightDarkColorMode() throws {
-        let lightColor = Colors.appRed
-        let darkColor = Colors.appWhite
+        let lightColor = UIColor(red: 50, green: 100, blue: 150)
+        let darkColor = UIColor(red: 10, green: 20, blue: 30)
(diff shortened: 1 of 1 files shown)


AlphaWalletTests/Extensions/UIColorExtensionTests.swift
```diff
@@ -41,8 +41,8 @@ class UIColorExtensionTests: XCTestCase {
     }
 
     func testLightDarkColorMode() throws {
-        let lightColor = Colors.appRed
-        let darkColor = Colors.appWhite
+        let lightColor = UIColor.red
```
hboon, line 44: Why does the original version not work? It seems like it should work with any pair of 2 different colors?

hboon: How about a different pair of arbitrary RGB colors that is not any of the semantic colors used in the app that passes? Eg 50, 100, 150 and another one? Helps show that it doesn’t rely on any style semantics

oa-s (approved):

AlphaWalletTests/Extensions/UIColorExtensionTests.swift
eviltofu: Ok I used 1,0,0,1 for red and 1,1,1,1 for white.

PR #5918, opened 2022-12-08
Fix: TokenScript attributes calling the same function but with different return types specified will return wrong results

### LINKED ISSUE #5917: TokenScript attributes calling the same function but with different return types specified will return wrong results
If we define 2 TokenScript attributes, mapping to the same function call but with different function type, eg.

```
attribute1: someFunction(), as return type uint256

attribute2: someFunction(), as return type bool
```

The first call may return `1` (as expected), but the second call, which will read from our cache if it's within the TTL, will not cast `1` to bool, instead it will be treated as `0` (as the cached value isn't `true`).

hboon:
Fixes #5917


oa-s (approved):

hboon:

PR #5929, opened 2022-12-12
Fix crash when enter value bigger than Int.max for amount text field #5928

### LINKED ISSUE #5928: Fix crash when enter value bigger than Int.max for amount text field

oa-s:
Fixes #5928


hboon (approved):
Would you help create a test for this one? Just test `string(double:,minimumFractionDigits:,maximumFractionDigits:)`

Because it looks like it's easy to regress.

PR #5930, opened 2022-12-12
Apply dark mode for erc875 sell paymant flow screens #5927

### LINKED ISSUE #5927: Apply dark mode for erc875 sell paymant flow screens

oa-s:
Closes #5927


hboon (commented):
Do you need the other commit to be included? Also there seems to be some formating changes. Can you see if it's possible to separate them into a separate "Format code" commit? Don't worry about if it takes too long.

PR #5932, opened 2022-12-12
Move Apollo callback to main queue to avoid bad access #5931

### LINKED ISSUE #5931: Fix bad access crash with Apollo

oa-s:
Fixes #5931


hboon (approved):
How do you reproduce this?

oa-s:
wasn't able to reproduce it, there is only possible reason of crash could be callback on background queue

PR #5954, opened 2022-12-15
Extract alamofire requests from import magic token coordinator ( use base network service part 2)

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/UniversalLink/ImportMagicLinkNetworkService.swift
```diff
@@ -0,0 +1,200 @@
...
+//
+//  Created by Vladyslav Shepitko on 15.12.2022.
+//
+
+import Foundation
+import BigInt
+import Combine
+import AlphaWalletCore
+
+public class ImportMagicLinkNetworkService {
```
hboon, line 13: Maybe not a "service". Just "ImportMagicLinkNetwork" ?

hboon (approved):
Didn't check everything. Seems ok :)

PR #5955, opened 2022-12-15
Extract alamofire requests from fetch xml file coordinator

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/AssetDefinitionStore.swift
```diff
@@ -389,6 +364,55 @@ extension AssetDefinitionStore {
...
+        }()
+
+        private static let lastModifiedDateFormatter: DateFormatter = {
+            let df = DateFormatter()
+            df.dateFormat = "E, dd MMM yyyy HH:mm:ss z"
+            df.timeZone = TimeZone(secondsFromGMT: 0)
+            return df
+        }()
+
+        private static func httpHeadersWithLastModifiedTimestamp(includeLastModifiedTimestampHeader: Bool, lastModifiedDate: Date?) -> HTTPHeaders {
```
hboon, line 402: Some of these can be `functional` if they don't modify global state?

hboon: But otherwise seems to work. Thanks!

hboon (approved):

PR #5965, opened 2022-12-19
Disable NFT transfer functionality with feature flag. Feature flag is default to false (i.e. disabled)

hboon:
This `Send` button is hidden with this PR (because of app store review):

Screenshot 2022-12-19 at 4 16 26 PM


oa-s (approved):

hboon:

PR #5967, opened 2022-12-19
add single network provider, add single provider for calling rpc call  #5966

### LINKED ISSUE #5966: Standardize rpc calls
What should this issue resolve:
- single network provider (making rpc and url requests), because now we are using alamofire, url session and whatever else)
- single provider for calling rpc call, (easy to handle request trying, switching beetwen rpc ulrs, (in case server supports multiple rpcs))
- single instance on provider passed across the app

oa-s:
Closes #5966


hboon (approved):
Too big to check. Let me know if there's something specific to look at? Otherwise I'll assume it's all good.

oa-s:
latest updates retry logic, for now we got 404 error code, skip retyring on it, and when error.isCancellable

oa-s:
@hboon can u please check latest commit `Restrict upload xml file (for status checking) only one time, extract alamofire from xml handler`, is it good?

hboon:
> @hboon can u please check latest commit `Restrict upload xml file (for status checking) only one time, extract alamofire from xml handler`, is it good?

Seems ok

hboon:
@oa-s any chance to merge this in even if we don't finish up all the work? Would that reduce the maintenance work keeping this PR alive?

oa-s:
> any chance to merge this in even if we don't finish up all the work? Would that reduce the maintenance work keeping this PR alive?

there need more testing, for now i am on other issues return to this once would have time

hboon:
Close for now

PR #5973, opened 2022-12-20
[SwiftFormat] Enable SwiftFormat, but with minimal set of rules that doesn't modify existing code

hboon:
Also include rules (but commented out) that should be enabled gradually.

Also fixes invocation of SwiftFormat in build phase so:

* It only runs for `Debug` builds
* It uses the CocoaPod instead of testing for a system wide `swiftformat` binary


hboon:

PR #5974, opened 2022-12-21
Upgrade to Swift 5.0 (some external pods are kept at 4.2)

hboon:
Compatible with enabling SwiftFormat in #5973


AlphaWallet/Common/Views/SvgImageView.swift
```diff
@@ -43,7 +43,7 @@ final class SvgImageView: WKWebView {
     }
 
     func setImage(url: URL, completion: @escaping () -> Void) {
-        if let data = try? ImageCache.default.diskStorage.value(forKey: url.absoluteString), let svgString = data.flatMap({ String(data: $0, encoding: .utf8) }) {
+        if let data = try? ImageCache.default.diskStorage.value(forKey: url.absoluteString), let svgString = String(data: data, encoding: .utf8) {
```
hboon, line 46: Could you verify this is correct this line is correct?:

I think in Swift 4.2, the `if let data` is interpreted as `Data?` (because it's a `try?` and `value(forKey:)` returns an optional, so it's actually `Data??`) but Swift 5 coalesces "double" optionals, so `data` becomes `Data` as it should, so we no longer need the `flatMap(_:)`

oa-s (approved):

AlphaWallet/Common/Views/SvgImageView.swift
oa-s: looks good

Issue #5693, opened 2022-10-28
Update WalletConnect Sign v2 SDK


hboon:

oa-s:
looks like this not working properly, not able to run example app from `WalletConnect Sign v2` stucks on pairing, as well as our updated impl (for 1.0.2 version)

PR #5978, opened 2022-12-21
Replace WalletConnect v1 and Firebase Crashlytics with pod version #5977

### LINKED ISSUE #5977: Replace walletConnect v1, v2 and firebase with pod version, easier to resolve it ( while switching between branches pods don't resets like spm packages do)

oa-s:
Closes #5977


oa-s:
after #5693

AlphaWallet/Common/Services/FirebaseReportService.swift
```diff
@@ -21,9 +21,9 @@ extension AlphaWallet {
             }
 
             let file = isRunningTests() ? R.file.googleServiceInfoTestsPlist() : R.file.googleServiceInfoPlist()
-            if let options = file.flatMap({ FirebaseOptions(contentsOfFile: $0.path) }) {
+            if let options = file.flatMap({ FIROptions(contentsOfFile: $0.path) }) {
```
hboon, line 24: Does the code difference means we are using an older or newer version of Crashlytics?

hboon:
BTW, in the commit message: "crachlytics" -> "crashlytics"

so we can find it when looking in history :)

AlphaWallet/Common/Services/FirebaseReportService.swift
oa-s: i am not quite sure, because we import same version, mabe it different for pods.

PR #5984, opened 2022-12-22
fix crashing when decoding swap quote response #5981

### LINKED ISSUE #5981: Native Swap crashes while waiting (before tapping Continue)
1. Wallet tab
2. Tap Ether
3. Tap …
4. Tap Swap
5. Tap Native Swap
6. Enter 1 ETH
7. Tap Select Token
8. Choose USDC
9. Wait
10. Crashes in this screen:

Screenshot 2022-12-22 at 12 20 24 PM

oa-s:
Closes #5981

### DIFF
--- a/AlphaWallet/Swap/Swap/ViewModels/SwapQuoteDetailsViewModel.swift
+++ b/AlphaWallet/Swap/Swap/ViewModels/SwapQuoteDetailsViewModel.swift
@@ -42,7 +42,27 @@ final class SwapQuoteDetailsViewModel {
                     let subSteps = step.estimate.gasCosts.map {
                         SwapSubStep(gasCost: $0, type: step.type, amount: pair.swapQuote.estimate.toAmount, token: pair.swapQuote.action.toToken, tool: step.tool)
                     }
-                    return SwapStep(tool: pair.swapQuote.type, subSteps: subSteps)
+
+                    if subSteps.isEmpty {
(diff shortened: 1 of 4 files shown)


oa-s:
For some reasons api returns empty `gasCosts` field. `GasCost` displays with `-` value.

hboon:
> For some reasons api returns empty `gasCosts` field. `GasCost` displays with `-` value.

What does `GasCost` mean? Is it literally the Ethereum gas fee for the transaction?

hboon (approved):
Works for me and swap successful https://etherscan.io/tx/0x112622693fae497b9636efbdd1473493ac4742ef43eeea8b4bfe84d4795556f0.

Do we know why it crashed before the PR? Was it because something has change on the API side; is it something we can prevent in the future?

oa-s:
> Do we know why it crashed before the PR? Was it because something has change on the API side; is it something we can prevent in the future? 

guess something has changed on api side, and they return json without `gasPrice` field, that why we get crash while decoding

oa-s:
make a new issue to avoid crash  when decode is failed

oa-s:
sorry missed that the crash caused  line, don't remember why it was needed. shouldn't crash anymore, 
`                guard let data = try? JSONSerialization.data(withJSONObject: data) else { throw SwapError.invalidJson }`

PR #5985, opened 2022-12-22
Fix fiat currency symbol duplication for formatted string #5980

### LINKED ISSUE #5980: Duplicate "USD USD" in confirmation actionsheet

oa-s:
Fixes #5980


hboon (approved):
Shows "0.33 USD" instead of "0.33 USD USD" now.

But seems like many lines of code and files touched. Was it necessary to fix it?

hboon:
Ah, I see, it's the oldest 2 commits which fixed it. All good.

oa-s:
@hboon can u please check it one more time, now should be ok

hboon (approved):

PR #5986, opened 2022-12-22
Take initial value from amount input, when transaction type has no initial amount #5859

### LINKED ISSUE #5859: When sending test ETH (Goerli), user not able to send 0 unless amount text field has been altered
Simulator Screen Recording - iPhone 14 Pro - 2022-11-25 at 13 17 11

oa-s:
Closes #5859

### DIFF
--- a/AlphaWallet/Transfer/ViewModels/SendViewModel.swift
+++ b/AlphaWallet/Transfer/ViewModels/SendViewModel.swift
@@ -85,10 +85,19 @@ final class SendViewModel: TransactionTypeSupportable {
     }
 
     func transform(input: SendViewModelInput) -> SendViewModelOutput {
+        var isInitialAmountToSend: Bool = true
+        //NOTE: override initial value if `transactionType.amount` in nil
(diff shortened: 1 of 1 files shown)


hboon (changes requested):
Before (entering an amount) |After PR (not entering an amount)|After PR (entering an amount)
-|-|-
Screenshot 2022-12-23 at 11 00 57 AM|Screenshot 2022-12-23 at 10 56 17 AM|Screenshot 2022-12-23 at 11 06 55 AM

After the PR, not entering an amount displays the balance in the actionsheet as "3. ETH" as well as gas used ("~0.ETH"). Similar if we entered "0 ETH" to send. But if we entered "1.2 ETH" it seems to work (3rd screenshot).

So seems something wrong when it's "0 ETH" — doesn't matter whether we enter it explicitly?

hboon:
@oa-s also tests are failing (not just the ones that Travis sometimes fails because of timeout or rate limit)

oa-s:
formatting is fixed in #5980, here is taking initial value for transaction type

hboon (approved):

PR #6011, opened 2022-12-28
Remove AtomicDictionary from AssetDefinitionStore to avoid bad access #6010

### LINKED ISSUE #6010: Crash with AtomicDictionary
Seems to occur when I switch wallets. Maybe related to TokenScript, but can't reproduce reliably. Stack track empty in Xcode crashes.

https://console.firebase.google.com/u/0/project/alphawallet-ios-prod/crashlytics/app/ios:com.stormbird.alphawallet/issues/9d0d1fe16f41d8b2a64431ac61d67b62?time=last-hour&sessionEventKey=3ec788fcc1bb48dfa6c2c1b9c5ebbd23_1760509057806674704

oa-s:
Fixes #6010

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/AssetDefinitionStore.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/AssetDefinitionStore.swift
@@ -36,7 +36,7 @@ public class AssetDefinitionStore: NSObject {
     private var bodyChangeSubject: PassthroughSubject<AlphaWallet.Address, Never> = .init()
     private var listOfBadTokenScriptFilesSubject: CurrentValueSubject<[TokenScriptFileIndices.FileName], Never> = .init([])
     private let network: AssetDefinitionNetworking
-    private var cancelable = AtomicDictionary<Int, AnyCancellable>()
+    private var cancelable: [Int: AnyCancellable] = [:]
 
     public var listOfBadTokenScriptFiles: AnyPublisher<[TokenScriptFileIndices.FileName], Never> {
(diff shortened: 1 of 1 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/AssetDefinitionStore.swift
```diff
@@ -222,8 +222,8 @@ public class AssetDefinitionStore: NSObject {
 
         cancelable[request.hashValue] = network
             .fetchXml(request: request)
-            .sink(receiveCompletion: { [cancelable] _ in
-                cancelable[request.hashValue] = .none
+            .sink(receiveCompletion: { [weak self] _ in
+                self?.cancelable[request.hashValue] = .none
```
hboon, line 226: Torn over situations like this. This means access to `cancelable` is on the thread that `.sink(receiveCompletion:)`runs on, which is dependent on what `.fetchXml(request:)` runs on (which is `.main` since there's a `receive(on:)` in the implementation) , unless we had a specific `receive(on:)` before it. This would probably run decently for us as we migrate most things to run on main, and then it'll just take 1 or a few calls that runs on non-main thread to cause bugs to appear again mysteriously.

hboon: I don't know of a fullproof way, but I think what we agreed earlier to do — always "return" to the main thread is the best way. Kind of like what `fetchXml(request:)` does, so if a function has to branch off the non-main, it must always "return" back on main.

Nothing to change here. Just sharing some thoughts.

hboon (approved):

PR #6029, opened 2022-12-30
remove unused account parameter for DappBrowserCoordinator

jqphu:
The DappBrowser passes in the account field but it is currently unused. This is because we rely on the unconfirmed transaction from address to be the appropriate account we need to sign with.

Tests: builds


hboon (approved):
Thank you for your contribution!

PR #6060, opened 2023-01-05
Extract logic from WalletConnect coordinator #6059

### LINKED ISSUE #6059: Extract logic from WalletConnect coordinator

oa-s:
Closes #6059


modules/AlphaWalletFoundation/AlphaWalletFoundation/Settings/Types/Constants+Credentials.swift
```diff
@@ -35,7 +35,7 @@ extension Constants {
         public static let analyticsKey = ""
         public static let mailChimpListSpecificKey = ""
         public static let walletConnectProjectId = env("WALLETCONNECTPROJECTID") ?? "8ba9ee138960775e5231b70cc5ef1c3a"
-        static let infuraKey = env("INFURAKEY") ?? "ad6d834b7a1e4d03a7fde92020616149"
+        static let infuraKey = env("INFURAKEY") ?? "5814931d81004ad9b3fcce706be81006"
```
hboon, line 38: Should not include this though. BTW, don't you use a `.credentials` file?

hboon (approved):
Too big to check. Presume it's just refactoring and that you tested :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Settings/Types/Constants+Credentials.swift
oa-s: oh, good catch , thanks!

PR #6063, opened 2023-01-05
Fix stucking on loading screen when call approve(), fix displaying small amounts when confirmation swap action. #6055

### LINKED ISSUE #6055: After initiating a swap on Polygon from WETH to MATIC and confirming the approve(), stuck at "Loading…"
Verified that transaction was confirmed on Polyscan and waited for a bit. Maybe the code to watch for the transaction completion and follow up with the actual swap transaction is not working?

After killing the app and initiating swap again (this time the approve() call wasn't necessary), the actionsheet to confirm the swap transaction appears and works correctly

oa-s:
Fixes #6055

### DIFF
--- a/AlphaWallet/Swap/Swap/SwapTokensCoordinator.swift
+++ b/AlphaWallet/Swap/Swap/SwapTokensCoordinator.swift
@@ -184,6 +184,15 @@ extension SwapTokensCoordinator: ApproveSwapProviderDelegate {
     func didFailure(in approveSwapProvider: ApproveSwapProvider, error: Error) {
         rootViewController.hideLoading()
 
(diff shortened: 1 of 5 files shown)


hboon (commented):
That seems like quite a lot of changes. Do we know why it didn't work before the PR?

oa-s:
> That seems like quite a lot of changes. Are all the changes needed? Do we know why it didn't work before the PR?

spliting beetwen separate prs

oa-s:
> Do we know why it didn't work before the PR?

yes, loading indicator wasn't  hidden on failure

oa-s:
After #6076

PR #6073, opened 2023-01-06
Fix encoding missing gasPrice for swap response #6072

### LINKED ISSUE #6072: Fix encoding missing gaslimit for swap response
sometimes response contains nil for gaslimit field

oa-s:
Fixes #6072

### DIFF
--- a/AlphaWallet/Swap/Swap/ViewModels/SwapQuoteDetailsViewModel.swift
+++ b/AlphaWallet/Swap/Swap/ViewModels/SwapQuoteDetailsViewModel.swift
@@ -48,7 +48,7 @@ final class SwapQuoteDetailsViewModel {
                             type: step.type,
                             amount: pair.swapQuote.unsignedSwapTransaction.value,
                             amountUsd: pair.swapQuote.action.fromToken.priceUSD,
-                            estimate: pair.swapQuote.unsignedSwapTransaction.gasPrice ?? .zero,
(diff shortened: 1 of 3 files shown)


AlphaWallet/Swap/Swap/ViewModels/SwapQuoteDetailsViewModel.swift
```diff
@@ -48,7 +48,7 @@ final class SwapQuoteDetailsViewModel {
                             type: step.type,
                             amount: pair.swapQuote.unsignedSwapTransaction.value,
                             amountUsd: pair.swapQuote.action.fromToken.priceUSD,
-                            estimate: pair.swapQuote.unsignedSwapTransaction.gasPrice ?? .zero,
+                            estimate: pair.swapQuote.unsignedSwapTransaction.gasPrice,
```
hboon, line 51: I didn't test this, what would happen when the `gasPrice` is missing from the swap response with this change?

Maybe we should check with the li.fi folks too (but still good to have a fix ourselves)

oa-s: it just displays an error when fetching swap routes, i have no idea why it happens, something with api i suppose, because one request (eth-> usdt swap ) returns an error an the next same request returns sucessfull response.

PR #6075, opened 2023-01-06
Fix displaying gasLimit error when configuring transaction #6074

### LINKED ISSUE #6074: Fix displaying gas Limit error when configuring transaction

oa-s:
Fixes #6074

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/GasLimitWarning.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/GasLimitWarning.swift
@@ -10,9 +10,6 @@ import Foundation
 extension TransactionConfigurator {
     public enum GasLimitWarning {
         case tooHighCustomGasLimit
-
-        public var description: String {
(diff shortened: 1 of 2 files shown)


oa-s:
| Before PR | After PR |
| ------ | ------ |
|  |  |

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/GasLimitWarning.swift
```diff
@@ -10,9 +10,6 @@ import Foundation
 extension TransactionConfigurator {
     public enum GasLimitWarning {
         case tooHighCustomGasLimit
-
```
hboon, line 13: This is strange. Why does it not work?

oa-s: Have no idea, maybe it's because it was defined in foundation pod, different namespaces.

hboon (approved):

PR #6081, opened 2023-01-06
Fix displaying small amounts when swap #6080

### LINKED ISSUE #6080: Fix displaying small amounts when swap

oa-s:
Fixes #6080


oa-s:
| Before PR | After PR |
| -------- | -------- |
|  |  |

PR #6084, opened 2023-01-09
Removed all direct references and definition of `Colors.alabaster`

eviltofu:
# Colors.alabaster/.white

## BrowserErrorView
1. From main wallet screen, tap browser button.
2. Enter 127.0.0.1 in address bar of browser.
3. The background has been changed to adapt to dark mode.

Before | Light After | Dark After
-|-|-
original|light|dark

## ButtonStyle.borderColor 

`.solid`, `.squared`, `.border` are not used as far as I can determine.

```swift
    var borderColor: UIColor {
        switch self {
        case .solid, .squared, .border: return Configuration.Color.Semantic.defaultButtonBorder
        case .borderless, .system, .special: return .clear
        case .green: return ButtonsBarViewModel.primaryButton.buttonBorderColor
        }
    }
```

### DIFF
--- a/AlphaWallet/Browser/Views/BrowserErrorView.swift
+++ b/AlphaWallet/Browser/Views/BrowserErrorView.swift
@@ -57,7 +57,7 @@ final class BrowserErrorView: UIView {
     }
 
     private func finishInit() {
-        backgroundColor = .white
+        backgroundColor = Configuration.Color.Semantic.defaultViewBackground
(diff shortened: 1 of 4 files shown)


hboon (approved):
Didn't test, but maybe better to be:

> Removed direct reference to `Colors.alabaster`

instead of:

> Removed Colors.alabaster

(if that's what this PR is about?)

PR #6093, opened 2023-01-09
[Refactor] Remove spurious `AlphaWallet.Address.sameContract(as: AlphaWallet.Address)` and replace with `==`

hboon:
Removes an unnecessary function.

@oa-s can you help to verify too? Have to be careful in particular for 2 types of replacements:

* When we replace !`sameContract(as:)` with `!=`
* When we replace `contains { sameContract(as:) }` with `!=`

Also, we already have `AlphaWalletAddressTests.testEquality()`


hboon:
Removes an unnecessary function.

@oa-s can you help to verify too? Have to be careful in particular for 2 types of replacements:

* When we replace !`sameContract(as:)` with !=
* When we replace `contains { sameContract(as:) }` with `!=`

oa-s (approved):

PR #6106, opened 2023-01-10
Updated Makefile to move the compressed chains file into the AlphaWallet/Resources directory

eviltofu:
Fixes #6104

No change required in Swift code as the file is referenced as `R.file.chainsZip` which is taken care of by R.swift.

PR #6107, opened 2023-01-10
Updated chains zip file

eviltofu:
After #6106 
`chains.zip` in `Resources` have been updated.


hboon:
@eviltofu actually, this PR can be merged before #6106 if it only contains the commit "Updated chains.zip file"

hboon:

hboon (approved):

PR #6113, opened 2023-01-10
Fix long waiting for wallet balance resolution #6033

### LINKED ISSUE #6033: Tapping on the top left (list of wallets) button in Wallet tab freezes app for ~5 seconds before list of wallets is shown
Only a few wallets loaded in the app

Seems like a combination of factors

oa-s:
Closes #6033


hboon (commented):
Why does the PR make it faster? (It's been behaving faster these last few days before this PR, but I have no idea why)

oa-s:
looks like when we subscribe for all tokens to resolve wallet balance we trying to resolve it on flight, fetch, calculate and so on. current approach returns publisher that is going to be be resolved, and makes balance fetch calculate async.

hboon (approved):
Can't reproduce the original very laggy performance. But *seems* to be faster now :)

PR #6116, opened 2023-01-10
Fix app crashing when trying to generate a link when private key unavailable #6114

### LINKED ISSUE #6114: Fix app crashing when trying to generate a link when private key unavailable

oa-s:
Fixes #6114

### DIFF
--- a/AlphaWallet/Common/Types/Error.swift
+++ b/AlphaWallet/Common/Types/Error.swift
@@ -34,6 +34,8 @@ extension KeystoreError {
             return R.string.localizable.keystoreAccessKeyNeedImportOrPasscode()
         case .userCancelled:
             return R.string.localizable.keystoreAccessKeyCancelled()
+        case .signDataIsEmpty:
+            return R.string.localizable.accountsErrorFailedToSignEmptyMessage()
         }
     }
 }
--- a/AlphaWallet/Localization/en.lproj/Localizable.strings
+++ b/AlphaWallet/Localization/en.lproj/Localizable.strings
@@ -421,6 +421,7 @@
 "accounts.error.accountNotFound" = "Account not found";
 "accounts.error.failedToSignMessage" = "Failed to sign message";
 "accounts.error.failedToExportPrivateKey" = "Failed to export private key";
+"accounts.error.failedToSignEmptyMessage" = "Failed to sign an empty message";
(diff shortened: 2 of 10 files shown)


hboon (commented):
@oa-s Can you describe how you reproduce the crash and which wallet and NFT contract + tokenId?

Note to self from Boon: I get a crash with an ExpansionPunks both before and after the PR. Want to know if it's a different problem

oa-s:
steps: 
- set flag `shouldPretendIsRealWallet = true`
- watch tomek.eth wallet
- enabled xDai chain, and look for `Tickers` collection
- open any item `Tickers` 
- tap sell
- go forvard for generation url link
- tap `generate` 

**expectation**:
- app displays an error

**actual**
- app crashes with error `accountMayNeedImportingAgainOrEnablePasscode`

oa-s:
@oa-s can u show error you got?

hboon:
@oa-s it doesn't crash anymore with 39b117bdc2249645498b7aa747a5f39136f7b515, but I get this:

Screenshot 2023-01-11 at 10 42 06 PM

oa-s:
@hboon can u try it one more time?

hboon:
Not crashing anymore. Thanks.

But maybe an unrelated problem: If we hardcode `generateSellLink(tokenHolder:,linkExpiryDate:,ethCost:,server:)` to return a URL string and then proceed to share, tapping on `Messages` does nothing. Console says:

> [MFMessageComposeViewController] Unable to initialize due to + [MFMessageComposeViewController canSendText] returns NO.

So maybe it's not a problem. But I thought it worked previously, the Messages sharing sheet at least popped up.

oa-s:
> > [MFMessageComposeViewController] Unable to initialize due to + [MFMessageComposeViewController canSendText] returns NO.
> 
> So maybe it's not a problem. But I thought it worked previously, the Messages sharing sheet at least popped up.

but we displaying `UIActivityViewController` and not `MFMessageComposeViewController`

hboon:
Yes, it should show it after we tap to share via Messages in the sharing sheet. But maybe it’s only a problem on simulator?

oa-s:
yes, that is only on simulator, works on real device

hboon:
Good enough :) thanks.

PR #6117, opened 2023-01-10
Add done button to amount test field on `EnterSellTokensCardPriceQuantity` screen #6095

### LINKED ISSUE #6095: `EnterSellTokensCardPriceQuantityViewController ` does not allow keyboard to be dismissed after entering token price
# Problem
User is unable to dismiss the keyboard to tap next. Process halts at this point and is not able to continue.

App is modified per instructions from https://github.com/AlphaWallet/alpha-wallet-ios/pull/5956#issuecomment-1358793677. 

I believe this escaped testing because the tester has "Connect hardware keyboard" and is using their physical keyboard to enter the price. 

Tested on iPhone hardware.

Ignore the dynamic island.

Screen Shot 2023-01-09 at 10 11 07 PM

oa-s:
Closes #6095


hboon (changes requested):
Tapping `Done` doesn't work for me. Maybe it's missing an implementation of `AmountTextFieldDelegate.doneButtonTapped(for:)`?

I wonder if the default empty implementation is too prone to error? This — https://github.com/AlphaWallet/alpha-wallet-ios/blob/028f36d79be9c443e180420e47ff7fb93650e86e/AlphaWallet/Common/Views/AmountTextField.swift#L16

PR #6119, opened 2023-01-10
Fix displaying total cost when sell token #6098

### LINKED ISSUE #6098: Price entered does not match total cost in `EnterSellTokensCardPriceQuantityViewController`
# Problem
The price entered does not match the total cost

# Set up
App is modified per instructions from https://github.com/AlphaWallet/alpha-wallet-ios/pull/5956#issuecomment-1358793677.

Screen_Shot_2023-01-09_at_10 11 07_PM_copy

oa-s:
Fixes #6098

### DIFF
--- a/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
+++ b/AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
@@ -6,7 +6,7 @@ import Combine
 import AlphaWalletFoundation
 
 protocol EnterSellTokensCardPriceQuantityViewControllerDelegate: AnyObject, CanOpenURL {
-    func didEnterSellTokensPriceQuantity(token: Token, tokenHolder: TokenHolder, ethCost: Ether, in viewController: EnterSellTokensCardPriceQuantityViewController)
+    func didEnterSellTokensPriceQuantity(token: Token, tokenHolder: TokenHolder, ethCost: Double, in viewController: EnterSellTokensCardPriceQuantityViewController)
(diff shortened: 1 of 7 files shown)


AlphaWallet/Sell/ViewControllers/EnterSellTokensCardPriceQuantityViewController.swift
```diff
@@ -79,14 +79,14 @@ class EnterSellTokensCardPriceQuantityViewController: UIViewController, TokenVer
     private let tokenRowView: TokenRowView & UIView
     private let buttonsBar = HorizontalButtonsBar(configuration: .primary(buttons: 1))
     private var viewModel: EnterSellTokensCardPriceQuantityViewModel
-    private var totalEthCost: Ether {
+    private var totalEthCost: Double {
```
hboon, line 82: I didn't check carefully, but isn't it better to pass around `Ether` than a `Double` for precision? Can we fix the formatting instead?

Also switching from `Ether` to `Double` would undo 4706f530a4a1431201e1dd29aea00f9a854f9583 which seems related to precision

```
commit 4706f530a4a1431201e1dd29aea00f9a854f9583
Date:   2018-08-22 16:08:24 +0800

    Fix: ticket total ETH price should be precise
```

oa-s:
case with Double works for me find, whether using conversations from double to bigInt and vice versa causes precession lost

hboon:
> case with Double works for me find, whether using conversations from double to bigInt and vice versa causes precession lost

What if we do things like sell more than 1 token?

oa-s:
> What if we sell more than 1 token?

checked with updating number of tokens, looks good, could you please check it on your side one more time?

hboon (approved):

PR #6120, opened 2023-01-11
    Forward select environment variables with Xcode test scheme so the keys that we set in Travis dashboard as environment variables is forwarded to the app

### LINKED ISSUE #6112: Make Travis tests use keys set in Travis dashboard
@hboon can u please create a new infura key and make it being used only for tests purposes, or just share a key with me and i will do the rest

hboon:
Closes #6112

### DIFF
--- a/AlphaWallet.xcworkspace/xcshareddata/xcschemes/AlphaWalletTests.xcscheme
+++ b/AlphaWallet.xcworkspace/xcshareddata/xcschemes/AlphaWalletTests.xcscheme
@@ -94,8 +94,88 @@
             isEnabled = "YES">
          </EnvironmentVariable>
          <EnvironmentVariable
+            key = "SOURCE_ROOT"
+            value = "$(SOURCE_ROOT)"
+            isEnabled = "YES">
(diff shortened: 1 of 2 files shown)


hboon:
@oa-s sorry, accidentally triggered adding you as review due to `CODEOWNERS`. Ignore for now.

oa-s (approved):

hboon:
Should be good to go. @oa-s can you help run it locally once to see if it works for you. When running locally, it ought to still use check for `.credentials` and then fallback to the hardcoded values in `Constants+Credentials.swift`.

oa-s:
that is strange can't find definitions of `EnvironmentVariable` variables in tests scheme

hboon:
See if you can find it in the diff. It’s linked from the ‘Run’ from ‘Test’ when you go to Xcode and edit scheme for the test target

hboon:
Like that. Under `Run` and not `Test`:

Screenshot 2023-01-11 at 10 35 10 PM

oa-s:
strange, might need to restart xcode

PR #6125, opened 2023-01-12
Semantic renaming of variable `Colors.appGreenContrastBackground` to `Configuration.Color.Semantic.textFieldContrastText`. No change in actual colour.

eviltofu:
Semantic renaming of variable. No change in actual colour.


hboon (approved):
This actually makes a better commit message and PR title:

> Semantic renaming of variable. No change in actual colour

PR #6129, opened 2023-01-12
Remove unnecessary pod dependencies from share extension so we remove build warning: Get rid of warning: "ld: linking against a dylib which is not safe for use in application extensions"

### LINKED ISSUE #6127: Get rid of warning: "ld: linking against a dylib which is not safe for use in application extensions"
> ld: linking against a dylib which is not safe for use in application extensions

hboon:
Fixes #6127

Includes #6128 

Not bazillion lines of these anymore, which makes build warnings too noisy to be useful:

> ld: linking against a dylib which is not safe for use in application extensions


hboon:
Force-pushed to fix commit message

oa-s (approved):

oa-s (approved):

PR #6148, opened 2023-01-16
display an error when `not enough gas` instead of displaying prompt action sheet #6092

### LINKED ISSUE #6092: If we are on testnet and there isn't enough gas, we should not display actionsheet to "Buy ETH"
Forgot what was the previous behavior (we should do that instead). Otherwise do nothing

Screenshot 2023-01-09 at 2 09 27 PM

oa-s:
Fixes #6092


hboon (approved):

hboon:
Should mention "testnet" in the commit message and PR title.

Maybe a separate issue, we have a regression so the nicer error message isn't showing up again:

Screenshot 2023-01-17 at 8 40 58 AM

oa-s:
now it should be good

PR #6161, opened 2023-01-18
Remove static assetAttributeProvider from XmlHandler

oa-s:
to reduce size of #6154


hboon:
@oa-s heads-up, I will get to this tomororw

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/TokenAdaptor.swift
```diff
@@ -82,6 +87,37 @@ public class TokenAdaptor {
         }
     }
 
+    public static func getTokenHolders(name: String,
```
hboon, line 90: Seems like this is unused?

hboon (approved):

PR #6178, opened 2023-01-18
add conformance to LocalizedError for SendTransactionNotRetryableError, tweak SendTransactionNotRetryableError #6158

### LINKED ISSUE #6158: SendTransactionNotRetryableError are not displayed in a user-friendly manner after user taps Confirm in the transaction confirmation actionsheet on testnet
1. Create new wallet (or any wallet with 0 ETH)
2. Send to self
3. Tap Confirm

Expected
---
4. Show a user friendly error message

Observed
---
4. We see the screenshot:

Screenshot 2023-01-18 at 9 48 19 AM

---

Maybe we can address it here: https://github.com/AlphaWallet/alpha-wallet-ios/blob/70ed519339cdf61a55d5f83649a63134afc6f67f/AlphaWallet/Common/Types/Error.swift#L80

I added this as `SendTransactionNotRetryableError.title` seems to have the user-friendly values:

```
case let error as SendTransactionNotRetryableError:
    let model = SendTransactionErrorViewModel(server: .goerli, error: error)
    return model.title
```

Maybe tweak it (extract the strings etc) if it's the correct behavior?

With the change:

Screenshot 2023-01-18 at 9 57 25 AM

oa-s:
Closes #6158

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/SendTransactionError.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/SendTransactionError.swift
@@ -2,15 +2,25 @@
 
 import Foundation
 
-public enum SendTransactionNotRetryableError: Error {
-    case insufficientFunds(message: String)
-    case nonceTooLow(message: String)
-    case gasPriceTooLow(message: String)
-    case gasLimitTooLow(message: String)
-    case gasLimitTooHigh(message: String)
-    case possibleChainIdMismatch(message: String)
-    case executionReverted(message: String)
-    case unknown(code: Int, message: String)
+public struct SendTransactionNotRetryableError: Error {
+    public enum ErrorType {
+        case insufficientFunds(message: String)
+        case nonceTooLow(message: String)
+        case gasPriceTooLow(message: String)
+        case gasLimitTooLow(message: String)
+        case gasLimitTooHigh(message: String)
+        case possibleChainIdMismatch(message: String)
+        case executionReverted(message: String)
+        case unknown(code: Int, message: String)
+    }
+
+    public let type: ErrorType
+    public let server: RPCServer
+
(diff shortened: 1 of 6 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/SendTransactionError.swift
```diff
@@ -2,15 +2,25 @@
...
-    case insufficientFunds(message: String)
-    case nonceTooLow(message: String)
-    case gasPriceTooLow(message: String)
-    case gasLimitTooLow(message: String)
-    case gasLimitTooHigh(message: String)
-    case possibleChainIdMismatch(message: String)
-    case executionReverted(message: String)
-    case unknown(code: Int, message: String)
+public struct SendTransactionNotRetryableError: Error {
+    public enum ErrorType {
```
hboon, line 6: Why is it neccessary to wrap it with a separate `ErrorType` enum?

hboon (approved):
Works for me.

hboon:
@oa-s BTW, the PR title and commit message should describe the "why" rather than the "what" if it's available. In this case maybe something like:

> Fix: show user-friendly message when send transaction fails, eg. not enough ether #6158

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transfer/Types/SendTransactionError.swift
oa-s: enum cant keep variables and as some of cases want't rpcserver i wrapped it into ErrorType with being able to keep server as well for error

hboon: Oh right. But `server` isn't used :P

PR #6182, opened 2023-01-19
[Refactor] Changed all Colors.clear to Configuration.Color.Semantic.backgroundClear and Configuration.Color.Semantic.borderClear. No change in actual color.

eviltofu:
I didn't see a need to have a color called clear when we already have .clear.

### DIFF
--- a/AlphaWallet/Browser/Views/DappBrowserNavigationBar.swift
+++ b/AlphaWallet/Browser/Views/DappBrowserNavigationBar.swift
@@ -57,7 +57,7 @@ final class DappBrowserNavigationBar: UINavigationBar {
         cancelEditingButton.setContentHuggingPriority(.defaultHigh, for: .horizontal)
         cancelEditingButton.isHidden = true
         cancelEditingButton.clipsToBounds = true
-        cancelEditingButton.backgroundColor = .clear
+        cancelEditingButton.backgroundColor = Configuration.Color.Semantic.backgroundClear
         cancelEditingButton.setTitleColor(Configuration.Color.Semantic.defaultForegroundText, for: .highlighted)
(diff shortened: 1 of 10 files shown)


hboon (commented):
I don't know why it was added, but keeping it might be better. Say we want to debug colors and figure out where/if it is use in a certain screen? We can override the value to `.red` with alpha and it'll jump out.

hboon:
If we keep it, can you add a comment to the color definition to explain why?

eviltofu:
> I don't know why it was added, but keeping it might be better. Say we want to debug colors and figure out where/if it is use in a certain screen? We can override the value to `.red` with alpha and it'll jump out.

We would need to split this into two. There is a clear for borders and clear for backgrounds. I'd split them into `clearBackground` and `clearBorder`. @hboon

oa-s:
> I don't know why it was added, but keeping it might be better. Say we want to debug colors and figure out where/if it is use in a certain screen? We can override the value to .red with alpha and it'll jump out.

the idea was to keep colors in single namespace, `Colors` as it was, its easier to track actual used colors and yes, override them when needed.

hboon:
>> I don't know why it was added, but keeping it might be better. Say we want to debug colors and figure out where/if it is use in a certain screen? We can override the value to .red with alpha and it'll jump out.
> We would need to split this into two. There is a clear for borders and clear for backgrounds. I'd split them into clearBackground and clearBorder

> Does this make sense?

@eviltofu Yes, this works; bottom line is to not replace them with `UIColor.clear`

hboon:
Remember to update PR title and commit messages too.

eviltofu:
Needs approval @hboon

hboon (approved):

PR #6194, opened 2023-01-20
Restrict token symbol to half screen width to avoid broken amount text field view #6185

### LINKED ISSUE #6185: Make sure amount textfield and All Funds button always show up and works even if the token symbol is too long
Symbol is probably too long? https://blockscout.com/xdai/mainnet/address/0xA68b7779504b0AE372ddCC109f8786DB9B91e93e

&nbsp;| &nbsp;| &nbsp;
-|-|-
|vs|Screenshot 2023-01-19 at 9 53 08 AM

oa-s:
Closes #6185

### DIFF
--- a/AlphaWallet/Common/Views/AmountTextField.swift
+++ b/AlphaWallet/Common/Views/AmountTextField.swift
@@ -53,7 +53,9 @@ final class AmountTextField: UIControl {
         button.heightConstraint.flatMap { NSLayoutConstraint.deactivate([$0]) }
         button.heightAnchor.constraint(equalToConstant: 25).isActive = true
         button.contentEdgeInsets = .zero
-
+        button.setContentHuggingPriority(.required, for: .horizontal)
+        button.setContentCompressionResistancePriority(.required, for: .horizontal)
+        
         return button
     }()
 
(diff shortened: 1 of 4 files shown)


hboon (commented):
Seeing this on iPhone 14 Pro iOS 16.2:

Screenshot 2023-01-25 at 8 25 42 AM

hboon (approved):
The little down arrow button to switch between crypto and fiat doesn't appear, but I guess it's ok since all (or at least most) crypto-fiat pairs should not have such a long symbol.

Can you help verify on a narrow device to see it works properly? I see it shows "1.23456" without truncating (it would truncate "silently" if there's 1 more digit, so would not be good if it's too narrow) on iPhone 14 Pro.

hboon:
> maybe its not available, i mean switching beetwen fiat, then its hidden, or you able to toogle beetwen currencies?

Oh, I forgot it's hidden if fiat rates aren't available. I hardcoded it to have a rate and it appears. All good.

Does the All Funds button display correctly with narrower devices?

oa-s:
> Does the All Funds button display correctly with narrower devices?

yes, still good

hboon (approved):
Haven't retested. But look good earlier.

PR #6195, opened 2023-01-20
[Refactor] Renamed Colors.green to Configuration.Color.Semantic.pass

eviltofu:
No actual color change.


eviltofu:
I haven't been able to check out this screen yet, I've only modified the code.

hboon:
> I've checked out the ActivityLoadingIndicatorControl and it appears to be too brief? 

Yes.

> Or is this based on external network + mining factors which we have no control over?

It depends on when the transaction is sent successfully (not depending on when it's mined). Can you help create an issue? We can look at how to improve it.

hboon (approved):

PR #6214, opened 2023-01-24
rename RawTransactionBridge to WalletConnectTransaction

oa-s:


AlphaWallet/WalletConnect/Types/WalletConnectRequestConverter.swift
```diff
@@ -30,9 +30,9 @@ struct WalletConnectRequestDecoder {
             return .value(.signMessage(message))
         case .signPersonalMessage(_, let message):
             return .value(.signPersonalMessage(message))
-        case .signTransaction(let bridgeTransaction):
```
hboon, line 33: @oa-s Remember to prefix with "[Refactor]" if it is just a rename and no change to functionality. Easier to include/exclude them.

PR #6217, opened 2023-01-25
drop initial 0 value when on send token screen #6216

### LINKED ISSUE #6216: drop initial `0` value when on send token screen

oa-s:
Closes #6216


oa-s:
|Before pr | After pr|
|----|----|
|||

hboon (approved):
Need to fix test suite

PR #6230, opened 2023-01-26
restore receiving some of network calls on background queue, like it was before

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/BlockscanChat/BlockscanChat.swift
```diff
@@ -25,6 +25,7 @@ public class BlockscanChat {
         infoLog("[BlockscanChat] Fetching unread count for \(address.eip55String)…")
         return networkService
             .dataTaskPublisher(GetUnreadCountEndpointRequest(address: address))
+            .receive(on: DispatchQueue.global())
```
hboon, line 28: @oa-s Why? Is it because it's too slow?

oa-s: yes, especially its able to see when on initial app lanuch, when loading tickers and other calls

PR #6234, opened 2023-01-27
When we sign a short message that is still long enough to be truncated, the "Show" button doesn't appear #6221

### LINKED ISSUE #6221: When we sign a short message that is still long enough to be truncated, the "Show" button doesn't appear
Message "Hello there, Boon" on iPhone 14 Pro. It does if the message is much longer

Short, but longer enough|Much longer
-|-
|

Tested with https://sigtool.ambire.com + WalletConnect

oa-s:
Closes #6221


hboon (changes requested):
Same problem now with "Hello there, Bo1" :)

oa-s:
now it should be ok

hboon:
Oh!

hboon (commented):
It looks like the `Show` button always show up even when the message is short:

Screenshot 2023-02-06 at 11 14 46 AM

The condition seems wrong. How about we simplify this and just always return true/false so the Show button is always shown?

```
private var availableToShowFullMessage: Bool {
    return false/true //I can't figure out which
}
```

oa-s:
> Show button is always shown?

that would be easier way, let's do that

oa-s:
updated pr with always displaying show more button

PR #6244, opened 2023-01-27
[Refactoring] Replace with class definitions ContractMethodCalls

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/Models/ContractMethodCalls/AnyContractMethodCall.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-struct AnyContractMethodCall: ContractMethodCall {
+class AnyContractMethodCall: ContractMethodCall {
```
hboon, line 10: Why?

PR #6246, opened 2023-01-27
[Refactoring] Replace back to class definition for XMLHandler

oa-s:


hboon:
What problems did making it a value object cause?

oa-s:
i moved it to struct to avoid momory leak, there was a ton of non delated XMLHandlers, but looks like structs cause bad access sometimes, could try to figure our memory leaks in different way

modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/XMLHandler.swift
```diff
@@ -725,7 +725,7 @@ final class ThreadSafe {
 }
 
 /// This class delegates all the functionality to a singleton of the actual XML parser. 1 for each contract. So we just parse the XML file 1 time only for each contract
-public struct XMLHandler {
+public class XMLHandler {
```
hboon, line 728: Why do you need to change it back to a reference type?

PR #6250, opened 2023-01-29
Fix bad access crash for FetchLatestTransactionsOperation, AutoDetectTransactedTokensOperation, FetchLatestTransactionsOperation #6249

### LINKED ISSUE #6249: Fix bas access crash for FetchLatestTransactionsOperation

oa-s:
Fixes #6249

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/AutoDetectTokensOperation.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/AutoDetectTokensOperation.swift
@@ -17,7 +17,7 @@ protocol AutoDetectTokensOperationDelegate: AnyObject {
 
 final class AutoDetectTokensOperation: Operation {
     private let tokens: [ContractToImport]
-    private var cancelable = Set<AnyCancellable>()
+    private var cancellable: AnyCancellable?
 
     weak private var delegate: AutoDetectTokensOperationDelegate?
     override var isExecuting: Bool {
@@ -40,7 +40,7 @@ final class AutoDetectTokensOperation: Operation {
     override func main() {
         guard let delegate = delegate else { return }
 
-        delegate.autoDetectTokensImpl(withContracts: tokens)
+        cancellable = delegate.autoDetectTokensImpl(withContracts: tokens)
             .sink(receiveCompletion: { _ in
 
             }, receiveValue: { values in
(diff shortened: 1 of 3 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/AutoDetectTokensOperation.swift
```diff
@@ -40,7 +40,7 @@ final class AutoDetectTokensOperation: Operation {
     override func main() {
         guard let delegate = delegate else { return }
 
-        delegate.autoDetectTokensImpl(withContracts: tokens)
+        cancellable = delegate.autoDetectTokensImpl(withContracts: tokens)
```
hboon, line 43: Why would the original crash? Seems standard.

hboon:
Typo in the commit message: "Fix bas access crash "

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/AutoDetectTokensOperation.swift
oa-s: accessing to set in background

hboon: I don't understand. Are you referring to the `.store(in: &cancelable)` call running in non-main because `autoDetectPartnerTokens()` runs in the background? But `cancelable` is only accessed from that one place, right?

hboon: We should also change `TokensAutodetector.start()` to not run in a non-main queue, maybe separate PR if necessary: https://github.com/AlphaWallet/alpha-wallet-ios/blob/91688ab5937220c1d6c2e37d9f0ce9c377b531a8/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/TokensAutodetector.swift#L69

This:

```
//Since this is called at launch, we don't want it to block launching
queue.async { [weak self] in
    self?.autoDetectTransactedTokens()
    self?.autoDetectPartnerTokens()
}
```

Async is fine (and probably necessary), but on main instead? The code being called runs `NSOperation`s anyway and can spin up other queues if they need to, right? Always `main` across boundaries make sense here too.

PR #6252, opened 2023-01-29
Fix bad access crash for fetchOlderTransactions #6251

### LINKED ISSUE #6251: Fix bas access crash for fetchOlderTransactions

oa-s:
Closes #6251

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/EtherscanSingleChainTransactionProvider.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/EtherscanSingleChainTransactionProvider.swift
@@ -38,7 +38,6 @@ class EtherscanSingleChainTransactionProvider: SingleChainTransactionProvider {
             ercTokenDetector: ercTokenDetector)
     }()
     private let networkService: NetworkService
-    private var cancelable = Set<AnyCancellable>()
 
     init(session: WalletSession,
          analytics: AnalyticsLogger,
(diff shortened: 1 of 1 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/EtherscanSingleChainTransactionProvider.swift
```diff
@@ -184,7 +183,7 @@ class EtherscanSingleChainTransactionProvider: SingleChainTransactionProvider {
 
         transactionsNetworkProvider
             .getTransactions(startBlock: 1, endBlock: oldestCachedTransaction.blockNumber - 1, sortOrder: .desc)
-            .sink(receiveCompletion: { [transactionsTracker] result in
+            .sinkAsync(receiveCompletion: { [transactionsTracker] result in
```
hboon, line 186: Wouldn't it be the same?

oa-s: there is an issue with accessing value type types like arrays, sets, dictionaries in background, updating it with `sinkAsync` avoids this issue as we don't store it in any array, and just release subscription when receive completion

hboon: Ah I see.

hboon (approved):

hboon:
"bas" typo here too

PR #6257, opened 2023-01-30
Fix memory leaks #6256

### LINKED ISSUE #6256: Fix memory leaks

oa-s:
Fixes #6256

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
@@ -14,38 +14,47 @@ import Combine
 
 final class JsonFromTokenUri {
     private let tokensService: TokenProvidable
-    private lazy var getTokenUri = NonFungibleContract(blockchainProvider: blockchainProvider)
+    private let getTokenUri: NonFungibleContract
     private let blockchainProvider: BlockchainProvider
     private var inFlightPromises: [String: AnyPublisher<NonFungibleBalanceAndItsSource<JsonString>, SessionTaskError>] = [:]
     private let queue = DispatchQueue(label: "org.alphawallet.swift.jsonFromTokenUri")
     //Unlike `SessionManager.default`, this doesn't add default HTTP headers. It looks like POAP token URLs (e.g. https://api.poap.xyz/metadata/2503/278569) don't like them and return `406` in the JSON. It's strangely not responsible when curling, but only when running in the app
     private let networkService: NetworkService
 
-    public init(blockchainProvider: BlockchainProvider, tokensService: TokenProvidable, networkService: NetworkService) {
+    public init(blockchainProvider: BlockchainProvider,
+                tokensService: TokenProvidable,
+                networkService: NetworkService) {
+
         self.networkService = networkService
         self.blockchainProvider = blockchainProvider
         self.tokensService = tokensService
+        self.getTokenUri = NonFungibleContract(blockchainProvider: blockchainProvider)
+    }
+
+    func clear() {
+        inFlightPromises.removeAll()
     }
 
     func fetchJsonFromTokenUri(forTokenId tokenId: String, tokenType: TokenType, address: AlphaWallet.Address, enjinToken: GetEnjinTokenQuery.Data.EnjinToken?) -> AnyPublisher<NonFungibleBalanceAndItsSource<JsonString>, SessionTaskError> {
         return Just(tokenId)
             .receive(on: queue)
             .setFailureType(to: SessionTaskError.self)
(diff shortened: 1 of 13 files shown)


oa-s:
should be before #6255

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
```diff
@@ -14,38 +14,47 @@ import Combine
...
     func fetchJsonFromTokenUri(forTokenId tokenId: String, tokenType: TokenType, address: AlphaWallet.Address, enjinToken: GetEnjinTokenQuery.Data.EnjinToken?) -> AnyPublisher<NonFungibleBalanceAndItsSource<JsonString>, SessionTaskError> {
         return Just(tokenId)
             .receive(on: queue)
             .setFailureType(to: SessionTaskError.self)
-            .flatMap { [queue, getTokenUri] tokenId -> AnyPublisher<NonFungibleBalanceAndItsSource<JsonString>, SessionTaskError> in
+            .flatMap { [weak self, queue, weak getTokenUri] tokenId -> AnyPublisher<NonFungibleBalanceAndItsSource<JsonString>, SessionTaskError> in
+                guard let strongSelf = self, let getTokenUri = getTokenUri else { return .fail(.cancelledError) }
                 let key = "\(tokenId).\(address.eip55String).\(tokenType.rawValue)"
 
-                if let promise = self.inFlightPromises[key] {
```
hboon, line 37: Maybe not the best way to have accessed `self` previously, but `.flatMap()`'s `transform` arg is escaping, why would it retain `self` and cause a leak?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/NonFungibleErc1155JsonBalanceFetcher.swift
```diff
@@ -20,27 +20,39 @@ class NonFungibleErc1155JsonBalanceFetcher {
...
+    }
+
     func fetchErc1155NonFungibleJsons(enjinTokens: EnjinTokenIdsToSemiFungibles) -> AnyPublisher<[AlphaWallet.Address: [NonFungibleBalanceAndItsSource<NonFungibleFromTokenUri>]], SessionTaskError> {
         return erc1155TokenIdsFetcher
             .detectContractsAndTokenIds()
             .mapError { SessionTaskError(error: $0) }
-            .flatMap { contractsAndTokenIds -> AnyPublisher<Erc1155TokenIds.ContractsAndTokenIds, SessionTaskError> in
-                self.addUnknownErc1155ContractsToDatabase(contractsAndTokenIds: contractsAndTokenIds.tokens)
-            }.flatMap { contractsAndTokenIds -> AnyPublisher<(contractsAndTokenIds: Erc1155TokenIds.ContractsAndTokenIds, tokenIdMetaDatas: [TokenIdMetaData]), SessionTaskError> in
-                self._fetchErc1155NonFungibleJsons(contractsAndTokenIds: contractsAndTokenIds, enjinTokens: enjinTokens)
```
hboon, line 43: Does the strong `self` reference here only cause a problem when the upstream code (`detectContractsAndTokenIds()` etc) don't emit any results/error?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
oa-s: but this sscaping closure captures self, and self captures publisher, that causes issue. we got strong ref to JsonFromTokenUri, in balacne fetcher, when we delete balance fetcher, e.g: switch rpc server, balance fetcher delates, (that is fine) but with it we want to delete all inner objects. so JsonFromTokenUri stays in memory

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/NonFungibleErc1155JsonBalanceFetcher.swift
oa-s: it causes an issue when its in progress, we ran function and when in same time switch servers, object should receive cancel as its parent dellocated

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
oa-s: flatMap from combine creates a new publisher. this escaping block captures self as well as self captures publisher

hboon: 😭 of course! I had read "escaping" as "non-escaping" (so I thought it should retain `self`). All good!

hboon (approved):

PR #6274, opened 2023-02-01
Add ability to support video for nft #6272

### LINKED ISSUE #6272: Add ability to support video for nft

oa-s:
Closes #6272


oa-s:
@hboon could you please check if it works for you?

hboon (commented):
Do you have an easy way to reproduce it? I tried one and the token instance isn't detected :)

oa-s:
you can watch wallet 0x7c1555a3cae07da85488e941949dfea2122c727f
on polygon there is erc721 token  with video

oa-s:
Screenshot 2023-02-02 at 12 53 32

oa-s:
for now instead of playing a view we display video preview image

hboon:
Ok

hboon:
Works for me, video (I presume so, since it animates) renders in Wallet tab and list of assets as well as single asset after I tap them. Did you notice if memory suffers? Especially with a few of them. How about in the Wallet tab, we only render the video in the Collectibles sub tab? (can do in a separate PR if it's cleaner).

Also a bug, but not sure if it's caused by this PR. The EDNS token's "cover" image is wrong but the tokens inside look correct:

Screenshot 2023-02-03 at 9 49 52 AM

hboon:
Created #6286, maybe not related

oa-s:
> we only render the video in the Collectibles sub tab? 

we display it on all of views

oa-s:
> Did you notice if memory suffers?
haven't seen that for now

hboon:
Oh yeah. I was suggesting we only show the video in the Collectibles tab. The other suv-tabs in Wallet have very small toke image views so we can see the video anyway.

hboon:
Done?

oa-s:
not yet, need to debug it more, to make sure its good, and add support `animation_url` handling for nft assets from open sea

oa-s:
@hboon pr is ready for review

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+                        .mapError { ContentLoaderError.internal($0) }
+                        .flatMap { response -> LoadContentPublisher in
+                            do {
+                                guard let resp = response.response as? HTTPURLResponse else {
+                                    throw ContentLoaderError.invalidData
+                                }
+                                let value = try decoder.decode(response: resp, data: response.data)
+                                try cache.set(data: value.data, for: ImageCacheKey.cacheKey(for: url, prefix: .raw))
+
+                                return interceptor.intercept(response: value.image)
```
hboon, line 144: This line doesn't read very well because we can't tell what it's doing. Do you forsee we need the extensibility? Otherwise this would be better right?

```
return generateStillIfVideo(response: value: image)
```

Still not so good, but at least it's not very generic like "intercept" which doesn't say what it's doing.

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+protocol ContentLoaderResponseInterceptor: AnyObject {
+    func intercept(response: ContentLoader.Content) -> ContentLoader.LoadContentPublisher
+}
+
+fileprivate enum ImageCacheKey: CaseIterable {
+    case raw
+    case videoPreview
```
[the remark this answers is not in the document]
hboon: Do you think it's better if we just have 2 caches — 1 for "raw", 1 for stills for videos. Then we can avoid the low level work of maintaining a prefix

If we want to go further, we can even split "raw" into 3 for .image, .svg, .video. Splitting this way should help like these:

* improved performance, eg. if a piece of content is a video, when reading from cache, it doesn't always try to convert to image and String first and fail before being recognized as a video
* we can assign different disk cache sizes (I think?)

(But I notice we don't cache the videos content ourselves)

AlphaWallet/Common/Views/ViewModels/AVPlayerViewModel.swift
```diff
@@ -0,0 +1,241 @@
...
+                asset.cancelLoading()
+            }
+        }.removeDuplicates()
+        .eraseToAnyPublisher()
+    }
+
+    func extractUIImageAsync(size: CGSize) -> AnyPublisher<UIImage?, ImageGeneratorError> {
+        let generator = _generator(size: size)
+        return AnyPublisher<UIImage?, ImageGeneratorError>.create { observer in
+            generator.generateCGImagesAsynchronously(forTimes: [NSValue(time: .zero)]) { (_, image, _, _, error) in
```
hboon, line 215: In your tests, does `.zero` usually give good results?

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+    init(cache: ContentCacheStorage) {
+        self.cache = cache
+    }
+
+    func intercept(response: ContentLoader.Content) -> ContentLoader.LoadContentPublisher {
+        switch response {
+        case .svg, .image:
+            return .just(.done(response))
+        case .video(let video):
+            return AVAsset(url: video.url)
```
hboon, line 63: The videos' still preview generation might fail since our input is a URL. Do we need to retry (where we access it, I think in `buildFetchPublisher()`) if we know it's a video but yet have no still preview?

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+
+        private func decode(data: Data, url: URL) throws -> Response {
+            if let image = UIImage(data: data) {
+                return (data: data, .image(image))
+            } else if let svg = String(data: data, encoding: .utf8) {
+                return (data: data, .svg(svg))
+            } else {
+                let data = try JSONEncoder().encode(VideoUrl(url: url))
```
hboon, line 228: So we don't cache the video's data like we do for images and SVGs right? Do we not do it because of size? But does it mean if user taps a token and load the video, dismiss it and taps the token again, the video might be downloaded again?

If we don't cache it to disk, should we cache it to memory with `ImageCache`? Does `ImageCache` purge the memory cache properly?

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+
+        return Just(urlRequest)
+            .receive(on: queue)
+            .setFailureType(to: ContentLoaderError.self)
+            .flatMap { [weak self, queue, session, cache, decoder, interceptor] urlRequest -> LoadContentPublisher in
+                if let publisher = self?.publishers[urlRequest] {
+                    return publisher
+                } else {
+                    let publisher = session.dataTaskPublisher(for: urlRequest)
+                        .receive(on: queue)
```
hboon, line 134: Is the `.receive(on: queue)` a few lines above this needed? Since it's called here before work starts. Might need to double check access to `publishers` no matter whether we remove the `receive(on:)` call.

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+                        .mapError { ContentLoaderError.internal($0) }
+                        .flatMap { response -> LoadContentPublisher in
+                            do {
+                                guard let resp = response.response as? HTTPURLResponse else {
+                                    throw ContentLoaderError.invalidData
+                                }
+                                let value = try decoder.decode(response: resp, data: response.data)
+                                try cache.set(data: value.data, for: ImageCacheKey.cacheKey(for: url, prefix: .raw))
```
hboon, line 142: Is it safe to call `cache.set(data:for:)` in a non-main thread? Seems we sometimes access `ImageCache` from `main`, sometimes not

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+                }.replaceError(with: .done(.video(video)))
+                .setFailureType(to: ContentLoader.ContentLoaderError.self)
+                .eraseToAnyPublisher()
+        }
+    }
+}
+
+final class ContentLoader {
+    private let queue = DispatchQueue(label: "org.alphawallet.swift.imageLoader.processingQueue", qos: .utility)
+    private var publishers: [URLRequest: LoadContentPublisher]
```
hboon, line 81: Rename to `publishersInFlight` maybe?

hboon (approved):
Didn't test. Left some comments.

AlphaWallet/Common/Services/ContentLoader.swift
```diff
@@ -0,0 +1,233 @@
...
+import AVKit
+import Kingfisher
+import AlphaWalletFoundation
+
+protocol ContentCacheStorage: AnyObject {
+    func value(for key: String) throws -> Data?
+    func set(data: Data, for key: String) throws
+}
+
+final class KingfisherImageCacheStorage: ContentCacheStorage {
```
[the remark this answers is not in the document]
oa-s: it stores only images and video urls, we are not storing videos

AlphaWallet/Common/Views/ViewModels/AVPlayerViewModel.swift
oa-s: zero its a first frame of video

AlphaWallet/Common/Services/ContentLoader.swift
oa-s: its actually calls only on main thread

AlphaWallet/Common/Services/ContentLoader.swift
oa-s: for now yes, its gonna to be downloaded again, only if caching not applied inside on AVAsset

AlphaWallet/Common/Views/ViewModels/WebImageViewModel.swift
```diff
@@ -0,0 +1,118 @@
...
+//
+//  Created by Vladyslav Shepitko on 22.02.2023.
+//
+
+import UIKit
+import Combine
+import AlphaWalletFoundation
+
+//TODO: move later most of logic from app coordinator here
+class App {
```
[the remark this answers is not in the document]
oa-s: yes, i am going to do it later

AlphaWallet/Common/Views/ViewModels/AVPlayerViewModel.swift
hboon: Yes. But not all videos start immediately :)

oa-s: yes, we can tweak it later

PR #6281, opened 2023-02-02
add replacing ipfs.infura.io with https://alphawallet.infura-ipfs.io #6280

### LINKED ISSUE #6280: add replacing ipfs.infura.io with `https://alphawallet.infura-ipfs.io`

oa-s:
Closes #6280

### DIFF
--- a/modules/AlphaWalletCore/AlphaWalletCore/URL+Extensions.swift
+++ b/modules/AlphaWalletCore/AlphaWalletCore/URL+Extensions.swift
@@ -23,6 +23,13 @@ extension URL {
             }()
             return URL(string: urlString)
         } else {
+            guard var components = URLComponents(url: self, resolvingAgainstBaseURL: false) else { return nil }
+            if components.host == "ipfs.infura.io" {
+                components.scheme = "https"
+                components.host = "alphawallet.infura-ipfs.io"
+
+                return components.url
+            }
             return nil
         }
     }


modules/AlphaWalletCore/AlphaWalletCore/URL+Extensions.swift
```diff
@@ -16,11 +16,15 @@ extension URL {
             }()
             return URL(string: urlString)
         } else {
+            guard let components = URLComponents(url: self, resolvingAgainstBaseURL: false) else { return nil }
+            if components.host == "ipfs.infura.io" {
+                return URL(string: "https://alphawallet.infura-ipfs.io" + components.path)
```
hboon, line 21: I wonder if it is better just do a search and replace for "https://ipfs.infura.io/" once the host check passes. It breaks *if* the URL also includes that string outside the host (which seems unlikely or impossible?), but otherwise it works with URLS that have query strings. Maybe with anchors too? Anchors are probably unlikely, but query strings seems possible.

eg. the "?x=a" in "https://ipfs.infura.io/ipfs/abc?x=a"

Unless there's a better way than `components.path`?

hboon (approved):

modules/AlphaWalletCore/AlphaWalletCore/URL+Extensions.swift
oa-s: updated pr replacing host and scheme, so path and params will be the same as in source

PR #6294, opened 2023-02-06
[Refactoring] Remove displaying sign message from uiview for token script, move to deletage function

oa-s:


hboon:
@oa-s don't forget to push to upstream repo instead to use keys

hboon (approved):
Only skimmed. Presumed it's just refactoring :)

PR #6296, opened 2023-02-07
avoid accessing token id by index, for non fungible screen, to avoid ndex out of range crash happens sometimes when not able to decode non fungible json value

oa-s:
not always reproducible but catched few times


hboon (commented):
There seems to be some refactoring and formatting mixed it, can you help move them to separate commits?

PR #6301, opened 2023-02-07
replace atomic array to avoid xcode block issue warnings

oa-s:


oa-s:
version with dispatch queue as we are not able to set priority for DispatchSemaphore

hboon (approved):
Presume it's ok :)

But we should reduce, or not use at all, AtomicArray and AtomicDictionary though.

PR #6308, opened 2023-02-08
[Refactoring] Remove unnecessary func arg

hboon:


oa-s (approved):
seems we have such changed on  https://github.com/AlphaWallet/alpha-wallet-ios/pull/5172

hboon:
@oa-s ok. Thanks for the heads-up. Will sort out if there's conflicts

PR #6331, opened 2023-02-14
[Refactoring] upgrade alamofire up to latest version

oa-s:
as part of #6304


hboon (approved):
Does 'pod upgrade <pod>' (or update, can't remember) work? Then we don't need to modify Podfile. Just modifies the lock file

oa-s:
i used pod `update Alamofire`

PR #6355, opened 2023-02-16
Check that chain ID in EIP-712 message is the same as the current/enabled chain before showing user the prompt to sign it #6330

### LINKED ISSUE #6330: Check that chain ID in EIP-712 message is the same as the current/enabled chain before showing user the prompt to sign it
- [ ] A. in the dapp browser and if not using WalletConnect (if the chain ID is different from the current one in the browser), return or display an error to the dapp. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You are currently on the Polygon [replace with current chain name] chain"
- [ ] B. for WalletConnect v1 (if the chain ID is different from the current one for the WC session), return error to WalletConnect server. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You are currently on the Polygon [replace with current chain name] chain"
- [ ] C. for WalletConnect v2 (if the chain ID is not included in the list of active chains for the WC session), return error to WalletConnect server. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You need to enable this chain for your WalletConnect session"

oa-s:
Closes #6330

### DIFF
--- a/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
+++ b/AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
@@ -208,13 +208,38 @@ final class DappBrowserCoordinator: NSObject, Coordinator {
         browserViewController.goTo(url: url)
     }
 
+    private func validateMessage(session: WalletSession, message: SignMessageType) -> AnyPublisher<Void, PromiseError> {
+        do {
+            switch message {
+            case .eip712v3And4(let typedData):
+                let validator = DappOrTokenScriptEip712v3And4Validator(server: session.server, source: .dappBrowser)
+                try validator.validate(message: typedData)
+            case .typedMessage(let typedData):
+                let validator = TypedMessageValidator()
+                try validator.validate(message: typedData)
+            case .message, .personalMessage:
+                break
+            }
+            return .just(())
+        } catch {
+            return .fail(PromiseError(error: error))
+        }
+    }
+
     private func requestSignMessage(session: WalletSession,
                                     delegate: DappBrowserCoordinatorDelegate,
                                     message: SignMessageType,
                                     callbackId: Int) {
 
-        delegate.requestSignMessage(message: message, server: session.server, account: session.account.address, source: .dappBrowser, requester: nil)
-            .sink(receiveCompletion: { [browserViewController] result in
+        validateMessage(session: session, message: message)
+            .flatMap { _ in
+                delegate.requestSignMessage(
+                    message: message,
+                    server: session.server,
+                    account: session.account.address,
+                    source: .dappBrowser,
+                    requester: nil)
+            }.sink(receiveCompletion: { [browserViewController] result in
                 guard case .failure = result else { return }
                 browserViewController.notifyFinish(callbackID: callbackId, value: .failure(DAppError.cancelled))
             }, receiveValue: { [browserViewController] data in
@@ -467,6 +492,7 @@ extension DappBrowserCoordinator: BrowserViewControllerDelegate {
                 message: .typedMessage(typedData),
                 callbackId: callbackId)
         case .signEip712v3And4(let typedData):
+            
             requestSignMessage(
                 session: session,
                 delegate: delegate,
--- a/AlphaWallet/Transfer/Coordinators/SignMessageCoordinator.swift
+++ b/AlphaWallet/Transfer/Coordinators/SignMessageCoordinator.swift
@@ -5,8 +5,27 @@ import UIKit
 import AlphaWalletFoundation
 
 protocol SignMessageCoordinatorDelegate: AnyObject {
-    func coordinator(_ coordinator: SignMessageCoordinator, didSign result: Swift.Result<Data, KeystoreError>)
+    func coordinator(_ coordinator: SignMessageCoordinator, didSign result: Data)
     func didCancel(in coordinator: SignMessageCoordinator)
+}
+
+extension SignMessageValidatorError: LocalizedError {
+    var localizedDescription: String {
+        switch self {
+        case .emptyMessage:
+            return R.string.localizable.signMessageValidationEmptyMessage()
+        case .notMatchesToChainId(let active, let requested, let source):
+            switch source {
+            case .dappBrowser, .deepLink, .tokenScript:
+                return R.string.localizable.signMessageValidationDappBrowserRequestedChainUnavailable(active.name, requested.name)
+            case .walletConnect:
+                return R.string.localizable.signMessageValidationWalletConnectV1RequestedChainUnavailable(active.name, requested.name)
+            }
+        case .notMatchesToAnyOfChainIds(let active, _, _):
+            let active = String(active.map { $0.name }.joined(by: ","))
+            return R.string.localizable.signMessageValidationWalletConnectV2RequestedChainUnavailable(active)
+        }
+    }
 }
 
 class SignMessageCoordinator: Coordinator {
@@ -17,7 +36,6 @@ class SignMessageCoordinator: Coordinator {
     private let message: SignMessageType
     private let source: Analytics.SignMessageRequestSource
     private weak var signatureConfirmationDetailsViewController: SignatureConfirmationDetailsViewController?
-    private var canBeDismissed = true
     private let requester: RequesterViewModel?
     private lazy var rootViewController: SignatureConfirmationViewController = {
         let controller = SignatureConfirmationViewController(viewModel: .init(message: message, requester: requester))
@@ -50,7 +68,7 @@ class SignMessageCoordinator: Coordinator {
 
     func start() {
         analytics.log(navigation: Analytics.Navigation.signMessageRequest, properties: [
-            Analytics.Properties.source.rawValue: source.rawValue,
+            Analytics.Properties.source.rawValue: source.description,
             Analytics.Properties.messageType.rawValue: mapMessageToAnalyticsType(message).rawValue
         ])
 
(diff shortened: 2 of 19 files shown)


AlphaWallet/Browser/Coordinators/DappBrowserCoordinator.swift
```diff
@@ -485,7 +493,14 @@ extension DappBrowserCoordinator: BrowserViewControllerDelegate {
...
                 signMessage(with: .typedMessage(typedData), account: account, callbackID: callbackID)
             case .signEip712v3And4(let typedData):
-                signMessage(with: .eip712v3And4(typedData), account: account, callbackID: callbackID)
+                do {
+                    guard let session = sessionsProvider.session(for: server) else { throw DappBrowserError.serverUnavailable }
+                    let typedData = try verifyIfEip712v3And4(typedData, walletSession: session)
+                    signMessage(with: .eip712v3And4(typedData), account: account, callbackID: callbackID)
+                } catch {
+                    browserViewController.notifyFinish(callbackID: callbackID, value: .failure(DAppError.cancelled))
+                    navigationController.topViewController?.displayError(error: ActiveWalletViewModel.Error.onlyWatchAccount)
```
hboon, line 502: Have to change the error message, or error shown

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/EIP712TypedData.swift
```diff
@@ -35,6 +35,34 @@ public struct EIP712TypedData: Codable {
         }
     }
 
+    public var server: RPCServer? {
+        switch domain {
+        case .object(let dictionary):
+            switch dictionary["chainId"] {
+            case .string(let str):
+                if let server = eip155URLCoder.decodeRPC(from: str) {
```
hboon, line 43: I think it shouldn't ever be be EIP155 format? Just `"123"` or `123`? (it probably shouldn't be `"123"` too, but we can loose a little I guess.

EIP-712 (https://eips.ethereum.org/EIPS/eip-712) says:

> uint256 chainId the [EIP-155](https://eips.ethereum.org/EIPS/eip-155) chain id. The user-agent should refuse signing if it does not match the currently active chain.

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/EIP712TypedData.swift
```diff
@@ -35,6 +35,34 @@ public struct EIP712TypedData: Codable {
...
+                    return server
+                } else if let value = Int(str, radix: 10) {
+                    return RPCServer(chainID: value)
+                } else if let value = Int(str) {
```
hboon, line 47: Isn't this the same as with `radix: 10` above?

If we keep just 1, let's keep the `radix: 10` variation because sometimes chain IDs are expected to be base 16, so it's good to be explicit in our handling

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/EIP712TypedData.swift
```diff
@@ -35,6 +35,34 @@ public struct EIP712TypedData: Codable {
...
+                    return RPCServer(chainID: value)
+                } else {
+                    return nil
+                }
+
+            case .number(let value):
+                if let value = Int(value.description) {
```
hboon, line 54: Is `Int(value)` good enough?

AlphaWallet/WalletConnect/WalletConnectServer.swift
```diff
@@ -55,6 +58,8 @@ enum WalletConnectError: Error {
             return R.string.localizable.walletConnectFailureWalletsNotFound()
         case .delayedOperation, .cancelled:
             return nil
+        case .chainIdNotMatches(let server):
+            return "You were requested to sign a message for the \(server.name) chain. You need to enable this chain for your WalletConnect session"
```
hboon, line 62: Remember to update localized strings files

AlphaWallet/WalletConnect/WalletConnectProvider.swift
```diff
@@ -258,6 +264,14 @@ extension WalletConnectProvider: WalletConnectServerDelegate {
             .eraseToAnyPublisher()
     }
 
+    private func verifyIfEip712v3And4(_ typedData: EIP712TypedData, walletSession: WalletSession) throws -> EIP712TypedData {
+        if let server = typedData.server, walletSession.server != server {
```
hboon, line 268: Does this only work for WalletConnect v1? Or does it cover v2 too?

Context: https://github.com/AlphaWallet/alpha-wallet-ios/issues/6330#issue-1583535346

AlphaWallet/WalletConnect/WalletConnectProvider.swift
```diff
@@ -258,6 +264,14 @@ extension WalletConnectProvider: WalletConnectServerDelegate {
             .eraseToAnyPublisher()
     }
 
+    private func verifyIfEip712v3And4(_ typedData: EIP712TypedData, walletSession: WalletSession) throws -> EIP712TypedData {
```
hboon, line 267: And like the other `verifyIfEip712v3And4` (for dapp browser), better to return `Bool` + functional?

AlphaWallet/WalletConnect/WalletConnectProvider.swift
```diff
@@ -299,12 +313,18 @@ extension WalletConnectProvider: WalletConnectServerDelegate {
...
+                return dappRequestProvider.requestSignMessage(
+                    message: .eip712v3And4(typedData),
+                    account: walletSession.account.address,
+                    requester: requester)
+                .mapError { WalletConnectError(error: $0) }
+                .eraseToAnyPublisher()
+            } catch let e as WalletConnectError {
+                return .fail(e)
+            } catch { /*no-op*/ fatalError("inaccessible code path") }
         case .typedMessage(let typedData):
```
hboon, line 328: Can you help check if the old EIP-712 (which is what `.typedMessage` is for) has a chain ID too? (also for dapp browser handling)

AlphaWallet/WalletConnect/WalletConnectProvider.swift
```diff
@@ -299,12 +313,18 @@ extension WalletConnectProvider: WalletConnectServerDelegate {
...
+                let typedData = try verifyIfEip712v3And4(typedData, walletSession: walletSession)
+                return dappRequestProvider.requestSignMessage(
+                    message: .eip712v3And4(typedData),
+                    account: walletSession.account.address,
+                    requester: requester)
+                .mapError { WalletConnectError(error: $0) }
+                .eraseToAnyPublisher()
+            } catch let e as WalletConnectError {
+                return .fail(e)
+            } catch { /*no-op*/ fatalError("inaccessible code path") }
```
hboon, line 327: Not a `no-op` if it calls `fatalError()` :)

hboon (changes requested):

AlphaWallet/WalletConnect/WalletConnectProvider.swift
oa-s: it works for both

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/EIP712TypedData.swift
oa-s: guess should be

oa-s:
@hboon pr is ready for review

AlphaWallet/Transfer/Coordinators/SignMessageCoordinator.swift
```diff
@@ -5,35 +5,80 @@ import UIKit
...
+
+    enum ValidatorError: Error {
+        case emptyMessage
+        case chainIdNotMatches(activeServer: RPCServer, server: RPCServer, source: Analytics.SignMessageRequestSource)
+    }
+
+    let server: RPCServer
+    let source: Analytics.SignMessageRequestSource
+
+    func validate(mesage: SignMessageType) throws {
```
hboon, line 22: Let's keep this for now. But I'm on the fence on this: should we return bools for such validations or should we throw?

pros for throw
---
* more than 2 values (but we can return enums or optional enums but by then, throwing seems obviously better)
* keep the positive path local. More consistent with railway/reactive

pros for bool
---
* clearer and local
* easier to be sure we handle it

@oa-s any thoughts and more pros and cons?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Analytics/AnalyticsTypes.swift
```diff
@@ -213,7 +213,8 @@ public enum Analytics {
         case dappBrowser
         case deepLink
         case tokenScript
-        case walletConnect
+        case walletConnectV1
```
hboon, line 216: I think we should keep it as `walletConnect` but if you want to differentiate them, add a "type" property?

Rationale: so we keep the already-logged analytics compatible and also v1 is going away anyway, there wouldn't be much of v1 and would trend towards 0 soon.

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/EIP712TypedData.swift
hboon: `Int(value)` ? :)

hboon (changes requested):
A. Can you help change the commit message from:

> add sign message validation

to:

> Validate that chain ID in EIP-712 message is the same as the current chain according to EIP-712 specs

B. If I enable Goerli only, use WalletConnect v2 with https://react-app.walletconnect.com, and then tap eth_signTypedData, it complains the chain is not enabled:

Screenshot 2023-02-21 at 8 14 39 AM

oa-s:
> Int(value) ? :)

yes, `[EIP712TypedData](doc://com.apple.documentation.unknown/documentation/alphawalletfoundation/eip712typeddata).[NumberValue](doc://com.apple.documentation.unknown/documentation/alphawalletfoundation/eip712typeddata/numbervalue))` represents Double or Int, so converting shouldn't cause issues

oa-s:
> B. If I enable Goerli only, use WalletConnect v2 with https://react-app.walletconnect.com/, and then tap eth_signTypedData, it complains the chain is not enabled:

yes, right missed that case

PR #6362, opened 2023-02-17
[Refactoring] remove unused param

oa-s:


hboon:
@oa-s out if curiosity. Do you not use merge queues?

PR #6367, opened 2023-02-20
[Refactor] Changed UIColors into Configuration.Color.Semantic.xxx

eviltofu:
Changed UIColors into Configuration.Color.Semantic.xxx


hboon (approved):
Too many to check. But I presume they are mapped correctly, even if the variable color names might be a bit off (since if they are, we can still change them later on)

PR #6368, opened 2023-02-20
[Refactoring] grouping dapp requests in ActiveWalletCoordinator

oa-s:


hboon (approved):
Ok. I just looked at the structure, but not the exact code that was moved.

hboon:
Let me know if you need me to check in detail :P

PR #6400, opened 2023-02-24
fix sending dapp txs #6394

### LINKED ISSUE #6394: Posting transctions in dapp browser doesn't work
Uniswap + Goerli

Context: https://discord.com/channels/548357107267928064/801331663522627604/1078401125457350727

oa-s:
Fixes #6394


hboon (approved):
@oa-s I guess this is due to a recent refactoring that actually fixed signing but broke this functionality because this piece of code happened to be wrong too, right?

PR #6407, opened 2023-02-27
Updated tokens.json and added code to allow Spam group to be ignored from tokens.json file without crashing app

### LINKED ISSUE #6385: Update tokens.json
With copy: https://github.com/AlphaWallet/tokens/blob/a07813077de4fc77e96c360dc67c211cf6c52a87/tokens.json#L1

There is a new group called "Spam". It currently crashes when it encounters that. Ignore it for now, making it not crash

eviltofu:
closes #6385 

Added `public func hasContract(address: String, chainID: Int) -> Bool` in order to do allow test to detect if spam group entries are present. Should this function protocol and implementation be wrapped in `#ifdef DEBUG...#endif`?


hboon (changes requested):
We should update the commit message to mention that `tokens.json` is updated (because otherwise it seems like only code is updated to ignore spam, but groups and tokens didn't change. We might look for this commit in the near future). Maybe to this:

> Updated tokens.json and added code to allow Spam group to be ignored from tokens.json file without crashing app #6407

(alternatively 2 PRs, 1 for the code, another for the `tokens.json`)

So the tokens that are in the group `.spam` are just unused because we don't have code that reads it like the other groups (such as `.governance`) right? — if so, that's good 

Why do we need to have test case that ensure spam groups are in the file? Is that as a heuristic that the file is valid? — if so, Ok.

> Added public func hasContract(address: String, chainID: Int) -> Bool in order to do allow test to detect if spam group entries are present. Should this function protocol and implementation be wrapped in #ifdef DEBUG...#endif?

Is this because it's only used in tests? It's ok. Just leave it

eviltofu:
> So the tokens that are in the group `.spam` are just unused because we don't have code that reads it like the other groups (such as `.governance`) right? — if so, that's good

Technically if it's not classified in any group, it defaults to .asset. I should look further into this. Should the .asset definition be modified later to ignore .spam? I filter out all .spam groups via 
```swift
self.decodedTokenEntries = decodedTokenEntries.filter { _, group in
    return group != TokenGroup.spam
}
```

> 
> Why do we need to have test case that ensure spam groups are in the file? Is that as a heuristic that the file is valid? — if so, Ok.
> 

It's to test to see if any .spam entries are present (i.e., my filtering failed).

> > Added public func hasContract(address: String, chainID: Int) -> Bool in order to do allow test to detect if spam group entries are present. Should this function protocol and implementation be wrapped in #ifdef DEBUG...#endif?
> 
> Is this because it's only used in tests? It's ok. Just leave it

hboon:
> Technically if it's not classified in any group, it defaults to .asset. I should look further into this. Should the .asset definition be modified later to ignore .spam? I filter out all .spam groups via

The effect should be equivalent to processing `tokens.json` after manually removing the spam group from that file

eviltofu:
> > Technically if it's not classified in any group, it defaults to .asset. I should look further into this. Should the .asset definition be modified later to ignore .spam? I filter out all .spam groups via
> 
> The effect should be equivalent to processing `tokens.json` after manually removing the spam group from that `tokens.json`

That is the current effect now. The `.spam` group has been removed after reading from `tokens.json`.

hboon (approved):
Thanks

PR #6420, opened 2023-02-28
Remove mutually exclusive mode for mainnet and testnet #6034

### LINKED ISSUE #6034: Remove mutually exclusive mode for mainnet and testnet so mainnet chains are always available, but testnets are off by default until users flip the switch and acknowledge
A. Remove the switch for mainnet
B. Keep the switch for testnet — unchanged
C. Keep the prompt for acknowledgment when enabling testnets
D. List of mainnets should always be available in the list of chains
E. The header in Wallet tab for the current testnet-mode should be removed
F. We always show the header with fiat value for mainnets that are enabled (if any, otherwise $0), in Wallet tab

Screenshot 2022-12-30 at 3 23 46 PM
Screenshot 2022-12-30 at 3 23 50 PM

oa-s:
Closes #6034


hboon:

hboon (approved):

PR #6421, opened 2023-02-28
add retrying delay handling for open sea, `retry-after` field from http response for openSea #6304

### LINKED ISSUE #6304: add handling retrying delay for open sea, `retry-after` field for http response for openSea

oa-s:
Closes #6304


AlphaWallet/Tokens/Collectibles/ViewModels/NFTAssetViewModel.swift
```diff
@@ -105,10 +106,14 @@ class NFTAssetViewModel {
     func transform(input: NFTAssetViewModelInput) -> NFTAssetViewModelOutput {
         let whenOpenSeaStatsHasChanged = PassthroughSubject<Void, Never>()
         if let collectionId = tokenHolder.values.collectionId, collectionId.trimmed.nonEmpty {
-            nftProvider.collectionStats(collectionId: collectionId).done { [weak self] stats in
-                self?.configure(overiddenOpenSeaStats: stats)
-                whenOpenSeaStatsHasChanged.send(())
-            }.cauterize()
+            nftProvider.collectionStats(collectionId: collectionId)
+                .receive(on: RunLoop.main)
```
hboon, line 110: Shouldn't upstream be already on `.main` by convention? If it isn't, can we make it so in a separate PR?

AlphaWallet/Tokens/Collectibles/ViewModels/NFTCollectionInfoPageViewModel.swift
```diff
@@ -81,7 +81,7 @@ final class NFTCollectionInfoPageViewModel {
             .flatMapLatest { [nftProvider] tokenHolder -> AnyPublisher<Loadable<Stats, PromiseError>, Never> in
                 if let collectionId = tokenHolder?.values.collectionId, collectionId.trimmed.nonEmpty {
                     return nftProvider.collectionStats(collectionId: collectionId)
-                        .publisher()
+                        .receive(on: RunLoop.main)
```
hboon, line 84: Shouldn't upstream be already on `.main` by convention? If it isn't, can we make it so in a separate PR?

modules/AlphaWalletFoundation/AlphaWalletFoundation/NFT/NFTProvider.swift
```diff
@@ -53,8 +59,13 @@ public final class AlphaWalletNFTProvider: NFTProvider {
     public func nonFungible() -> AnyPublisher<NonFungiblesTokens, Never> {
         let key = AddressAndRPCServer(address: wallet.address, server: server)
 
-        let tokensFromOpenSeaPromise = getOpenSeaNonFungible().publisher(queue: .global()).replaceError(with: [:])
-        let enjinTokensPromise = enjin.fetchTokens(wallet: wallet).receive(on: DispatchQueue.global()).mapToVoid().replaceError(with: ())
+        let tokensFromOpenSeaPromise = openSea.nonFungible(wallet: wallet)
+            .replaceError(with: [:])
+
+        let enjinTokensPromise = enjin.fetchTokens(wallet: wallet)
+            .receive(on: DispatchQueue.global())
```
hboon, line 66: Ah, I see it used to be `.global` before this PR. Maybe we can switch this back to `.main`?

modules/AlphaWalletOpenSea/AlphaWalletOpenSea/OpenSea.swift
```diff
@@ -19,72 +20,203 @@ public protocol OpenSeaDelegate: AnyObject {
...
+        super.init(retryableHTTPStatusCodes: Set([429, 408, 500, 502, 503, 504]))
+    }
+    
+    override func retry(_ request: Alamofire.Request,
+                        for session: Session,
+                        dueTo error: Error,
+                        completion: @escaping (RetryResult) -> Void) {
+
+        if request.retryCount < retryLimit, shouldRetry(request: request, dueTo: error) {
+            if let httpResponse = request.response, let delay = OpenSeaRetryPolicy.retryDelay(from: httpResponse) {
```
hboon, line 62: Maybe we can check for this `first` and if goes to `else` then call the super class (`RetryPolicy`) version? Then we don't have to handle and duplicate the computation for exponential backoff ourselves.

hboon: Also, I think the gist of retrying is here and where this policy is installed right? The rest are mostly refactoring.

hboon (approved):
Didn't check very carefully. Was looking out for the retry code.

PR #6458, opened 2023-03-03
Fix broken reloading of wallet connect session when switching rpc server #6377

### LINKED ISSUE #6377: List of WalletConnect connections screen don't refresh when a session's chain is changed from the dapp
Tried with https://app.uniswap.org/#/swap

It updates if we go back and display the list again:

Screenshot 2023-02-21 at 11 23 45 AM

oa-s:
Fixes #6377


hboon (approved):
Didn't check carefully, but seems we can exclude the commit from the other PR so this can be merged first?

PR #6470, opened 2023-03-06
handle receiving wallet connect url with `awallet` scheme value #6466

### LINKED ISSUE #6466: awallet://wc regression
It looks like the native deeplink `awallet://` does not work for `/wc` (WalletConnect V1) anymore. 

https://user-images.githubusercontent.com/851393/222998440-4629831c-c32a-48e6-9cb5-ef4101ee9837.MP4

It would be great if native deep links remain usable.

Universal deep links are not sufficient and highly unreliable, see the following screenshot why:

Screenshot 2023-03-05 at 18 10 02

On iOS universal deep links are optimistic, if internet connectivity is not sufficient or safari deems a universal link is not a direct trigger of a user click, it will send you either to the app store or will stay in safari without opening the native app at all. The "universal" in universal link, means that the content can be displayed in the browser, which is not true for a wallet connection and nobody should rely on universal links when connecting to mobile wallets on a mobile device.

You can test the behavior yourself here: https://depayfi.github.io/widgets/demo.connect.html

oa-s:
Closes #6466


hboon:
Can you help add a test?

hboon (approved):
Thanks!

PR #6496, opened 2023-03-10
fetch collection name without creating token holder

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/Models/TokenScriptSupportable.swift
```diff
@@ -27,12 +27,8 @@ public extension TokenAdaptor {
     }
 
     func titleInPluralFormOptional(token: TokenScriptSupportable) -> String? {
-        if let tokenHolders = getTokenHolders(token: token).first {
-            guard let name = tokenHolders.tokens.first?.values.collectionValue?.name, name.nonEmpty else { return nil }
-            return name
-        } else {
-            return nil
-        }
+        guard let collection = token.balanceNft.first?.nonFungibleBalance?.collection else { return nil }
```
hboon, line 30: @oa-s This breaks if the token is a fungible, right?

oa-s: Firstly this method was added to display nft collection name, we fetched it from token holder, but easily it could be fetched right from nft collection.

PR #6497, opened 2023-03-11
tweak displaying error

oa-s:


AlphaWallet/Common/Types/Error.swift
```diff
@@ -73,12 +73,12 @@ extension Error {
             return error.localizedDescription
         case let error as SignMessageValidatorError:
             return error.localizedDescription
+        case let error as SessionTaskError:
```
hboon, line 76: Created #6498, but not sure if it can be done

PR #6507, opened 2023-03-13
Make it possible to enable/disable chains without rebooting app UI #6035

### LINKED ISSUE #6035: Make it possible to enable chains when the WalletConnect prompt to connect a session prompts for a disabled chain
So user doesn't have to do scan the QR code twice like this:

1. Have X enabled only
2. Scans WalletConnect QR code which ask for Y (or for WalletConnect v2, both X and Y)
3. Cancel prompt
4. Enable Y
5. Generate another QR code and scan again

Code has to be modified so that enabling/disabling chains on the fly (without UI) doesn't reload the entire app's UI. We can work on the UI especially for WalletConnect as a separate PR.

Works with #6034 so testnets can be enabled

Follow up:

- [ ] UI change in Settings
- [ ] UI change for WalletConnect v1 and v2
- [ ] UI change for other chain switching prompt (dapp browser, import token, what else?)

oa-s:
Closes #6035


oa-s:
should be good for testing

hboon (approved):
Too big diff. Didn't check the code.

But tried this and this worked for me with the WalletConnect v2 test dapp:

1. Don't have Goerli and Optimistic Goerli enabled
2. Scan WalletConnect QR code for Goerli and Optimistic Goerli
3. Tap Accept
4. It tells me to enable
5. Tap Enable button
6. Manually enable Goerli and Optimistic Goerli
7. Tap x to save
8. Tap Accept
9. Connected and can sign

I retried with modifying step 6 to only enable Optimistic Goerli and it forced me through step 5-7 again, which is good.

Can we create this as a follow up issue? But not high priority, change these:

Enable button

5\. Tap Enable button
6\. Manually enable Goerli and Optimistic Goerli
7\. Tap x to save

to:

5\. Tap Enable button
6\. Automatically enable Goerli and Optimistic Goerli (and not needing to save manually)

hboon:
Sometimes the "x" button to close and save doesn't work. I haven't figured out why and if it's just with this PR:

Screenshot 2023-03-20 at 9 42 49 AM

Would you help create an issue if you can't reproduce right away?

oa-s:
> Sometimes the "x" button to close and save doesn't work. I haven't figured out why and if it's just with this PR:

yep catched it on starting of development of this feature, but then its gone, thought its fixed, can't catch it for now

oa-s:
updated pr with this flow:
`
5. Tap Enable button
6. Automatically enable Goerli and Optimistic Goerli (and not needing to save manually)
`

hboon:
@oa-s looks great! Can you help modify the "Enable" button to "Enable and Connect" (or "Enable and Accept" or whatever is appropriate and fits), since it also accepts the new session?

hboon (approved):

PR #6508, opened 2023-03-13
[Refactoring] make enabled servers screen being able to reload automatically when enabled servers change

oa-s:


hboon:
@oa-s heads-up: this changes the functionality right? If the screen reload behavoir changes, then it's not a refactoring anymore

PR #6510, opened 2023-03-14
[Refactoring] Extract private key access + signing out. In the future, when a hardware wallet is used intead, this is replaced with prompting the user to sign with hardware wallet

hboon:
Follow up to: #6509

Extracts the part where we retrieve the private key and sign into `_signHash(_:,for:,prompt:)`. We do this so that in a future PR, `_signHash()` can check if it should read the private key and sign OR prompt for a hardware wallet to sign

There is also a new `_signHashes()`, but it works the same as `_sighHash()` except that it will read the private key once and use it multiple times to sign the hashes. And in the future, for hardware wallets, it will have to prompt the user to sign once for each has to sign (a worse UX, but it's just how hardware wallet signing works)

There'll be a follow up PR to make the calls async (await) and that should be it until we finally implement hardware wallet signing

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/KeyManagement/EtherKeystore.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/KeyManagement/EtherKeystore.swift
@@ -431,15 +431,14 @@ open class EtherKeystore: NSObject, Keystore {
         return signMessageData(prefix + message, for: account, prompt: prompt)
     }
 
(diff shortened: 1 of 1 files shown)


oa-s (approved):

hboon:
@oa-s please help take a lot to see if it breaks signing. Especially things like if we still +27 at the right place. I tried to minimize changes.

hboon:
@oa-s thanks!

PR #6519, opened 2023-03-14
Update transaction value with zero if gas price greater then balance,or all funds case

oa-s:


hboon (commented):
How do we reproduce this?

oa-s (commented):
I tried with hardcoding a value

PR #6524, opened 2023-03-15
[Refactoring] Extract extensions for TrustWalletCore into a new pod `AlphaWalletTrustWalletCoreExtensions`. To avoid circular dependencies for future hardware wallet support

hboon:
Had to duplicate a few extension functions in the pod to minimize changes. Added TODOs in the code base to clean those up


hboon:
@oa-s this too. Dependency for the other PR that was approved.

oa-s (approved):

PR #6531, opened 2023-03-16
support token uri rewriting

oa-s:
support token uri rewriting, some of returned token uris return uri with token id with zeros siffix to reach 64 characters, and api don't suport it such uris
such as `https://api.mintkudos.xyz/metadata/3ba0000000000000000000000000000000000000000000000000000000000000` it needs to be converted to 
`https://api.mintkudos.xyz/metadata/954`

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenUriMapSupportable.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenUriMapSupportable.swift
@@ -1,0 +1,80 @@
+//
+//  TokenUriMapSupportable.swift
+//  AlphaWalletFoundation
+//
+//  Created by Vladyslav Shepitko on 16.03.2023.
+//
+
+import Foundation
+import BigInt
+
+public protocol TokenUriMapSupportable {
+    func map(uri: URL) -> URL?
+}
+
+public struct TokenUriMapper: TokenUriMapSupportable {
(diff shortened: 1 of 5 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Types/TokenUriMapSupportable.swift
```diff
@@ -0,0 +1,72 @@
...
+
+protocol TokenUriMapSupportable {
+    func map(uri: URL) -> URL?
+}
+
+struct TokenUriMapper: TokenUriMapSupportable {
+    let hostMeppers: [TokenUriMapSupportable]
+
+    func map(uri: URL) -> URL? {
+        hostMeppers.compactMap { $0.map(uri: uri) }.first
```
hboon, line 19: It seems to work, but the logic here looks unintuitive and maybe slower? Can we rewrite it so it's something like this instead?:

```
hostMeppers.first { $0.match(uri: uri) }?.mappedUri
```

It finds the first match, rather than operate on all of them and gets the first

(and typo: `hostMeppers` -> `hostMappers` or `mappings`?)

hboon: And add a test case?

hboon (approved):

PR #6552, opened 2023-03-20
support gateway.pinata.cloud token uri mapping

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Helpers/JsonFromTokenUri.swift
```diff
@@ -35,7 +35,8 @@ final class JsonFromTokenUri {
             blockchainProvider: blockchainProvider,
             uriMapper: TokenUriMapper(hostMappers: [
                 HostBasedTokenUriMapper(host: "api.mintkudos.xyz"),
-                HostBasedTokenUriMapper(host: "api.walletads.io")
+                HostBasedTokenUriMapper(host: "api.walletads.io"),
```
hboon, line 38: How do we get the original URL that has to be converted? eg. `https://api.mintkudos.xyz/metadata/3ba0000000000000000000000000000000000000000000000000000000000000` ?

hboon (approved):

PR #6554, opened 2023-03-20
tweak switching rpc server when connection to wallet connect v1

oa-s:


hboon (commented):
What problem does this solve or improve on?

oa-s:
i found that after selecting another rpc server to connect, when connection wallet connect v1 server it stays unchanged on connection confirmation screeen, looks like a bug, so this pr fixes it.

hboon (approved):
Oh, it's the other one we talked about in a separate PR? The fix seems a little fragile though. (navigation controller/view controller navigation always seem fragile to me)

PR #6566, opened 2023-03-22
Update What's new to be more recent

hboon:


eviltofu:
I ran the master branch and created a wallet. The old WhatsNewListing showed up. Then I switched to this branch and ran the app again. The WhatsNewListing did not show up. 

https://github.com/AlphaWallet/alpha-wallet-ios/blob/024f73c912c187d734845969180d68b0cf117b69/AlphaWallet/WhatsNew/Coordinators/WhatsNewExperimentCoordinator.swift#L46-L49

When I ran this on the master branch, `hasRan` was false but was later set to true.
When I ran this on this branch, `hasRan` was true so the display of the WhatsNewListing was skipped.
I don't see anywhere else in the app where `hasRan` was reset.

@hboon

eviltofu:
https://github.com/AlphaWallet/alpha-wallet-ios/blob/024f73c912c187d734845969180d68b0cf117b69/AlphaWallet/WhatsNew/Coordinators/WhatsNewExperimentCoordinator.swift#L43-L45

When we start with watching or importing an existing wallet, `lastCreatedWalletTimestamp` somehow is `nil`.

hboon:
> When we start with watching or importing an existing wallet, lastCreatedWalletTimestamp somehow is nil.

Yes, from this comment, it looks it only protects against new wallets (not imports or watch) because we only prompt users to backup if it's a new wallet:

> //We don't want to show the user What's New right after a wallet is created (not imported) because this will crash with the UI which we used to prompt users to back up the new wallet

hboon:
> I don't see anywhere else in the app where hasRan was reset.

By design. We haven't gotten to do that yet. It would require us to refresh the this "What's new" list more diligently. Getting there!

PR #6601, opened 2023-03-27
[Refactoring] move more code to view appropriate view model

oa-s:
RP refactors transaction confirmation view model with actual class for each transaction type without bridging to each type, remove TransactionTypeViewModel enum, updates way to handle updates, removes TransactionConfigurator delegate. Adds displaying blockie image for recipient, displaying dapp icon and name for Dapp/WalletConnect transaction type


hboon (approved):
Too big; didn't check. But seems ok when I ran it.

> Adds displaying blockie image for recipient, displaying dapp icon and name for Dapp/WalletConnect transaction type

- [ ] Better to include before/after screenshots for these.

> [Refactoring] move more code to view appropriate view model #6601

- [ ] Also not a refactoring since it modifies the display. Would you help rename the commit message and PR title? Any chance to break it out into separate commits if it's not too much work?

- [ ] Need to fix the test suite though, it doesn't build anymore

oa-s:
| After PR | Before PR |
|----|----|
| Screenshot 2023-03-28 at 10 41 41 | Screenshot 2023-03-28 at 10 43 51 |
| Screenshot 2023-03-28 at 10 41 38 | - |

hboon:
@oa-s looks great!

PR #6605, opened 2023-03-28
Fix: Optimism Goerli should not use Etherscan key. Include a test based on heuristics to catch future errors

hboon:


hboon:
@eviltofu heads-up; see the new test case for a defensive check. We probably need to do more of these as we come across them

oa-s (approved):

PR #6610, opened 2023-03-28
Make enjinUserName and enjinUserPassword nil #6603

### LINKED ISSUE #6603: Make `enjinUserName` and `enjinUserPassword` nil in `Constants+Credentials.swift` and change code to disable when either of them are nil
- [x] Make codebase handle when `enjinUserName` and `enjinUserPassword` are nil
- [x] Make `enjinUserName` and `enjinUserPassword` nil in `Constants+Credentials.swift` (to avoid someone accessing the account. We still have access during development due to `.credentials`)

oa-s:
Closes #6603

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/NFT/NFTProvider.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/NFT/NFTProvider.swift
@@ -34,6 +34,13 @@ extension OpenSea: NftAssetImageProvider {
     }
 }
 
+extension Constants.Credentials {
+    static var enjinCredentials: EnjinCredentials? {
+        guard let email = Constants.Credentials.enjinUserName, let password = Constants.Credentials.enjinUserName else { return nil }
+        return (email: email, password: password)
+    }
+}
+
 public final class AlphaWalletNFTProvider: NFTProvider {
     private let openSea: OpenSea
     private let enjin: Enjin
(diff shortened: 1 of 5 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/NFT/NFTProvider.swift
```diff
@@ -34,6 +34,13 @@ extension OpenSea: NftAssetImageProvider {
     }
 }
 
+struct BaseEnjinCredentialsProvider: EnjinCredentialsProvider {
```
hboon, line 37: Do we need to introduce 2 more types — `BaseEnjinCredentialsProvider` and `EnjinCredentialsProvider`? It seems the existing `EnjinCredentials`  was enough right?

oa-s: added EnjinCredentialsProvider defined as a source of EnjinCredentials, so to pass a source and not just a raw value, we would be able to update code later to support using users enjin creds. maybe add some screen to enter creds

hboon: Not sure I understand; if we introduce UI for the users to enter them, the UI code can just pass ‘EnjinCredentials’ around. Why would it need an additional pair of types to do that? But maybe those additional types are helpful because an orchestrating “manager” type or coordinator for that UI flow can conform to them and pass the credentials after the user enter them — then let’s not get there so soon :)

It seems unlikely to go that way and it’s an easy refactor if we have to. But in the meantime, if we keep it simpler, we have 2 fewer types to manage. It all adds up across the codebase for things we might never need and is easy to add. Does that logic make sense?

PR #6612, opened 2023-03-28
[browser] move handling a keyboard state to view model

oa-s:


hboon (approved):

hboon:
@oa-s better to mention "browser" in the commit message

PR #6616, opened 2023-03-29
Suport `action=token1155tx` #6275

### LINKED ISSUE #6275: Support `action=token1155tx`

oa-s:
Closes #6275


oa-s:
still in progress, reducing pr size

oa-s:
should be good to review at this point

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/AlphaWalletTokensService.swift
```diff
@@ -12,21 +12,9 @@ import CombineExt
...
-    private let autoDetectTokensQueue: OperationQueue = {
-        let queue = OperationQueue()
-        queue.name = "Auto-detect Tokens"
-        queue.maxConcurrentOperationCount = 1
-        return queue
-    }()
     private let sessionsProvider: SessionsProvider
     private let analytics: AnalyticsLogger
-    private let tokensDataStore: TokensDataStore
+    public let tokensDataStore: TokensDataStore
```
hboon, line 17: Looks like this can still be private?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/TransactedTokensAutodetector.swift
```diff
@@ -0,0 +1,154 @@
...
+                    self?.handle(error: e)
+                }).mapToVoid()
+                .eraseToAnyPublisher()
+        }
+
+        private func fetchOperation(walletAddress: AlphaWallet.Address,
+                                    startBlock: Int?) -> AnyPublisher<UniqueNonEmptyContracts, PromiseError> {
+            switch tokenType {
+            case .erc1155:
+                return apiNetworking.erc1155TokenInteractions(walletAddress: walletAddress, startBlock: startBlock)
```
hboon, line 131: In `CovalentApiNetworking`, `erc1155TokenInteractions(walletAddress:startBlock:)` returns `.empty()`, but does `EtherscanCompatibleApiNetworking`'s implementation handle when not every chain supports it?

Also, does every chain that is `EtherscanCompatibleType.etherscan` actually support `action=token1155tx`?

How does the code know to fallback to events? Or do we always do both? If we do that, can we change it so that we prefer getting them from blockchain explorer APIs (i.e. Etherscan and friends) and if that isn't available, then we fallback to events? We could change this in a separate PR, since we already do both for `ERC20` and `ERC721` too right? But by doing both for ERC-1155, does it confuse our token ID-count bookkeeping?

oa-s: > How does the code know to fallback to events?

it works without fallback to events.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/AlphaWalletTokensService.swift
oa-s: will check, might be migration issue

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/TransactedTokensAutodetector.swift
oa-s: > In CovalentApiNetworking, erc1155TokenInteractions(walletAddress:startBlock:) returns .empty(), but does EtherscanCompatibleApiNetworking's implementation handle when not every chain supports it?

Covalent, as well as Oklink they don't have separate endpoint for retrieving such transactions, might be need to update with additionaly filter logic for each one

oa-s: > Also, does every chain that is EtherscanCompatibleType.etherscan actually support action=token1155tx?

haven't check for all of available in the app, but some of them work only when startBlock, endBlock provided, also need to handle this. I don't know is that acceptable if we would take end block as some number (100 for example to make that range, also with checking if api accepts such rangle)

hboon: I'm guessing that not every chain on Etherscan will support it. If that's the case, we'll have to hardcode the supported chains.

> but some of them work only when startBlock, endBlock provided, also need to handle this. I don't know is that acceptable if we would take end block as some number (100 for example to make that range, also with checking if api accepts such rangle)

Ouch. Yes I guess we'll have to paginate and figure out if the limits are the same for each chain

oa-s:
pr adds support of `token1155tx` method for etherscan provider
adds ability to retrieve erc1155 transactions
adds ability to retrieve erc1155 contract interactions

hboon (approved):

PR #6631, opened 2023-03-31
[RPC] Fix: blockpi has a limit for block range of 1024 for `eth_getLogs`

hboon:


hboon:
@justindg not sure if this affects Android?

PR #6635, opened 2023-03-31
set valid token balance when importing nonfungible token #4937

### LINKED ISSUE #4937: "Impossible To Build Configuration"
Refer https://github.com/AlphaWallet/alpha-wallet-ios/issues/4888#issuecomment-1175351309

@Hoanib would you help to describe how to reproduce this?

oa-s:
Fixes #4937

### DIFF
--- a/AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
+++ b/AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
@@ -229,11 +229,22 @@ class NewTokenViewController: UIViewController {
     }
 
     //int is 64 bits, if this proves not enough later we can convert to BigUInt
-    public func updateBalanceValue(_ balance: [String], tokenType: TokenType) {
+    public func updateBalanceValue(_ balance: NonFungibleBalance, tokenType: TokenType) {
         //TODO this happens to work for CryptoKitty now because of how isNonZeroBalance() is implemented. But should fix
-        let filteredTokens = balance.filter { isNonZeroBalance($0, tokenType: tokenType) }
-        viewModel.erc875TokenBalance = filteredTokens
-        balanceTextField.value = viewModel.erc875TokenBalanceAmount.description
+        let filteredBalance: NonFungibleBalance
+        switch balance {
+        case .assets(let values):
+            filteredBalance = .assets(values.filter { isNonZeroBalance($0.json, tokenType: tokenType) })
+        case .erc875(let values):
+            filteredBalance = .erc875(values.filter { isNonZeroBalance($0, tokenType: tokenType) })
+        case .erc721ForTickets(let values):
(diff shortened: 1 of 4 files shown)


hboon:
@oa-s Does this only fix it for ERC-1155? Do we have the same problem for ERC-721?

oa-s:
nope it should be for all non fungible types

AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
```diff
@@ -314,20 +325,28 @@ class NewTokenViewController: UIViewController {
...
         let decimals = Int(decimalsTextField.value) ?? 0
         guard let tokenType = self.tokenType else { return }
-        //TODO looks wrong to mention ERC875TokenBalance specifically
-        var balance: [String] = viewModel.erc875TokenBalance
+
+        var nonFungibleBalance: NonFungibleBalance
+        switch tokenType {
+        case .erc875:
+            if let balance = viewModel.nonFungibleBalance {
+                nonFungibleBalance = balance.rawValue.isEmpty ? .erc875(["0"]) : balance
```
hboon, line 333: Should this be `.erc875([])` or `.erc875([0])`? Ditto for 2 lines below this.

AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
```diff
@@ -314,20 +325,28 @@ class NewTokenViewController: UIViewController {
...
+            } else {
+                nonFungibleBalance = .erc875(["0"])
+            }
+        case .nativeCryptocurrency, .erc20, .erc721, .erc1155:
+            nonFungibleBalance = viewModel.nonFungibleBalance ?? .balance([])
```
hboon, line 338: This one's an impossible path right? Shall we do an `assertImpossibleCodePath(message:)`? 

(we'll change the implementation for `assertImpossibleCodePath(message:)` and `assertImpossibleCodePath()` to use `preconditionFailure(_:)` instead of `assert(Bool,String)` in the future)

hboon (approved):
Looks OK. Assume you tested. A little hard to predict what would happen without stepping through it (which I didn't), unfortunately.

AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
oa-s: actually i don't know, previous implementation setting initial value as "0", if array is empty

AlphaWallet/Tokens/ViewControllers/NewTokenViewController.swift
oa-s: i guess for that we need to tweak ErcToken model, because nonFungible balance isn't optional, and in any case balance need to be calculated

PR #6645, opened 2023-04-04
Modified TokenGroupIdentifier to detect spam, added SpamTokenService to autohide spam tokens

### LINKED ISSUE #6445: [tokens.json] Auto mark tokens as hidden when a token is first discovered (and going to be added to the datastore) if they are in the "Spam Group"

eviltofu:
Closes #6445 

# Tests

## Stand alone

1. Delete the app 
2. Start the app
3. Tap "I already have a wallet"
4. Tap "Track Wallet Address"
5. Type in "hweeboon.eth" in the Ethereum Address field
6. Tap "Watch Wallet"
7. Wait 2 minutes
8. Tap ellipses in upper right corner
9. Tap "Add / Hide Tokens"
10. Scroll down to Hidden Tokens
11. Expects there to be up to 8 tokens

## Switch wallets

1. Delete the app 
2. Start the app
3. Tap "I already have a wallet"
4. Tap "Track Wallet Address"
5. Type in "hweeboon.eth" in the Ethereum Address field
6. Tap "Watch Wallet"
7. Tap Blockie in upper left corner
8. Tap + in upper right corner
9. Tap "Create Wallet"
10. Wait 2 minutes
11. Tap Blockie in upper left corner
12. Switch to "hweeboon.eth" wallet
13. Tap ellipses in upper right corner
14. Tap "Add / Hide Tokens"
15. Scroll down to Hidden Tokens
16. Expects there to be up to 8 tokens

# Code

## TokenGroupIdentifier

This is moved up from ActiveWalletCoordinator to AppCoordinator so that there is only one copy of it in the app instead of one copy for each wallet.

Functions to detect spam are added.


AlphaWallet/ActiveWalletCoordinator.swift
```diff
@@ -28,11 +28,13 @@ class ActiveWalletCoordinator: NSObject, Coordinator {
     private let activitiesPipeLine: ActivitiesPipeLine
     private let sessionsProvider: SessionsProvider
     private let currencyService: CurrencyService
+    private let tokenGroupIdentifier: TokenGroupIdentifierProtocol
```
hboon, line 31: Not this PR, but missed this one a while back. But better to rename `TokenGroupIdentifierProtocol` to `TokenGroupMakerProtocol`. The "identifier" suggests IDs. Ambiguous. And `identifier(fromFileName:)` too. `identify(token:)` isn't so bad.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
```diff
@@ -162,6 +168,10 @@ public class TokenGroupIdentifier: TokenGroupIdentifierProtocol {
     public func hasContract(address: String, chainID: Int) -> Bool {
         return decodedTokenEntries["\(address):\(chainID)"] != nil
     }
+
+    public func isSpam(address: String, chainID: Int) -> Bool {
```
hboon, line 172: I wondered if the function signature should be this instead:

```
func isSpam(address: AlphaWallet.Address, chain: RPCServer) -> Bool
```

On one hand, this is stronger typing (and we can change the implementation so the key is a new type instead of appending lowercased strings).

On the other hand, this introduces a dependency on 2 pods into `TokenGroupIdentifier.swift`. The dependency is not an issue now, but if/when we extract it into a pod, it would be. For example when we extracted the `ENS` pod https://github.com/AlphaWallet/alpha-wallet-ios/blob/eefc84d63bd800a4c6690a3750d4f0dcaede3a6d/modules/AlphaWalletENS/AlphaWalletENS/ENS.swift#L22, we replace RPCServer with a `ChainId = Int` typealias.

I guess we can leave it for now. Doesn't matter in the grand scheme of things and is easy to change

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
```diff
@@ -162,6 +168,10 @@ public class TokenGroupIdentifier: TokenGroupIdentifierProtocol {
...
+        return spamTokenEntries.contains("\(address):\(chainID)".lowercased())
```
hboon, line 173: Can you extract this key generation? It's used in a few place. Looks like good as wrapped in an `enum functional {}`

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
```diff
@@ -130,11 +130,13 @@ public protocol TokenGroupIdentifierProtocol {
     static func identifier(fromFileName: String) -> TokenGroupIdentifierProtocol?
     func identify(token: TokenGroupIdentifiable) -> TokenGroup
     func hasContract(address: String, chainID: Int) -> Bool
+    func isSpam(address: String, chainID: Int) -> Bool
 }
 
 public class TokenGroupIdentifier: TokenGroupIdentifierProtocol {
 
     private var decodedTokenEntries: TokenGroupDictionary = TokenGroupDictionary()
+    private var spamTokenEntries: Set<String> = Set<String>()
```
hboon, line 139: Both `var` can be `let` right?

AlphaWallet/Common/Services/SpamTokenService.swift
```diff
@@ -0,0 +1,48 @@
...
+        self.tokensService = tokensService
+    }
+
+    func startMonitoring() {
+        stopMonitoring()
+        tokensService.addedTokensPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { tokens in
+                tokens.forEach { token in
+                    if self.tokenGroupIdentifier.isSpam(address: token.contractAddress.eip55String, chainID: token.server.chainID) {
```
hboon, line 31: Seems the abstraction is done better if instead of:

```
func startMonitoring() {
    stopMonitoring()
    tokensService.addedTokensPublisher
        .receive(on: DispatchQueue.main)
        .sink { tokens in
            tokens.forEach { token in
                if self.tokenGroupIdentifier.isSpam(address: token.contractAddress.eip55String, chainID: token.server.chainID) {
                    self.performSpamHandler(token: token)
                }
            }
        }.store(in: &cancellableSet)
}
```

We do:

```
func startMonitoring() {
    stopMonitoring()
    tokensService.addedTokensPublisher
        .receive(on: DispatchQueue.main)
        .sink { tokens in
            self.performSpamHandler(tokens: tokens)
        }.store(in: &cancellableSet)
```

hboon (approved):
Looks good. Assume you tested.

- [ ] Follow up to see if this needs cleaning up https://github.com/AlphaWallet/alpha-wallet-ios/blob/52081f6e986dbd33c5c76d66cc42907585d95ffb/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Lists/SpamTokens.swift#L5

@eviltofu trying again to provide some feedback and pre-approving. Don't merge until you are done :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
eviltofu: This could also be combined with the one where a "-" is used instead of ":". Like in `ContractAddressObject. generatePrimaryKey`. Maybe standardize the usage.
```swift
static func generatePrimaryKey(fromContract contract: AlphaWallet.Address, server: RPCServer) -> String {
        return "\(contract.eip55String)-\(server.chainID)"
    }
```

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
eviltofu:
```swift
    private init(decodedTokenEntries: TokenGroupDictionary) {
        decodedTokenEntries.forEach { key, value in
            if value == TokenGroup.spam {
                spamTokenEntries.insert(key.lowercased())
            } else {
                self.decodedTokenEntries[key] = value
            }
        }
    }
```
Not without changing this code. Do you want me to?

AlphaWallet/Common/Services/SpamTokenService.swift
eviltofu: What about
```swift
func startMonitoring() {
    stopMonitoring()
    tokensService.addedTokensPublisher
        .receive(on: DispatchQueue.main)
        .sink { tokens in
            self.performSpamHandler(tokens: tokens.filter{ self.tokenGroupIdentifier.isSpam(address $0.contractAddress.eip55String, chainID: $0.server.chainID} )
        }.store(in: &cancellableSet)
```

eviltofu:
> Looks good. Assume you tested.

I did not do the test where two wallets have the same spam tokens. Only the two tests mentions in the introduction.

eviltofu:
> * [ ]  Follow up to see if this needs cleaning up https://github.com/AlphaWallet/alpha-wallet-ios/blob/52081f6e986dbd33c5c76d66cc42907585d95ffb/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Lists/SpamTokens.swift#L5

This is not used and should be removed.

AlphaWallet/ActiveWalletCoordinator.swift
eviltofu: Maybe just `TokenGroupProtocol`? We are not making any groups, just identifying the groups.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
hboon: This file internally first? Doesn't need to extract for the other one seems it's just a coincidence.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
hboon: Yeah, but you can do it with a follow up PR to keep this PR smaller if you like. (`skip-review` if you are confident)

AlphaWallet/Common/Services/SpamTokenService.swift
hboon: This one's better because the level of abstraction in `startMonitoring()` is more consistent:

```
self.performSpamHandler(tokens: tokens)
```

Sure, renaming sounds good.

hboon:
> > * [ ]   Follow up to see if this needs cleaning up https://github.com/AlphaWallet/alpha-wallet-ios/blob/52081f6e986dbd33c5c76d66cc42907585d95ffb/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Lists/SpamTokens.swift#L5
> 
> ~This is not used and should be removed.~ This is used in `TickerIdsFetcherImpl` and looks like we might have duplicate functionality. I can't find which coordinator this is created under...

In `TickerIdsFetcherImpl`. Think it's only meant to be temporarily, can probably remove it and add the contracts into `tokens.json` if you haven't and if they are missing.

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/JSON/TokenGroupIdentifier.swift
eviltofu: but both represent the same thing

AlphaWallet/Common/Services/SpamTokenService.swift
```diff
@@ -0,0 +1,48 @@
...
+/// When a spam token is detected, this service will hide it.
+
+import Foundation
+import AlphaWalletFoundation
+import Combine
+
+class SpamTokenService {
+
+    private let tokenGroupIdentifier: TokenGroupIdentifierProtocol
+    private var cancellableSet: Set<AnyCancellable> = []
```
oa-s, line 17: replace with `private var cancellable = Set<AnyCancellable>()`, most of subscriptions in the app store in `cancellable`

AlphaWallet/Common/Services/SpamTokenService.swift
```diff
@@ -0,0 +1,48 @@
...
+                tokens.forEach { token in
+                    if self.tokenGroupIdentifier.isSpam(address: token.contractAddress.eip55String, chainID: token.server.chainID) {
+                        self.performSpamHandler(token: token)
+                    }
+                }
+            }.store(in: &cancellableSet)
+    }
+
+    private func stopMonitoring() {
+        cancellableSet.forEach { cancellable in
```
oa-s, line 39: use `cancellable.cancellAll()` to cancel all subscriptions in the set

eviltofu:
@hboon updated

hboon:
@oa-s in case you are still looking at it.

hboon (approved):
@eviltofu go for it

PR #6648, opened 2023-04-04
remove GasPriceObject from list of objectTypes for MigrationInitializerForOneChainPerDatabase

### LINKED ISSUE #6644: Test suite fails at 6d3022aae27c0727c0d3ccf7ea13049e83448032
ActiveWalletViewTests.testShowTabAlphwaWalletWallet() fails if running the entire suite

The same test case works if ran alone

oa-s:
Closes #6644


hboon (approved):
Is it better to include `GasPriceObject.self` instead of removing `TransactionObject.self`?

oa-s:
`TransactionObject ` it haven't existed when MigrationInitializerForOneChainPerDatabase was intruduced, so it shouldn't cause any issue, removing it

hboon:
Ok

PR #6657, opened 2023-04-05
Added logic to use an internal SpamToken image to TokenImageFetcherImpl instead of fetching it over the network

### LINKED ISSUE #6446: [tokens.json] Automatically apply the Spam token icon instead of the one from sources like `iconassets` repo

eviltofu:
Closes #6446

Simulator Screenshot - iPhone 14 Pro - 2023-04-11 at 09 06 17 copy


hboon:
@eviltofu I just realized that we can't remove the spam token icon files from the `iconassets` repo rightaway since we need to give it some time after both the Android and iOS app has been live. Can you help create/move an issue into the `iconassets` repo to remove those files and assign to me, referring back to this comment so we remember to finish it up?

eviltofu:
@hboon 

So I schedule a task to remove the spam token icons files 1 month later?

Also, this process of adding to the `adjustments.js` file after checking `iconassets` needs to be streamlined. Maybe we can just have a `spam.json` file and then from that generate the `adjustments.js` file and remove the spam logo from `iconassets`?

hboon:
> So I schedule a task to remove the spam token icons files 1 month later?

Yes, but can you do it this way: Set up a new test class that does this:

```
class Foo_Some_Name_Reminders: XCTestCase {
    func testRemoveSpamTokenIconFiles() {
        //Fails with a TODO to create an issue with "some title" on and after 20230601
    }
}
```

Thought I wrote about this approach in the Android repo for some other issue, but I forgot where

hboon:
> Also, this process of adding to the adjustments.js file after checking iconassets needs to be streamlined. Maybe we can just have a spam.json file and then from that generate the adjustments.js file and remove the spam logo from iconassets? Generate a script file from spam.json that checks the iconassets repo.

Improving the process is good. `adjustments.json` currently only holds contracts for the `Spam` group though. Maybe wait a bit to see what we do with the file before we modify the code/process? Seems unnecessary now.

> Also instead of removing the logo.png file, maybe just rename it to spam.png to hide it from the app? Otherwise if we make a mistake in classifying a token as spam and later discovering it was an error, we would need to go find and download the token logo icon again.

It wouldn't be because it's the same icon file contents (and it's still in git any way)

eviltofu:
@oa-s still reviewing?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokenImageFetcher.swift
```diff
@@ -130,6 +136,10 @@ public class TokenImageFetcherImpl: TokenImageFetcher {
                                           blockChainNameColor: UIColor,
                                           serverIconImage: UIImage?) -> TokenImage? {
 
+        if tokenGroupsIdentifier.isSpam(address: contractAddress.eip55String, chainID: server.chainID) {
+            return .init(image: .image(.loaded(image: spamImage)), isFinal: true, overlayServerIcon: nil)
```
hboon, line 140: Didn't test. But I think `overlayServerIcon: ` still has to be either `nil` or `not-nil` using the `switch-case` a few lines later. That's how the current hardcoded spam (or any token image) loaded from the `iconassets` repo are displayed, right?

hboon (changes requested):

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokenImageFetcher.swift
eviltofu: Updated. Please check @hboon

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokenImageFetcher.swift
```diff
@@ -132,14 +138,18 @@ public class TokenImageFetcherImpl: TokenImageFetcher {
...
             }
         case .erc20, .erc875, .erc721, .erc721ForTickets, .erc1155:
-            if let img = tokenImage {
-                return .init(image: .image(.loaded(image: img)), isFinal: true, overlayServerIcon: staticOverlayIcon)
+            if let img = iconImageForContractAndChainID(image: tokenImage, address: contractAddress.eip55String, chainID: server.chainID) {
+                return TokenImage(image: .image(.loaded(image: img)), isFinal: true, overlayServerIcon: staticOverlayIcon)
             }
         }
+        
+        if tokenGroupsIdentifier.isSpam(address: contractAddress.eip55String, chainID: server.chainID) {
```
hboon, line 150: Seems like this will never be true because it was checked in the `switch` within the `iconIamgeForContractAndhainID(image:address:chain:)` call, so the `if`-block can be removed.

hboon (changes requested):

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokenImageFetcher.swift
eviltofu: No, it will fail if the tokenImage or serverIconImage does not exist which could happen if the user adds in a custom network

eviltofu: Removed.

hboon (approved):

PR #6658, opened 2023-04-05
reduce eth_log calls from event log send event logs for activities #6540

### LINKED ISSUE #6540: We seem to be making much more calls to Infura with v3.62

oa-s:
Closes #6540

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/EventSource.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/TokenScriptClient/EventSource.swift
@@ -188,8 +188,6 @@ final class EventSource {
 
                     workers[request.token.contractAddress] = worker
                     self.workers[request.token.contractAddress] = worker
-
-                    worker.send(request: request)
                 }
             }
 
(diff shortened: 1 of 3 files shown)


modules/AlphaWalletWeb3/AlphaWalletWeb3/Promises/Promise+Batching.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 import PromiseKit
 
 public class JSONRPCrequestDispatcher {
-    public var MAX_WAIT_TIME: TimeInterval = 0.1
+    public var MAX_WAIT_TIME: TimeInterval = 0.5
```
hboon, line 13: Have you try measuring/logging to get a feel how many requests this cuts down to?

hboon: Also: https://github.com/AlphaWallet/alpha-wallet-ios/issues/6540#issuecomment-1498346374

oa-s:
> 
> Took a quick look. Can you explain how it helps?

Yes, this pr fixes force updating event logs, current implementation has two types of updates, force, and .waitForCurrent, on initial call we make force update, as well as token script changes. But when tokens balance charges we make wait for current update. So current implementation have bug when we always send force update. This pr fixes that

hboon (approved):
Thanks

PR #6660, opened 2023-04-05
log failure respponse for api provider #6602

### LINKED ISSUE #6602: Catch errors with API keys sooner/better
We mixed up a few keys and only caught some (what else did we miss?) by accident. eg. #6596

This helps: #6597

But our monitor doesn't catch them, and how would it be sync with the app codebase, so it seems like it should be in the test suite

oa-s:
Closes #6602

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractInteractions.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractInteractions.swift
@@ -21,10 +21,22 @@ class GetContractInteractions {
         let request = GetContractList(walletAddress: walletAddress, server: server, startBlock: startBlock, erc20: erc20)
         return networkService
             .dataTaskPublisher(request)
+            .handleEvents(receiveOutput: { self.log(response: $0) })
             .receive(on: DispatchQueue.global())
             .tryMap { UniqueNonEmptyContracts(json: try JSON(data: $0.data)) }
             .mapError { PromiseError.some(error: $0) }
             .eraseToAnyPublisher()
+    }
+
+    private func log(response: URLRequest.Response) {
+        switch URLRequest.validate(statusCode: 200..<300, response: response.response) {
+        case .failure:
+            let json = try? JSON(response.data)
+            let message = json.flatMap { $0["message"].stringValue } ?? ""
+            infoLog("[API] request failure with status code: \(response.response.statusCode), message: \(message)")
+        case .success:
+            break
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


oa-s:
@hboon let me know if i missed something

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractInteractions.swift
```diff
@@ -21,11 +21,23 @@ class GetContractInteractions {
         let request = GetContractList(walletAddress: walletAddress, server: server, startBlock: startBlock, erc20: erc20)
         return networkService
             .dataTaskPublisher(request)
+            .handleEvents(receiveOutput: { self.log(response: $0) })
```
hboon, line 24: Can shorten it a little to:

```
.handleEvents(receiveOutput: self.log)
```

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractInteractions.swift
```diff
@@ -21,11 +21,23 @@ class GetContractInteractions {
...
             .tryMap { UniqueNonEmptyContracts(json: try JSON(data: $0.data)) }
             .mapError { PromiseError.some(error: $0) }
             .eraseToAnyPublisher()
     }
+
+    private func log(response: URLRequest.Response) {
+        switch URLRequest.validate(statusCode: 200..<300, response: response.response) {
+        case .failure:
+            let json = try? JSON(response.data)
+            let message = json.flatMap { $0["message"].stringValue } ?? ""
```
hboon, line 35: Maybe this is slightly better?

```
let message: String = json?["message"].stringValue ?? ""
```

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/ApiNetworking.swift
```diff
@@ -36,3 +38,16 @@ public protocol ApiNetworking {
                                          pagination: TransactionsPagination,
                                          sortOrder: GetTransactions.SortOrder?) -> AnyPublisher<TransactionsResponse<TransactionInstance>, PromiseError>
 }
+
+extension ApiNetworking {
+    func log(response: URLRequest.Response) {
```
hboon, line 43: Seems definition for `log(response:) is duplicated a few times. Can we extract them?

hboon (approved):
Assume you tested :)

modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/GetContractInteractions.swift
hboon: @oa-s not super important, but seems not included. In case you pushed the wrong commit…

PR #6661, opened 2023-04-05
Modifications to project to allow compilation under Xcode 14.3

eviltofu:
Updated KingFisher to version 7.6.2. Added missing imports Xcode 14.3 was complaining about.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -10,7 +10,7 @@ target 'AlphaWallet' do
   pod 'StatefulViewController'
   pod 'QRCodeReaderViewController', :git=>'https://github.com/AlphaWallet/QRCodeReaderViewController.git', :commit=>'30d1a2a7d167d0d207ae0ae3a4d81bcf473d7a65'
   pod 'KeychainSwift', :git=>'https://github.com/AlphaWallet/keychain-swift.git', :commit=> 'b797d40a9d08ec509db4335140cf2259b226e6a2'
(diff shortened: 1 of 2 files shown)


eviltofu:
`make clean` and `make test` might have issues.

AlphaWallet/Common/Services/MediaContentNetworking.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
 import Alamofire
 import AlphaWalletFoundation
 import Combine
+import AlphaWalletOpenSea
```
hboon, line 12: I was curious so I tried. Doesn't seem like the 2 `import` statements are needed for `14.3`. Can you verify?

hboon (approved):

hboon:
Seems OK to me otherwise

AlphaWallet/Common/Services/MediaContentNetworking.swift
eviltofu: You can compile it but it gives warnings. I figured it would be better to just add in these missing imports. @hboon

hboon: Can you see why it needs `AlphaWalletOpenSea` in particular? Seems unnecessary given the purpose, so would like to reduce dependency on it

> warnings

OK, let's remove it and fix it separately/later so we can merge this. The warning would remind us.

PR #6663, opened 2023-04-06
Added `testReminderToRemoveSpamTokenIconFilesFromIconAssetsRepo` to schedule removal of spam token icon from iconassets repository

eviltofu:
When date is past, we get this output:
```text
ScheduleSpamTokenIconRemovalFromIconAssetsTest
    ✖ testDueDate, XCTAssertTrue failed - TODO:- Spam Token Icon Removal task: https://github.com/AlphaWallet/iconassets/issues/30
```

### DIFF
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -49,6 +49,7 @@
 		02D7F8CE27D8AAC900CA1140 /* VerticalButtonsBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D7F8CD27D8AAC900CA1140 /* VerticalButtonsBar.swift */; };
 		02D8BF8B277D570900EEE8E9 /* SaveCustomRpcManualEntryViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D8BF8A277D570900EEE8E9 /* SaveCustomRpcManualEntryViewController.swift */; };
 		02D8BF8D277D572600EEE8E9 /* SaveCustomRpcManualEntryView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D8BF8C277D572600EEE8E9 /* SaveCustomRpcManualEntryView.swift */; };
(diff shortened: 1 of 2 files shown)


hboon (approved):
Didn't check the logic, but can you change it so that the test class is generic for such scheduled TODOs and the function name to reflect the purpose instead?

hboon:
@eviltofu also notice the PR title is truncated.

eviltofu:
> Didn't check the logic, but can you change it so that the test class is generic for such scheduled TODOs and the function name to reflect the purpose instead?

@hboon `testDueDate` isn't descriptive enough? What do you suggest? `testDueDateElapsed`?

hboon:
@eviltofu Like that (and a few more):

* Add class/header — `//This class abuses the test suite to generate reminders when TODOs are due`
* Remove `ScheduledToDoTest`
* Rename `SpamTokenIconRemovalFromIconAssetsTest` -> `ScheduledToDoTest`
* Remove `setUpWithError()` and `tearDownWithError()`
* Rename `testDueDateElapsed()` to `testReminderToRemoveSpamTokenIconFilesFromIconAssetsRepo()`

So we only have 1 new class which is for all such scheduled TODOs and every test function represents a TODO (or maybe in the future a few related TODOs)

eviltofu:
> @eviltofu Like that (and a few more):
> 
> * Add class/header — `//This class abuses the test suite to generate reminders when TODOs are due`
> * Remove `ScheduledToDoTest`

I think we should keep it in case we have more than one ToDoTest.

> * Rename `SpamTokenIconRemovalFromIconAssetsTest` -> `ScheduledToDoTest`
> * Remove `setUpWithError()` and `tearDownWithError()`
> * Rename `testDueDateElapsed()` to `testReminderToRemoveSpamTokenIconFilesFromIconAssetsRepo()`
> 
> So we only have 1 new class which is for all such scheduled TODOs and every test function represents a TODO (or maybe in the future a few related TODOs)

Oh, you wanted one test for all the Todos. I was thinking of one test for each Todo.

hboon (approved):
Yeah, less lines of code where reasonable.

PR #6664, opened 2023-04-06
remove extra sign from percentageChange24h #6662

### LINKED ISSUE #6662: Extra $ sign in percentage figure
Should the **$** in front of the percentage not be there?
error copy

oa-s:
Fixes #6662


AlphaWallet/Accounts/ViewModels/WalletSummaryViewModel.swift
```diff
@@ -77,13 +77,12 @@ extension WalletSummaryViewModel {
 
             let changePercentage: String = {
                 guard let changePercentage = summary?.changePercentage else { return "-" }
-                let formatter = NumberFormatter.priceChange(currency: changePercentage.currency)
```
hboon, line 80: Maybe replace the `let` variable with `NumberFormatter.percent` instead? Few lines and better

hboon (approved):

PR #6686, opened 2023-04-13
Initial SwiftFormat tests

eviltofu:
Initial SwiftFormat testing.

1. [sortedImports], removed [trailing spaces] (empty lines) image 
2. [trailing comma] image 
3. [indent] image

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,6 +1,11 @@
 #Disable all and explicitly enable them to avoid unintentionally activating new rules in newer versions of SwiftFormat
 --disable all
 
+--enable indent
+--indent 4
+
+--enable typeSugar
+
 #TODO enable
 #--enable redundantVoidReturnType
 #--closurevoid remove
@@ -22,7 +27,7 @@
 #--enable enumNamespaces
 #--enumnamespaces always
 
---enable genericExtensions
+#--enable genericExtensions
 
 #TODO enable
 #--enable indent
@@ -33,13 +38,13 @@
 #--tabwidth unspecified
 #--xcodeindentation disabled
 
---enable linebreaks
---linebreaks lf
+#--enable linebreaks
+#--linebreaks lf
 
---disable wrapSingleLineComments
+#--disable wrapSingleLineComments
 
 #TODO enable
-#--enable modifierorder 
+#--enable modifierorder
 #--modifierorder private,lazy
 
 #TODO enable
@@ -60,8 +65,8 @@
 #--enable typeSugar
 #--shortoptionals always
 
---enable unusedArguments
---stripunusedargs closure-only
+#--enable unusedArguments
+#--stripunusedargs closure-only
 
 #TODO enable
 #--enable trailingSpace
@@ -75,14 +80,14 @@
 #--enable void
 #--voidtype void
 
---enable yodaConditions
---yodaswap always
+#--enable yodaConditions
+#--yodaswap always
 
 --disable spaceInsideComments
 
 #Files
 
---exclude build,vendor
---exclude Pods,Generated
+#--exclude build,vendor
+#--exclude Pods,Generated
 #Excluded because of error while processing
---exclude AlphaWallet/Browser/ViewModel/Dapps.swift
+#--exclude AlphaWallet/Browser/ViewModel/Dapps.swift
--- a/AlphaWallet.xcodeproj/project.pbxproj
+++ b/AlphaWallet.xcodeproj/project.pbxproj
@@ -4494,7 +4494,7 @@
 				C83F787024A1C1DB00783B67 /* Check correct credentials used for API keys */,
 				C887FD5D280806C20030E4BA /* Check correct GoogleService-Info.plist (for Crashlytics) */,
 				29336FE81F6B24FF005E3BFC /* R.Swift */,
-				87AF1609290A6D2A001C63D7 /* Swift Format */,
+				028E841529ECED780054EA02 /* Swift Format */,
 				2912CCF11F6A830700C6CBE3 /* Sources */,
 				2912CCF21F6A830700C6CBE3 /* Frameworks */,
 				2912CCF31F6A830700C6CBE3 /* Resources */,
@@ -4728,6 +4728,27 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		028E841529ECED780054EA02 /* Swift Format */ = {
+			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputFileListPaths = (
+				"$(SRCROOT)/resources/SwiftFormatFiles.xcfilelist",
+			);
+			inputPaths = (
+			);
+			name = "Swift Format";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [ \"${CONFIGURATION}\" = \"Debug\" ]; then\n    files_list = \"\"\n    while read F; do\n        files_list+=\"${F} \"\n    done < $SCRIPT_INPUT_FILE_LIST_0\n    \"${PODS_ROOT}/SwiftFormat/CommandLineTool/swiftformat\" $files_list\nfi\n";
+			showEnvVarsInLog = 0;
+		};
 		29282B511F7629DB0067F88D /* Run SwiftLint */ = {
 			isa = PBXShellScriptBuildPhase;
 			alwaysOutOfDate = 1;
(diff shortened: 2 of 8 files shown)


.swiftformat
```diff
@@ -1,5 +1,20 @@
 #Disable all and explicitly enable them to avoid unintentionally activating new rules in newer versions of SwiftFormat
---disable all
+#--disable all
+
+--rules \
```
hboon, line 4: @eviltofu shucks, so many files and include rule changes. Can you help select just a few files to illustrate what's changed and force push, so I don't have go through 136 files?

AlphaWallet.xcodeproj/project.pbxproj
```diff
@@ -4813,7 +4813,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [ \"${CONFIGURATION}\" = \"Debug\" ]; then\n    \"${PODS_ROOT}/SwiftFormat/CommandLineTool/swiftformat\" \"$SRCROOT\"\nfi\n";
+			shellScript = "if [ \"${CONFIGURATION}\" = \"Debug\" ]; then\n    \"${PODS_ROOT}/SwiftFormat/CommandLineTool/swiftformat\" \"$SRCROOT/AlphaWalletTests\"\nfi\n";
```
hboon, line 4816: And can you change ths to the script approach like we just discussed in the issue?

eviltofu: You want me to list each test file or can I just include the whole directory?

hboon: Each file; for now at least, I think we'll only switch to directories when we cover all the files in a directory. In this case since we are going to just select a few for the PR, definitely list the files.

.swiftformat
hboon: To be clear, I mean like, only 5 `.swift` files instead of ~136

AlphaWallet.xcodeproj/project.pbxproj
eviltofu: I've used the Xcode file lists facility. All the files that are to be SwiftFormatted are in the file `SwiftFormatFiles.xcfilelist`

.swiftformat
```diff
@@ -1,5 +1,20 @@
...
+	indent, \
+	duplicateImports, \
+	sortedImports, \
+	spaceInsideBraces, \
+	spaceInsideBrackets, \
+	spaceInsideComments, \
+	spaceInsideGenerics, \
+	spaceInsideParens, \
+	trailingCommas, \
```
hboon, line 13: Would you remove `trailingCommas`? Seems to generate an unnecessary diff when we reformat.

hboon (changes requested):
Thanks. Added checkboxes below so it's clearer what are the TODOs:

I tried a few random files to see what it does with indentation with just this to reduce the diff:

> --rules \
> 	indent, \
> 	typeSugar, \
> 	
> --indent 4

- [x] ^ Let's just introduce these first.

and this stood out:

Changed this (`ENS.swift`):

```
}.tryMap { ensName, resolvedAddress -> String in
    //snipped
}.mapError { e in SmartContractError.embeded(e) }
.eraseToAnyPublisher()
```

```
}.tryMap { ensName, resolvedAddress -> String in
    //snipped
}.mapError { e in SmartContractError.embeded(e) }
    .eraseToAnyPublisher()
```

- [ ] The original seems better as the indentation reflects the call targets better. Any way to make it so?

This is a little like the indentations here having meaning, tying the `doSomething()` with the `if`

```
if something {
    doSomething()
}

if something
    doSomething
```

vs. this which suggests (wrongly) that `doSomething` is not part of the `if`:

```
if something
doSomething
```

---

An example that is OK (`RPCServer.swift`):

```
case .custom(let custom):
    return custom.explorerEndpoint
            .flatMap { URL(string: $0) }
            .flatMap { $0.appendingPathComponent("api").absoluteString }
```
1G
to:

```
case .custom(let custom):
    return custom.explorerEndpoint
        .flatMap { URL(string: $0) }
        .flatMap { $0.appendingPathComponent("api").absoluteString }
```

The indentation amount doesn't matter (let's not get to 1-space indentation!) as much as whether it is indented or not, so this is fine.

Could you help to include these 2 files too and amend the PR title/commit message since they are not files in the test suite? I'm sorry if one of the other files in the test suite would already exhibit this behavior for indentation, feel free to use those instead if so.

eviltofu:
> Thanks. Added checkboxes below so it's clearer what are the TODOs:
> 
> I tried a few random files to see what it does with indentation with just this to reduce the diff:
> 
> > --rules 
> > indent, 
> > typeSugar, \
> > --indent 4
> 
> * [ ]  ^ Let's just introduce these first.

What does this mean? I remove the rest of the rules?

hboon:
> What does this mean? I remove the rest of the rules?

Yes. Indentation seems the most disruptive, lets focus on that one first.

eviltofu:
> Thanks. Added checkboxes below so it's clearer what are the TODOs:
> 
> I tried a few random files to see what it does with indentation with just this to reduce the diff:
> 
> > --rules 
> > indent, 
> > typeSugar, \
> > --indent 4
> 
> * [ ]  ^ Let's just introduce these first.
> 
> and this stood out:
> 
> Changed this (`ENS.swift`):
> 
> ```
> }.tryMap { ensName, resolvedAddress -> String in
>     //snipped
> }.mapError { e in SmartContractError.embeded(e) }
> .eraseToAnyPublisher()
> ```
> 
>  ```
> }.tryMap { ensName, resolvedAddress -> String in
>     //snipped
> }.mapError { e in SmartContractError.embeded(e) }
>     .eraseToAnyPublisher()
> ```
If you place the closing brace on the next line, it formats correctly.
```swift
}.mapError { e in SmartContractError.embeded(e)
}.eraseToAnyPublisher()
```

.swiftformat
```diff
@@ -1,5 +1,20 @@
 #Disable all and explicitly enable them to avoid unintentionally activating new rules in newer versions of SwiftFormat
---disable all
+#--disable all
```
hboon, line 2: We want this to stay uncommented, reason described in the comment

hboon:
> If you place the closing brace on the next line, it formats correctly.

Ah, this looks odd:

```
}.mapError { e in SmartContractError.embeded(e)
}.eraseToAnyPublisher()
```

should write like this instead, which indents with the correct level:

```
}.mapError { e in
    SmartContractError.embeded(e)
}.eraseToAnyPublisher()
```

But yes, looks like the formatting rule works if formatted this way. Good then.

Can you help update `.swiftformat`, removing the other rules that was added (eg. `duplicateImports`), including `--disable all` at the top again and see if we missed anything similar to that indentation issue above? Then we are good to go with adding a few files in the PR.

.swiftformat
eviltofu: the `--rules` option only allows the [rules](https://github.com/nicklockwood/SwiftFormat#rules) you specify, so this is redundent?
```
And then individually enable just the rules you want. Alternatively, use the--rules argument to only enable the rules you specify:

--rules indent,linebreaks
```

eviltofu:
> ```
> }.mapError { e in
>     SmartContractError.embeded(e)
> }.eraseToAnyPublisher()
> ```
> 
> But yes, looks like the formatting rule works if formatted this way. Good then.
> 

Unfortunately there is no SwiftLint rule to flag this as an error. So we might have to write a regex to find all of these occurrences.

eviltofu:
I'd suggest we next look at
`braces`
`sortedImports`

hboon:
> > ```
> > }.mapError { e in
> >     SmartContractError.embeded(e)
> > }.eraseToAnyPublisher()
> > ```
> > But yes, looks like the formatting rule works if formatted this way. Good then.
> 
> Unfortunately there is no SwiftLint rule to flag this as an error. So we might have to write a regex to find all of these occurrences.

It's ok, we can fix it somehow or as we see it I guess. As long as SwiftFormat doesn't undo it

hboon:
> I'd suggest we next look at
> braces
> sortedImports

Let's get this PR in first and try it with more files without necessarily committing them and we'll see.

eviltofu:
Ok I've pushed the changes to .swiftformat and the files list without the changed swift files. You can then discard any changed swift file after experimenting.

hboon:
> the --rules option only allows the rules you specify, so this is redundent?

Ah, I see. But why replace `--disable all` with `--rules` if they do the same thing? We should always strive to reduce unnecessary changes in a PR. Including unrelated things just makes it slower to merge and a higher chance of introducing problems.

---

Sorry, I wasn't clear, I mean:

* Make the diff for `.swiftformat` just introduce `indent (4)` and `typeSugar`, keeping the commented out `TODO`s in particular. Those are actually things we want to enable, or at least to look again
* Include the 5-7 `.swift` files too
* Flag the PR for review again

And we merge it, then do this outside of the PR:

> try it with more files without necessarily committing them and we'll see.

AlphaWallet.xcodeproj/project.pbxproj
```diff
@@ -4728,6 +4728,25 @@
...
+			);
+			inputPaths = (
+			);
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [ \"${CONFIGURATION}\" = \"Debug\" ]; then\n    while read F; do\n        \"${PODS_ROOT}/SwiftFormat/CommandLineTool/swiftformat\" $F\n    done < $SCRIPT_INPUT_FILE_LIST_0\nfi\n";
```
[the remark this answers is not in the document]
eviltofu:
```bash
if [ "${CONFIGURATION}" = "Debug" ]; then
    command_line = ""
    while read F; do
        command_line+="${F} "
    done < $SCRIPT_INPUT_FILE_LIST_0
    echo "${PODS_ROOT}/SwiftFormat/CommandLineTool/swiftformat" $command_line
fi
```

hboon (approved):

PR #6701, opened 2023-04-17
[Refactoring] Move function into same-file extension, reads better

hboon:
Small refactoring. Didn't mean to move code around unnecessarily, but hope to show what could be better.


hboon:
cc @eviltofu

PR #6702, opened 2023-04-17
Remove unnecessary `@MainActor` in Task

hboon:
@oa-s was it needed?


oa-s (commented):
not sure in this pr, might be needed to check whether other parts don't get updated on non main queue

hboon:
@oa-s ah I see. Ok, let's keep it.

PR #6711, opened 2023-04-21
Updated to iOS 16.2 and Xcode 14.2 in Travis

eviltofu:
Debugging test failures. Do not merge.


eviltofu:
@hboon In Travis build log. Will this affect anything in our build pipeline?
```
/Users/travis/.travis/job_stages: line 341: python: command not found
```

eviltofu:
@hboon It seems like Travis is really slow. I would like to set the expectation timeouts to a const which if it's running in Travis would be much longer then when we run locally if this possible.

hboon:
> @hboon In Travis build log. Will this affect anything in our build pipeline?
> 
> ```
> /Users/travis/.travis/job_stages: line 341: python: command not found
> ```

Not a problem

hboon:
> @hboon It seems like Travis is really slow.

Yes it is.

> I would like to set the expectation timeouts to a const which if it's running in Travis would be much longer then when we run locally if this possible.

Hmm.. but do the test pass on Travis if the timeouts are big enough?

eviltofu:
> > @hboon It seems like Travis is really slow.
> 
> Yes it is.
> 
> > I would like to set the expectation timeouts to a const which if it's running in Travis would be much longer then when we run locally if this possible.
> 
> Hmm.. but do the test pass on Travis if the timeouts are big enough?

I've set the timeouts for some tests that failed to 1 hour. Let's see if that works.

eviltofu:
@hboon I don't know how to use `travis_wait` with `set -o pipefail && xcodebuild -workspace AlphaWallet.xcworkspace -scheme AlphaWalletTests -destination platform\=iOS\ Simulator,OS\=16.0,name\=iPhone\ 14 build test | xcpretty`. Any ideas?

hboon:
Tried `travis_wait 30 <cmd>` instead of `travis_wait <cmd>`?

eviltofu:
> Tried `travis_wait 30 <cmd>` instead of `travis_wait <cmd>`?

Currently trying `travis_wait bash -c "set -o pipefail && xcodebuild -quiet -workspace AlphaWallet.xcworkspace -scheme AlphaWalletTests -destination platform\=iOS\ Simulator,OS\=16.2,name\=iPhone\ 14 test | xcpretty"`

PR #6716, opened 2023-04-21
fix crash when with rlp encoding

oa-s:


hboon (commented):
Hmm.. I thought I tested this. Does it only crash under certain conditions/input?

modules/AlphaWalletFoundation/AlphaWalletFoundation/EtherClient/TransactionSigning.swift
```diff
@@ -4,11 +4,11 @@ import BigInt
 import CryptoSwift
 
 public struct EthereumSignature {
-    public let v: UInt8
-    public let r: Data
-    public let s: Data
+    public let v: BigUInt
```
hboon, line 7: Do you want to just change `v` to `BigUInt` like you did, but keep `r` and `s` as `Data`? Seems unnecessary to change? Small diff and better performance too

hboon (approved):

PR #6717, opened 2023-04-21
fix displaying localized warning

oa-s:


hboon:
@oa-s can you help split this into 2 separate PRs? I'll look in the one about the warnings a little bit more

hboon (commented):
@oa-s see if https://github.com/AlphaWallet/alpha-wallet-ios/tree/alt-localized-warning-implementation is acceptable? The implementation has a bit of tight coupling (look for the comment "Bit of a tight coupling here, checking for a child type, but it makes implementation of `LocalizedWarning.warningDescription` mirror that of `LocalizedError.errorDescription`"), but it mirrors `Error`/`LocalizedError`, so for `Warning`/`LocalizedWarning`, we always implement`warningDescription` and for both: always call `localizedDescription`

If this looks right, you can replace your branch with it and update the commit message?

oa-s:
> https://github.com/AlphaWallet/alpha-wallet-ios/tree/alt-localized-warning-implementation

yes, looks good, lets use it

hboon:
@oa-s okie. Go for it. and let's merge

hboon (approved):

PR #6723, opened 2023-04-24
Remove ARTIS sigma1 and ARTIS tau1 since they are dead #6720

### LINKED ISSUE #6720: Remove ARTIS sigma1 and ARTIS tau1 since they are dead
Since they are dead

oa-s:
Closes #6720


AlphaWallet/Resources/Assets.xcassets/wallet_security_green_bar.imageset/Contents.json
```diff
@@ -1,5 +1,33 @@
 {
   "images" : [
+    {
```
hboon, line 3: The rest looks good, but looks like this shouldn't change?

AlphaWallet/Resources/Assets.xcassets/wallet_security_red_bar.imageset/Contents.json
```diff
@@ -1,5 +1,33 @@
 {
   "images" : [
+    {
```
hboon, line 3: And this too.

hboon (approved):
Also, need to edit `README.md` to remove the reference to Artis.

AlphaWallet/Resources/Assets.xcassets/wallet_security_green_bar.imageset/Contents.json
oa-s: don't know, maybe it because i deleted images related to artis networks?

hboon: Maybe, but seems odd that Xcode modifies an unrelated image. Think we can just remove them.

oa-s:
> Also, need to edit README.md to remove the reference to Artis. 

not sure if i can find it, could u please point me?

hboon:
There’s a reference to “artis” in https://github.com/AlphaWallet/alpha-wallet-ios/blob/ef7aed2a1c375f16da841b71ea3412ce67280892/README.md

PR #6724, opened 2023-04-25
Add code that allows sorting of RPCServer enums with CustomRPC structs such that mainnets come before testnets, RPCServers come before CustomRPC, and smaller chainIds before larger ones.

### LINKED ISSUE #6669: Modify the "weights in `RPCServers.displayOrderPriority` so that the same order is preserved, except testnets come after all mainnets
Include a test to make sure that in the results of `RPCServers.displayOrderPriority` all testnets come after mainnets

eviltofu:
Closes #6669

Test is included.

To sort a list of RPCServers, you can just use `.sorted` to get a list of RPCServers with testnets at the end of the list.
You can use `.sorted` on CustomRPC structs too. The testnets will be at the end of the list.
You can sort a mixture of RPCServers enums and CustomRPC structs together but it's a bit more complicated. You can use the code below to sort the list of mixed data structures:
```swift
let unsorted: [isTestnetSortableEnum] = [
            a.testnetSortable,
            b.testnetSortable,
            c.testnetSortable,
            d.testnetSortable,
            e.testnetSortable,
            f.testnetSortable,
            g.testnetSortable
        ]
        let sorted: [Any] = unsorted.sorted().map { item in
            item.server
        }
```
RPCServer enums will sort before CustomRPC structs.
Mainnets will sort before testnets.
Smaller ChainIds will sort before larger ones.

### DIFF
--- a/AlphaWallet/Extensions/RPCServer+Extensions.swift
+++ b/AlphaWallet/Extensions/RPCServer+Extensions.swift
@@ -138,3 +138,16 @@ extension RPCServer {
         }
     }
 }
+
+extension RPCServer: Comparable {
+    public static func < (lhs: RPCServer, rhs: RPCServer) -> Bool {
+        switch (lhs.isTestnet, rhs.isTestnet) {
+        case (true, false):
+            return false
+        case (false, true):
+            return true
+        default:
+            return lhs.displayOrderPriority < rhs.displayOrderPriority
+        }
+    }
+}
--- a/AlphaWalletTests/Tokens/RPCServerComparableTest.swift
+++ b/AlphaWalletTests/Tokens/RPCServerComparableTest.swift
@@ -1,0 +1,35 @@
+//
+//  RPCServerComparableTest.swift
+//  AlphaWalletTests
+//
+//  Created by Jerome Chan on 25/4/23.
+//
+
+@testable import AlphaWallet
+import XCTest
+import AlphaWalletFoundation
+
+final class RPCServerComparableTest: XCTestCase {
+
+    func testRPCServerComparable() {
+        let sorted: [RPCServer] = [
+            .custom(CustomRPC.custom(chainId: 102)), // chainId 102
+            .goerli, // chainId 5 - testnet
+            .main, // chainId 1
+            .arbitrumGoerli, // chainId 421613 - testnet
+            .arbitrum, // chainId 42161
+            .binance_smart_chain_testnet, // chainId 97 - testnet
+            .binance_smart_chain, // chainId 56
+            .custom(CustomRPC.custom(chainId: 101)) // chainId 101
+        ].sorted()
+        XCTAssertTrue(sorted[0] == .main)
+        XCTAssertTrue(sorted[1] == .binance_smart_chain)
+        XCTAssertTrue(sorted[2] == .arbitrum)
+        XCTAssertTrue(sorted[3] == .custom(chainId: 101))
+        XCTAssertTrue(sorted[4] == .custom(chainId: 102))
+        XCTAssertTrue(sorted[5] == .goerli)
+        XCTAssertTrue(sorted[6] == .binance_smart_chain_testnet)
+        XCTAssertTrue(sorted[7] == .arbitrumGoerli)
+    }
+
+}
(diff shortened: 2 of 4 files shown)


AlphaWalletTests/Tokens/RPCServerComparableTest.swift
```diff
@@ -0,0 +1,90 @@
...
+//  Created by Jerome Chan on 25/4/23.
+//
+
+@testable import AlphaWallet
+import XCTest
+import AlphaWalletFoundation
+
+final class RPCServerComparableTest: XCTestCase {
+
+    override func setUpWithError() throws {
```
hboon, line 14: Can you remove these empty functions?

hboon (changes requested):
Can you elaborate on what this PR does? I know there's tests. Does it change the app's functionality or replace how some parts are implemented? — I see there is non-test code, are those called only from the tests?

eviltofu:
> Can you elaborate on what this PR does? I know there's tests. Does it change the app's functionality or replace how some parts are implemented? — I see there is non-test code, are those called only from the tests?

Add code that allows sorting of RPCServer enums with CustomRPC structs such that mainnets come before testnets, RPCServers come before CustomRPC, and smaller chainIds before larger ones.

This is in response to ```
Modify the "weights in RPCServers.displayOrderPriority so that the same order is preserved, except testnets come after all mainnets
``` in #6669

File: AlphaWalletTests/Tokens/RPCServerComparableTest.swift (hunk 1 above)
Comment (eviltofu) [reply AlphaWalletTests/Tokens/RPCServerComparableTest.swift]: Removed

File: AlphaWallet/Extensions/RPCServer+Extensions.swift [hunk 2]
```diff
@@ -144,3 +144,104 @@ extension RPCServer {
...
+        case (true, false):
+            return false
+        case (false, true):
+            return true
+        default:
+            return lhs.displayOrderPriority < rhs.displayOrderPriority
+        }
+    }
+    public static func < (lhs: RPCServer, rhs: CustomRPC) -> Bool {
+        switch (lhs.isTestnet, rhs.isTestnet) {
```
hboon, line 160: But `lhs` can be `RPCServer.custom`. Do we need to take that into account?

hboon (commented):
Ah. Sorry, I thought that one's been closed and didn't notice the reference to the issue.

---

Maybe I misunderstood what `isTestnetSortableEnum` is for, but do we ever mix `RPCServer` and `CustomRPC` in a collection and want to sort them? (Also the name doesn't suggest what it does. Maybe you mean something like `SortableServer`?)

This PR only adds the functionality to sort, but not use it, right?

---

It looks like a lot of additional code and we introduced a few protocols. We don't need `RPCServer` to always be comparable/sortable. Does this simplify the implementation?

We already build a sorted list of `RPCServer`s instances in `ServersCoordinator.serversOrdered`. Can we do this:

1. Do a quick check if these are the only ways we get a list of servers. (This seems too time-consuming):
    * `RPCServer.allCases` (which of course lists them out explicitly)
    * `ServersCoordinator.serversOrdered`
    * `RPCServers.availableServers`

2. Then, modify `ServersCoordinator.serversOrdered` to involve a function or 2 (using the logic you implemented for conforming `Comparable`) that sorts them and make sure to access that list when we need a sorted list. The property is not in the best place and can be improved, but the sorted list is really only used in the UI.

AlphaWallet/Extensions/RPCServer+Extensions.swift
eviltofu: The next two functions takes that into consideration.
`public static func < (lhs: CustomRPC, rhs: RPCServer) -> Bool`
`public static func < (lhs: CustomRPC, rhs: CustomRPC) -> Bool`

hboon: I mean, in the test, we have:

 ```
 let c = RPCServer.goerli // chainId 5
 let d = CustomRPC.custom(chainId: 100)
 ```

 What if it's:

 ```
 let c = RPCServer.goerli // chainId 5
 let d = RPCServer.custom(CustomRPC.custom(chainId: 100))
 ```

 Maybe this is solved by one of the comparison functions. But seems it needs to be handled in here instead? —

```
public static func < (lhs: RPCServer, rhs: RPCServer) -> Bool {
```

eviltofu: So where should this `RPCServer.custom` be in terms of precedence? Should it be treated the same as CustomRPC?

hboon: Yes.

hboon: Also, I don't think `RPCServer` ever gets compared to `CustomRPC` directly. Only indirectly vai `RPCServer.custom`  (in case you can remove some more code)

hboon:
@eviltofu Might have missed this? https://github.com/AlphaWallet/alpha-wallet-ios/pull/6724#pullrequestreview-1399090075

Looks like the additional protocols aren't needed, right?

Also, about the implementation for `< (lhs: RPCServer, rhs: RPCServer)`:

* should `default:` also handle if it's comparing custom+notCustom and nonCustom+custom?
* I *think* when it's custom vs custom, we can just use `displayOrderPriority`. It already takes into account the `chainID`, but not sure if it will be correct after we tweak the function and add things like comparing custom+nonCustom

Do we still need `< (lhs: RPCServer, rhs: CustomRPC)` and `< (lhs: CustomRPC, rhs: RPCServer)`? If we keep them we can also call them from within the implementation of `< (lhs: RPCServer, rhs: RPCServer)`?

eviltofu:
Removed the not needed protocols.

PR #6727, opened 2023-04-25
Reduce AppCoordinator file size, move logic to Application

oa-s:


hboon (approved):
Too big to check. Will test with next TestFlight

PR #6728, opened 2023-04-25
Update JsonFromTokenUri with api transporter to being able to handle retrieble error codes [429, 408, 500, 502, 503, 504]

oa-s:


hboon (approved):

hboon:
Merge when ready (I see you are still force-pushing, so didn't merge)

PR #6733, opened 2023-04-25
Stop matching ticker IDs by ignoring the platform, this produces false positives #5275

### LINKED ISSUE #5275: Stop matching ticker IDs by ignoring the platform, this produces false positives
Eg. both https://candleexplorer.com/address/0x6Ee592139e9DD84587a32831A33a32202d1f0F12 and https://candleexplorer.com/address/0xB910B763F3ee8eD2102a22ABC77401DAaa15587e would match USDC with price = $1 because name = "USDC" and symbol = "USDC", but anyone can create such a token on any chain

oa-s:
Closes #5275


hboon (approved):
Thanks! We might be able to improve on this a bit with the `tokens.json` file later

PR #6735, opened 2023-04-25
fix swap token networking, remove unused param caused decoding error

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/SwapToken/NativeSwap/SwapEstimate.swift
```diff
@@ -36,7 +36,6 @@ public struct SwapEstimate {
     }
 
     public struct SwapStep {
-        public let unsignedSwapTransaction: UnsignedSwapTransaction
```
hboon, line 39: We still use it in `SwapQuote`. Is it the same shape?

oa-s: yes, i checked `unsignedSwapTransaction: UnsignedSwapTransaction` and it looks unused, maybe has left after some migration.

hboon (approved):

PR #6749, opened 2023-04-27
[Refactoring] rename cancelable with cancellable

oa-s:


oa-s:
too many files, but not critical changes, just var renaming

hboon (commented):
@oa-s "cancellable" is British spelling so it's fine if you prefer to keep it.

hboon:
Oh, you renamed it. Then: "cancelable" is fine too, since it's American :)

oa-s:
oh, that we can close it

PR #6750, opened 2023-04-27
prepare for push notifications #6707

### LINKED ISSUE #6707: Implement push notification
Refer https://github.com/AlphaWallet/alpha-wallet-android/issues/3152

Should need some details like our backend's web API endpoints for registering a wallet and the payload format (suggest changes if you need, to support both iOS and Android). @justindg should have them

oa-s:
Closes #6707


modules/AlphaWalletNotifications/.gitignore
```diff
@@ -0,0 +1,98 @@
+# Xcode
```
hboon, line 1: I think we don't need the `.gitignore` file right? Maybe a remnant from when we (wanted?) to have a separate repo for each? Shall we remove them separately?

```
./modules/AlphaWalletCore/.gitignore
./modules/AlphaWalletTrackAPICalls/.gitignore
./modules/AlphaWalletAddress/.gitignore
./modules/AlphaWalletLogger/.gitignore
```

modules/AlphaWalletNotifications/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 AlphaWallet
```
hboon, line 3: 2023 :)

AlphaWalletTests/ViewControllers/PaymentCoordinatorTests.swift
```diff
@@ -24,10 +24,14 @@ extension CurrencyService {
 }
 
 extension WalletDataProcessingPipeline {
+<<<<<<< HEAD
```
hboon, line 27: There are a few merge conflict markers

hboon (commented):
Also problem building after removing the merge conflict markers

hboon:
Seems to include unrelated changes like for `SwitchTableViewCell`?

oa-s:
for this pr `SwitchTableViewCell` was needed to update with adding loadable state, for treaking push notification state enable/disable

modules/AlphaWalletNotifications/.gitignore
oa-s: don't know, will remove `.gitignore` related to this pr, the rest in another pr

hboon:
I'm getting this build error:

> (103, 25) Value of type 'Loadable<TransactionConfirmationViewModel.SendFungiblesTransactionViewModel.TransactedToken, any Error>' has no member 'mapValue'

Just me?

hboon:
Does it actually show a local notification already? If it does, can we hide a feature flag to disable it?

oa-s:
Yes, it does

hboon:
The feature flag should disable the registration and the accepting the notification and starting processing, so something like:

```
if feature_flag
    prompt_user_to_get_device_token
end

when_push_received
    if feature_flag
        refresh_timelines
        show_local_notification
    end
end
```

oa-s:
> mapValue

will check

oa-s:
@hboon  our current implementation of fetching transactions suppose to make timed call of transaction request (it looks like a bit different from Androids impl), i checked Android implementation and maybe not clearly understood but in general they just fetch latest transactions once a push notification received. Would it be correctly for us just run timers that make appropriate requests?

our SingleChainTransactionProvider has `start` method, so we could stop service in case its in `running` state, e.g waiting for next cycle call. and force start timer again. Our `TransactionsNotificationSource` will handle received transactions and will create appropriate local notifications.

oa-s:
> So upon receiving the remote notification, we just do this? —
> ``` _pull_with_etherscan() ```

yes

> This is what Android does too right?

yes

oa-s:
>  pull_transactions()

what does it should do?

oa-s:
> The push notification just triggers it right away.

Ok i see. No u understood it right, that is just me don't understood right what push notification should do, but now i do. So yes we can trigger to pull transactions by push notification (meaning to restart fetch transactions timers).

oa-s:
firstly i thought that all needed data will be in the received push notification, so we wouldn't need to do any additional actions to display push notification, but now i got it

hboon:
Yes. The remote notification is like a refresh button :) it doesn’t contain much information. For context, it can’t contain much information because we don’t have a backend wallet API that the wallet syncs with.

oa-s:
@hboon it looks like we are using different projects for ios and android applications, could u please check android app if u have access to it and invite me to it. Need choose one from two. push notification api configured with android project

hboon:
@oa-s seems like the server has now moved to using the project iOS is on right? Let's coordinate with Justin and move them to use the iOS one?

PR #6752, opened 2023-04-28
Should not be possible to go dapp browser home screen after opening browser in read-only mode #6613

### LINKED ISSUE #6613: Should not be possible to go dapp browser home screen after opening browser in read-only mode
1. Wallet tab
2. Tap ETH token
3. Tap token icon
4. Tap back (of back and forward) button — does nothing. Good
5. With device bezels enabled, swipe from left to right

Expected
---
6. nothing happens

Observed
---
6. The dapp browser home screen is revealed

Screenshot 2023-03-29 at 7 46 19 AM (2)

oa-s:
Closes #6613

### DIFF
--- a/AlphaWallet/Common/Types/AppStyle.swift
+++ b/AlphaWallet/Common/Types/AppStyle.swift
@@ -82,9 +82,9 @@ extension UITabBarAppearance {
     }
 }
 
-extension UITabBarController {
-    static func withOverridenBarAppearence(appearence tabBarAppearance: UITabBarAppearance = .defaultAppearence) -> UITabBarController {
-        let tabBarController = UITabBarController()
+extension TabBarController {
+    static func withOverridenBarAppearence(appearence tabBarAppearance: UITabBarAppearance = .defaultAppearence) -> TabBarController {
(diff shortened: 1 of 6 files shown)


AlphaWallet/Common/Types/AppStyle.swift
```diff
@@ -82,9 +82,9 @@ extension UITabBarAppearance {
     }
 }
 
-extension UITabBarController {
-    static func withOverridenBarAppearence(appearence tabBarAppearance: UITabBarAppearance = .defaultAppearence) -> UITabBarController {
-        let tabBarController = UITabBarController()
+extension TabBarController {
```
hboon, line 85: This should still work even if we don't introduce `TabBarController`, right?

hboon (approved):

AlphaWallet/Common/Types/AppStyle.swift
oa-s: i was facing an issue when we do a swipe on browser tab its gonna to be swiped up to wallets screen, updating TabBar helps to solve that.

PR #6753, opened 2023-04-28
xDai chain should use gas price estimate returned from RPC node with eth_gasPrice #6550

### LINKED ISSUE #6550: xDai chain should use gas price estimate returned from RPC node with `eth_gasPrice`
It's about 1.9 gwei as of writing (20230320)

We should most probably remove `GasPriceConfiguration.xDaiGasPrice`.

Might have to modify the criteria ( < 1) added by #6548 so we don't pad small estimates by too much. ie we shouldn't do "1.9 + 1" gwei. Maybe "1.9 * 10%" for smaller amount of gas (say < 5gwei) ?

oa-s:
Closes #6550

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/RPC/BlockchainProvider.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/RPC/BlockchainProvider.swift
@@ -146,13 +146,15 @@ public final class RpcBlockchainProvider: BlockchainProvider {
             .handleEvents(receiveOutput: { [server] estimate in
                 infoLog("[RPC] Estimated gas price with RPC node server: \(server) estimate: \(estimate)")
             }).map { [params] gasPrice in
-                if (gasPrice + GasPriceConfiguration.oneGwei) > params.maxPrice {
+                //Add an extra gwei because the estimate is sometimes too low. We mustn't do this if the gas price estimated is lower than 1gwei since chains like Arbitrum is cheap (0.1gwei as of 20230320)
+                let surchargedGasPrice = params.gasPriceSurchargeding.bufferedGasPrice(estimatedGasPrice: gasPrice)
+
+                if surchargedGasPrice.value > params.maxPrice {
                     // Guard against really high prices
                     return LegacyGasEstimates(standard: params.maxPrice)
                 } else {
-                    if params.canUserChangeGas && params.shouldAddBufferWhenEstimatingGasPrice, gasPrice > GasPriceConfiguration.oneGwei {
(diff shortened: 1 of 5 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/RPC/BlockchainProvider.swift
```diff
@@ -146,13 +146,15 @@ public final class RpcBlockchainProvider: BlockchainProvider {
...
-                if (gasPrice + GasPriceConfiguration.oneGwei) > params.maxPrice {
+                let surchargedGasPrice = params.gasPriceSurchargeding.surchargedGasPrice(estimatedGasPrice: gasPrice)
+
+                if surchargedGasPrice.value > params.maxPrice {
                     // Guard against really high prices
                     return LegacyGasEstimates(standard: params.maxPrice)
                 } else {
-                    if params.canUserChangeGas && params.shouldAddBufferWhenEstimatingGasPrice, gasPrice > GasPriceConfiguration.oneGwei {
+                    if params.canUserChangeGas && params.shouldAddBufferWhenEstimatingGasPrice, gasPrice > surchargedGasPrice.surchargeding {
                         //Add an extra gwei because the estimate is sometimes too low. We mustn't do this if the gas price estimated is lower than 1gwei since chains like Arbitrum is cheap (0.1gwei as of 20230320)
```
hboon, line 156: Better to move the comment to be above this line:

```
let surchargedGasPrice = params.gasPriceSurchargeding.surchargedGasPrice(estimatedGasPrice: gasPrice)
```

modules/AlphaWalletFoundation/AlphaWalletFoundation/RPC/BlockchainProvider.swift
```diff
@@ -234,3 +236,20 @@ extension JSONRPCKit.JSONRPCError: LocalizedError {
...
+
+public enum GasPriceSurchargeding {
+    case percentage(BigUInt)
+    case fixed(BigUInt)
+
+    public func surchargedGasPrice(estimatedGasPrice: BigUInt) -> (value: BigUInt, surchargeding: BigUInt) {
+        let surchargeding: BigUInt
+        switch self {
+        case .percentage(let gasPriceSurchargePercent):
+            surchargeding = BigUInt(Int(Double(estimatedGasPrice) / 100.0 * Double(gasPriceSurchargePercent)))
```
hboon, line 248: Is it better like this? Less loss of precision and having to think about conversions when reading the code:

```
surchageding = estimatedGasPrice * gasPriceSurchargePercent / BigUInt(100)
```

oa-s:
@hboon updated the pr according to comments above

hboon (approved):
Didn't check. Assume still good :)

PR #6757, opened 2023-05-01
Group move oneTimeCreationOfOneDatabaseToHoldAllChains to db migration

oa-s:


AlphaWallet/ActiveWalletCoordinator.swift
```diff
@@ -228,7 +228,6 @@ class ActiveWalletCoordinator: NSObject, Coordinator {
     func start(animated: Bool) {
         donateWalletShortcut()
 
-        oneTimeCreationOfOneDatabaseToHoldAllChains()
```
hboon, line 231: @oa-s 

```
$ gl -S 'oneTimeCreationOfOneDatabaseToHoldAllChains'

commit e8ba1756106ed7628d5bfda380bb1f92367be3d1
Date:   2019-02-27 14:57:55 +0800

    Paradigm change: Multi-chain active at the same time instead of having to switch the active chain.
```

Multi-chain support was introduced more than 4 years ago. Only users who have used a build before that and *never* used an updated version for 4+ years would be affected — how about we remove `oneTimeCreationOfOneDatabaseToHoldAllChains(assetDefinitionStore:)` instead? Simplifies things a little, removing some baggage

PR #6758, opened 2023-05-01
move FetchTokenScriptFiles to TokensService

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/AlphaWalletTokensService.swift
```diff
@@ -118,6 +123,8 @@ public class AlphaWalletTokensService: TokensService {
                 return providers
             }.assign(to: \.value, on: providers, ownership: .weak)
             .store(in: &cancelable)
+
+        fetchTokenScriptFiles.start()
```
hboon, line 127: But the pipelines (over here) run once for each wallet, but pre-PR, it only runs once. Does it make the runs redundant?

oa-s: so the better would be to move it to app, and run it there only once

oa-s: since it uses tokens datasource we need run it when we have data source, so TokensService its a good place for that, it creates only once for each wallet, so logically nothing should change, except the place of locating `fetchTokenScriptFiles `

hboon: Up to you as long as you know the difference and intricacies :)

hboon (approved):

PR #6774, opened 2023-05-03
Classify a few types comforming to `Initializer` as a new protocol `Service` instead, to be more intention revealing, reflecting that they are long-running

hboon:


oa-s (approved):

oa-s:
is its possible to move this code to service too?
```
if AlphaWallet.Device.isSimulator {
            //Want to start as soon as possible
            TrackApiCalls.shared.start()

            UserDefaults.standard.set(!isRunningTests(), forKey: "_UIConstraintBasedLayoutLogUnsatisfiable")
        }
```

hboon:
Go ahead. But note the comment though

PR #6775, opened 2023-05-03
Handle long request retry time #6773

### LINKED ISSUE #6773: Handle long request retry time
Coingecko api returns retry time of 600 seconds

@hboon  do we need to drop request when its timeout is less than retry time?

oa-s:
Closes #6773

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Types/TickerIdFilter.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Types/TickerIdFilter.swift
@@ -73,6 +73,11 @@ public class TickerIdFilter {
     }
 
     func filterMathesInPlatforms(token: TokenMappedToTicker, tickerId object: TickerIdObject) -> Bool {
+        func isMatchingSymbolAndName(token: TokenMappedToTicker, tickerIdObject object: TickerIdObject) -> Bool {
+            guard token.contractAddress == Constants.nativeCryptoAddressInDatabase else { return false }
+            return object.symbol.compare(token.symbol, options: .caseInsensitive) == .orderedSame && object.name.compare(token.name, options: .caseInsensitive) == .orderedSame
+        }
+
         //We just filter out those that we don't think are supported by the API. One problem this helps to alleviate is in the API output, certain tickers have a non-empty platform yet the platform list might not be complete, eg. Ether on Ethereum mainnet:
         //{
         //   "symbol" : "eth",
@@ -97,7 +102,7 @@ public class TickerIdFilter {
                 return false
             }
         } else {
-            return false
+            return isMatchingSymbolAndName(token: token, tickerIdObject: object)
         }
     }
 }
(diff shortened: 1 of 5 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Types/TickerIdFilter.swift
```diff
@@ -73,6 +73,11 @@ public class TickerIdFilter {
     }
 
     func filterMathesInPlatforms(token: TokenMappedToTicker, tickerId object: TickerIdObject) -> Bool {
+        func isMatchingSymbolAndName(token: TokenMappedToTicker, tickerIdObject object: TickerIdObject) -> Bool {
```
hboon, line 76: This seems to be old code?

modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Types/TickerIdFilter.swift
```diff
@@ -97,7 +102,7 @@ public class TickerIdFilter {
                 return false
             }
         } else {
-            return false
+            return isMatchingSymbolAndName(token: token, tickerIdObject: object)
```
hboon, line 105: This too

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/ApiTransporter.swift
```diff
@@ -15,19 +15,29 @@ public protocol ApiTransporter {
...
                         completion: @escaping (RetryResult) -> Void) {
+        let d = request.cURLDescription()
+        if d.contains("api.coingeck") {
+            print("xxx.retry for \(request.cURLDescription()) error: \(error)")
+        }
 
         if request.retryCount < retryLimit, shouldRetry(request: request, dueTo: error) {
             if let httpResponse = request.response, let delay = ApiTransporterRetryPolicy.retryDelay(from: httpResponse) {
-                completion(.retryWithDelay(delay))
+                if delay <= timeout {
```
hboon, line 36: Based on:

```
if delay <= timeout {
    completion(.retryWithDelay(delay))
} else {
    completion(.doNotRetry)
}
```

and:

> * Timeout — we send a request and start waiting for a response. The max time we wait before dropping it is the timeout. This might be say, 1 to 10 seconds
> * Time to retry returned by CoinGecko — this is the time that CoinGecko says we should hold off before making requests again. In other words, we are rate limited. This could be say, 600 seconds

`delay` is "time to retry" right?

So:

```
if delay_600 <= timeout_60
    retry
else
    do_not_retry
```

I don't think they should be compared. They are different concepts.

We set a timeout of `60`, that's fine. That's because we presume the API takes a while to respond. Then it should never change.

And if the API responds that we are rate limited, that we should back off for at least 600s, then we should do something like this:

```
func call_coingecko()
    .request(timeout_60)
    .process {
        if delay
            sleep(delay)
            call_coingecko() #retry, only a number of times
        else
            return results
        end
    }
```

modules/AlphaWalletFoundation/AlphaWalletFoundation/CoinTicker/Types/TickerIdFilter.swift
oa-s: @oa-s  this is a bit different, (might move in another pr), updated code to return it back because i wasn't able to resolve ticker id for currencies like ether or other native cryptos

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/ApiTransporter.swift
oa-s: ok got it, than leave as it is for now

PR #6786, opened 2023-05-04
add ability to restrict qrcode resolutions #3787

### LINKED ISSUE #3787: Scanning wrong type of QR code results in the app asking you to open the browser but nothing happens after that
In the Import Wallet screen, if you scan a Singapore COVID QR entry code with the AlphaWallet qr code scanner, it asks if you want to open the url in the browser but if the user answers open nothing happens. If you scan in other places, the app does open a built in browser.

https://user-images.githubusercontent.com/1050309/150711372-749056c4-e24a-4220-82e3-47e119cb2ba9.mov

oa-s:
Closes #3787


hboon (commented):
What does this PR do?

oa-s:
this pr helps to avoid issues like in #3787 we manually set to resolver to deny resolve specific type of qr code resolution.
for example when we scanning qr code for address text field (we expect to resolve address), but received url resolved value we ignore it.

hboon (approved):

hboon:
@oa-s thanks. Good to describe that in the commit message :)

PR #6794, opened 2023-05-09
Split Etherscan and Blockscout on two independent networkings #6792

### LINKED ISSUE #6792: Split Etherscan and Blockscout on two independent networkings
`Blockscout` ignores params: `startblock`, `endblock`, and wants them to be `start_block`, `end_block`, so existed pagination doesn't work for `Blockscout`.

docs: https://explorer.palm-uat.xyz/api-docs#account

oa-s:
Closes #6792

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/BlockscoutApiNetworking.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/BlockscoutApiNetworking.swift
@@ -1,6 +1,8 @@
 //
-// Created by James Sangalli on 6/6/18.
-// Copyright © 2018 Stormbird PTE. LTD.
+//  BlockscoutApiNetworking.swift
+//  AlphaWalletFoundation
+//
+//  Created by Vladyslav Shepitko on 09.05.2023.
 //
 
 import Foundation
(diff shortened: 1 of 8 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/Types/GetTransactionsRequest.swift
```diff
@@ -8,44 +8,12 @@
...
-    struct NoBlockchainExplorerApi: Error {
-    }
-
-    let server: RPCServer
-    let address: AlphaWallet.Address
-    let startBlock: Int
-    let endBlock: Int
-    let sortOrder: SortOrder
-
     public enum SortOrder: String {
```
hboon, line 11: Seems like we don't need `GetTransactions` anymore. If we want to keep the PR smaller, can just remove `asURLRequest()` and make `GetTransactions` no longer conform to `URLRequestConvertible`?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/BlockscoutApiNetworking.swift
```diff
@@ -0,0 +1,318 @@
...
+            .tryMap { UniqueNonEmptyContracts(json: try JSON(data: $0.data), tokenType: .erc20) }
+            .mapError { PromiseError.some(error: $0) }
+            .print("xxx.blockscout.normalTransactions.server: \(server)")
+            .eraseToAnyPublisher()
+    }
+
+    func erc721TokenInteractions(walletAddress: AlphaWallet.Address,
+                                 startBlock: Int?) -> AnyPublisher<UniqueNonEmptyContracts, PromiseError> {
+
+        switch server {
```
hboon, line 58: Is this switch for logging the same for all cases in both `EtherscanCompatibleApiNetworking` and `BlockscoutApiNetworking`? Maybe useful to extract it out into an extension of `ApiNetworking`? Should be useful to keep it around and still log it with `debugLog()` and even crash with `precondition()`

hboon (commented):
Seems OK

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/Types/GetTransactionsRequest.swift
oa-s: yes, i will remove it in another pr, to avoid conflicts for now

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/BlockscoutApiNetworking.swift
oa-s: its a bit different for eip 1155

PR #6800, opened 2023-05-10
Validate token transfer transactions

oa-s:
validate token transfer transaction, add filtering for only expected transaction types
e.g Blockscout api might return eip20 an eip721 token transfer transactions in same response

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/UniqueNonEmptyContracts.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/UniqueNonEmptyContracts.swift
@@ -45,9 +45,9 @@ extension UniqueNonEmptyContracts {
             //NOTE: safe check to avoid incompatible contract matching with token type, blockout api returns erc20 and erc721 for same url, so we need to filter retults
             switch tokenType {
             case .erc20:
-                guard json["tokenID"].string == nil else { return nil }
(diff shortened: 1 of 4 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/UniqueNonEmptyContracts.swift
```diff
@@ -45,9 +45,9 @@ extension UniqueNonEmptyContracts {
             //NOTE: safe check to avoid incompatible contract matching with token type, blockout api returns erc20 and erc721 for same url, so we need to filter retults
             switch tokenType {
             case .erc20:
-                guard json["tokenID"].string == nil else { return nil }
+                guard json["tokenID"].stringValue.isEmpty && json["tokenValue"].stringValue.isEmpty else { return nil }
```
hboon, line 48: What does `json["tokenValue"]` represent in the JSON for various token types?

oa-s: `tokenValue` supported only for erc1155, and represents transferred token amount
here is docs. https://docs.etherscan.io/api-endpoints/accounts#get-a-list-of-erc1155-token-transfer-events-by-address

hboon (approved):

PR #6803, opened 2023-05-10
handle null rpc request id

oa-s:


modules/AlphaWalletWeb3/AlphaWalletWeb3/Promises/Promise+Batching.swift
```diff
@@ -57,6 +57,10 @@ public class JSONRPCrequestDispatcher {
         case .NoBatching:
             return provider.sendAsync(request, queue: queue)
         case .Batch:
+            guard request.id != nil else {
```
hboon, line 60: When is RPC request ID nil? I remember one RPC provider that behaves differently from the others; it only returns a single error instead of an array when there's an error. See 709cfde8479fb0e3c9a88c78297a57cd6f50ed71 in case it matters

oa-s: some of rpc nodes return an error with nil id, i will double check comments in https://github.com/AlphaWallet/alpha-wallet-ios/commit/709cfde8479fb0e3c9a88c78297a57cd6f50ed71, but suppose it should be good for it

oa-s: updated pr with sharing single error for batch of requests

hboon (approved):

PR #6812, opened 2023-05-12
Replace etherscan provider with schedulers

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/EtherscanCompatibleApiNetworking.swift
```diff
@@ -57,7 +57,7 @@ class EtherscanCompatibleApiNetworking: ApiNetworking {
         case .main, .classic, .goerli, .xDai, .polygon, .binance_smart_chain, .binance_smart_chain_testnet, .callisto, .optimistic, .cronosMainnet, .cronosTestnet, .custom, .arbitrum, .palm, .palmTestnet, .optimismGoerli, .arbitrumGoerli, .avalanche, .avalanche_testnet, .sepolia:
             break
         case .heco, .heco_testnet, .fantom, .fantom_testnet, .mumbai_testnet, .klaytnCypress, .klaytnBaobabTestnet, .ioTeX, .ioTeXTestnet, .okx:
-            return .empty()
+            return .fail(PromiseError(error: ApiNetworkingError.methodNotSupported))
```
hboon, line 60: Is it better to mention which method it is in the error?

oa-s: could be good for logging, will do

PR #6813, opened 2023-05-12
preparing transaction provider for push notifications, handle transaction provider state

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/TransactionsService.swift
```diff
@@ -142,17 +130,31 @@ public class TransactionsService {
...
+    }
+    //TODO: call when receive a push notification
+    public func forceResumeOrStart(server: RPCServer) {
+        guard let provider = providers[server] else { return }
+
+        switch provider.state {
+        case .pending:
+            provider.start()
+        case .running:
+            provider.stop()
```
hboon, line 154: Does it have to stop if it's already running? Is it because it might miss a new transaction?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/EtherscanSingleChainTransactionProvider.swift
```diff
@@ -30,6 +30,8 @@ class EtherscanSingleChainTransactionProvider: SingleChainTransactionProvider {
             ercTokenDetector: ercTokenDetector)
     }()
 
+    public private (set) var state: TransactionProviderState = .pending
```
hboon, line 33: Is `.pending` only used as the initial state?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/TransactionsService.swift
```diff
@@ -142,17 +130,31 @@ public class TransactionsService {
...
         for each in providers {
-            each.value.runScheduledTimers()
+            each.value.resume()
+        }
+    }
+    //TODO: call when receive a push notification
+    public func forceResumeOrStart(server: RPCServer) {
+        guard let provider = providers[server] else { return }
+
+        switch provider.state {
```
hboon, line 150: I think I understand, but the functions called in the `switch` statement reads a little strange, eg. `stop()` followed by `resume()` when it's `.running`. Usually it's `stop()` and then `start()`. Similarly calling `resume()` when it's `.stopped`.

Or how about changing `func stop()` to `func pause()` if it makes sense? Reads a little better?

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/EtherscanSingleChainTransactionProvider.swift
oa-s: yes, right

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/TransactionsService.swift
oa-s: > Or how about changing func stop() to func pause() if it makes sense? Reads a little better?

sure

modules/AlphaWalletFoundation/AlphaWalletFoundation/Transactions/TransactionsService.swift
oa-s: we stop it because in case its running (waiting for next call), that might take some time, (cancelling request shouldn't cause transactions missing)  and running a new cycle (calling resume method) we make force fetch making api calls.
So we reset current cycle in case its running and start new one.

hboon (approved):

PR #6826, opened 2023-05-18
improve reloading token balance

oa-s:
- Fire reloading token balance once server enabled or disabled;
- stop reload tokens balance scheduler when app enters in background state.

### DIFF
--- a/modules/AlphaWalletFoundation/AlphaWalletFoundation/WalletBalance/WalletBalanceFetcher.swift
+++ b/modules/AlphaWalletFoundation/AlphaWalletFoundation/WalletBalance/WalletBalanceFetcher.swift
@@ -8,6 +8,7 @@
 import Foundation
 import BigInt
 import Combine
+import AlphaWalletCore
 
 public protocol WalletBalanceFetcherType: AnyObject {
     var walletBalance: AnyPublisher<WalletBalance, Never> { get }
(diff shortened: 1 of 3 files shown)


modules/AlphaWalletFoundation/AlphaWalletFoundation/WalletBalance/WalletBalanceFetcher.swift
```diff
@@ -38,20 +39,35 @@ public class WalletBalanceFetcher: NSObject, WalletBalanceFetcherType {
...
-            .assign(to: \.value, on: walletBalanceSubject)
+            .assign(to: \.value, on: subject)
+            .store(in: &cancelable)
+
+        NotificationCenter.default.applicationState
+            .receive(on: RunLoop.main)
+            .sink { [weak scheduler] state in
+                switch state {
+                case .didEnterBackground:
+                    scheduler?.cancel()
```
hboon, line 63: Does this need to be explicit? Wouldn't it be paused automatically by the OS?

oa-s: no it wouldn't, it still works

hboon: No I mean. Before the PR, if the app goes into the background, it will pause too, right? What’s the difference?

oa-s: yes, it works in same way with prev code.
added logs to timed call and when app goes to background state
```
xxx.Timer
xxx.Timer
xxx.Timer
xxx.applicationState: didEnterBackground
xxx.applicationState: didEnterBackground
xxx.applicationState: didEnterBackground
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer
xxx.Timer

```

hboon (approved):

PR #6828, opened 2023-05-18
filter localized operation object to avoid crash when dublicates appear

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/Logic/EtherscanCompatibleApiNetworking.swift
```diff
@@ -553,7 +553,7 @@ extension EtherscanCompatibleApiNetworking.functional {
         for each in transactions {
             if let index = results.firstIndex(where: { $0.blockNumber == each.blockNumber }) {
                 var found = results[index]
-                found.localizedOperations.append(contentsOf: each.localizedOperations)
+                found.localizedOperations = Array(Set(found.localizedOperations + each.localizedOperations))
```
hboon, line 556: Does the output of the `Array()` need to be sorted?

oa-s: not sure about sorting,  is there any purposes for sorting it?

hboon (approved):

PR #6838, opened 2023-05-26
resolve memory leaks with TokenImportable

oa-s:


modules/AlphaWalletFoundation/AlphaWalletFoundation/Tokens/TokensAutodetector/ImportToken.swift
```diff
@@ -36,7 +36,7 @@ extension TokenOrContractFetchable {
 }
 
 //NOTE: actually its internal, public for tests
```
hboon, line 38: Unrelated, but making it `public` — this should be fixed by `@testable import`?

PR #6843, opened 2023-05-31
Added All Tests to Swift Format

eviltofu:


hboon:
@eviltofu thanks. But nah, let's hold off work on this to avoid massive changes now.

hboon:
Close for now

PR #6844, opened 2023-05-31
tokenscript attributes memory leak

oa-s:
pr fixes attributes resolving memory leaks for token script


hboon (commented):
What is the cause of the leak?

hboon:
Force-pushed after rebase

PR #6845, opened 2023-05-31
fix memory leaks when fetching event logs and calling call method

oa-s:


hboon (commented):
What is the cause for the leak?

hboon:
Closing for now.

PR #6971, opened 2023-08-29
[Explorer] Update  blockchain explorer URLs reflecting 301 result

hboon:


hboon:
@JamesSmartCell FYI. More

